PR #155, opened 2013-06-27
OHAI-485 Enable block device size given in bytes

nkrinner:
The current size we have is the size in blocks, which is not too helpful
for most people. What people usually want is the size in bytes.
Patch originally written by Vincent Untz vuntz@suse.com


lamont-granquist:
closing this due to inactivity

PR #156, opened 2013-07-02
[OHAI-487] use only those attributes that are provided by the linode provider

ranjib:


lamont-granquist:
closing due to inactivity

PR #237, opened 2013-12-10
Detect KVM guests using dmesg

maoe:
`ohai` cannot detect KVM guests if KVM/QEMU is running with `-cpu` option. Even if so, we still can detect them by searching for "Booting paravirtualized kernel on KVM" in `dmesg` output.


lamont-granquist:
closed due to inactivity

franklouwers:
elsif node.deep_fetch('dmi', 'system', 'manufacturer') == 'QEMU'
    virt = 'kvm'

PR #244, opened 2013-12-16
[OHAI-467] Prevents fe80:: link-local address from becoming ip6address

mpasternacki:


vegardx:
I don't think this works, I'm still getting fe80-addresses in node[:ip6address], or basically any address in the scope: Link, as long as nothing is defined in scope: Global.

mcquin:
@vegardx can you open an issue and provide your ohai version?

PR #317, opened 2014-04-18
OHAI-401 ohai on SmartOS uses the fqdn as the hostname

jacques:
One less patch set to throw around.  Updated my patchset to use hostname.rb vs the solaris/hostname.rb previously.


lamont-granquist:
This is actually working as expected, machinename is the reliable ohai attribute, fqdn is not necessarily reliable.

PR #318, opened 2014-04-18
OHAI-531 Regex for getting data from arp -na output is incorrect on Sola...

jacques:
...ris2 (SmartOS/IllumOS/etc.)  To replace the PR#229


lamont-granquist:
closing due to inactivity

PR #343, opened 2014-06-23
Fix network.rb for XenServer Creedence

sspans:
On the current alpha versions of xenserver the "ip addr" command skips "physical" interfaces, however "ip -d -s link" does list all interfaces.

This causes network.rb to fail on line 177 - because it tries to add state for an undefined interface.
The easiest fix is to create the iface[tmp_int] entry when we encounter a new interface during the parsing
of "ip -d -s link".

With this fix chef-client is again usable on XenServer.

Further debug output available on request.


sspans:
@btm: Any chance we could get this merged?

XenServer 6.5 has been released a month ago (http://xenserver.org/blog/entry/creedence-launches-as-xenserver-6-5.html), 
and chef-client/ohai will not work on the platform without this patch.

PR #400, opened 2014-08-27
Digital Ocean ohai/cloud support round 2

wingrunr21:
This is a follow-up to PR #347.

This PR:

* Adds an `IpHelper`. This provides a few useful methods for figuring out if an ip address is a private address and if it is a loopback address
* Implements a Digital Ocean ohai plugin. Digital Ocean is detected via a hint file or via the `/etc/digitalocean` file.
* Adds support for said plugin to the cloud plugin
* Adds support for said plugin to the cloud_v2 plugin

I decided to store the detected network addresses in a similar manner to how the v2 Digital Ocean API will be [reporting them](https://developers.digitalocean.com/v2/#droplets). Since we can detect network data directly, storing it in this way will help ease the transition over to the v2 API whenever `knife-digital_ocean` does so. The remainder of the hint file is kept under the v1 payload structure.

cc @mcquin @btm @brodock @wfarr

### DIFF
--- a/lib/ohai/plugins/cloud.rb
+++ b/lib/ohai/plugins/cloud.rb
@@ -25,6 +25,7 @@ Ohai.plugin(:Cloud) do
   depends "openstack"
   depends "azure"
   depends "cloudstack"
+  depends "digital_ocean"
 
   # Make top-level cloud hashes
   #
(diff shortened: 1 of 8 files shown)


lamont-granquist:
@juliandunn can we merge that as-is in order to not let the perfect be the enemy of the good?

wingrunr21:
Ya, I just haven't had time to get around to fixing this correctly. If the code doesn't work for this implementation then someone else will need to implement it.

btm:
This [travis failure](
https://travis-ci.org/chef/ohai/jobs/33734733) is a 1.8.7 failure, and can be ignored since [we no longer support 1.9.3](https://www.chef.io/blog/2014/11/25/ruby-1-9-3-eol-and-chef-12/), let alone 1.8.7. 1.8.7 was removed from .travis.yml in 7bd92d07cf04c71f3d62945126537bff68a89b3b.

PR #401, opened 2014-08-28
Update ec2.rb

rcsavage:
AWS/EC2 instance MAC addresses are no longer isolated to "fe:ff:ff:ff:ff:ff", and therefore throws off the EC2 detection logic.   The best approach is to just check for the AWS/EC2 metadata server.


lamont-granquist:
So we can't accept this approach because of the perf cost.  The MAC heuristic is poor, but hints are the better way to address this problem.

PR #416, opened 2014-09-24
Update mime-types

yourabi:
Update mime-types dep


lamont-granquist:
This probably should be ~> 2.0 instead of 2.3 since we don't have any hard requirement to avoid using 2.1 or 2.2.  It looks like chef was using 2.x (which doesn't entirely make any sense to me how that was working), and i think the major issue with 2.x was dropping ruby 1.8.7 support so we can go ahead and do this now.

To reiterate though, the more permissive ~> 2.0 constraint would be preferred.

PR #439, opened 2014-11-28
Fix connection detection logic.

kwilczynski:
The fact that we have socket back from IO::select, does not necessarily
mean that mean that we can actually ready and/or write to it.

Signed-off-by: Krzysztof Wilczynski <krzysztof.wilczynski@linux.com>


lamont-granquist:
if you want to finish this up, we'd still merge it.

kwilczynski:
@lamont-granquist I do. I need to push updates from my own fork now.

PR #441, opened 2014-11-30
Add additional information on the PHP engine versions to PHP plugin

### LINKED ISSUE #440: Zend Engine and Zend OPcache versions should be exposed in PHP plugin
I'd like to add the Zend Engine version and Zend Opcache version to the PHP plugin.  This data is really important for determining how you should template out the php.ini file and would be really helpful for anyone doing wordpress.  Before I go and toss something together I'm curious what Chef folks think the best way to handle the output would be.

Here's the current output:
    "php": {
      "version": "5.5.14",
      "builddate": "Sep  9 2014 19:09:25"
    },

Zend Engine version string will be in every version of PHP, but OPcache is optional so a big part of exposing this data is allowing people to see if the key exists.  Would something like this make sense?

    "php": {
      "version": "5.5.14",
      "builddate": "Sep  9 2014 19:09:25",
      "zend_engine_version": "v2.5.0",
      "zend_opcache_version": "v7.0.3"
    },

If opcache isn't present the key could be either nil or not present altogether.

tas50:
Closes #440

### DIFF
--- a/lib/ohai/plugins/php.rb
+++ b/lib/ohai/plugins/php.rb
@@ -1,6 +1,8 @@
 #
 # Author:: Doug MacEachern <dougm@vmware.com>
+# Author:: Tim Smith <tim@cozy.co>
 # Copyright:: Copyright (c) 2009 VMware, Inc.
+# Copyright:: Copyright (c) 2014 Cozy Services, Ltd.
 # License:: Apache License, Version 2.0
 #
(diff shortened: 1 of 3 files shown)


tas50:
Any word on getting this merged?

tas50:
ping @btm

tfitch:
This has been merged and will be in the next feature release.

Closes #440

PR #442, opened 2014-12-03
Support IBM System z s390 cpu

zshuo:
In IBM System z S390 archtecture,  the /proc/cpuinfo has totally different format with the x86 format. 
It looks like this:

```
# cat /proc/cpuinfo
vendor_id       : IBM/S390
<# processors    : 2
bogomips per cpu: 9328.00
features	: esan3 zarch stfle msa ldisp eimm dfp etf3eh highgprs 
processor 0: version = FF,  identification = 06E276,  machine = 2818
processor 1: version = FF,  identification = 06E276,  machine = 2818
```

So, we need to add additional code to make this s390 cpuinfo be resolved.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Ohai Changelog
 
 ## Unreleased:
-
+* Update specs to use RSpec 3 syntax
 
 ## Release 7.6.0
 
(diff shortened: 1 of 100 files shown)


lamont-granquist:
can you take out the CHANGELOG.md edit (that's our responsibility to do when merging) and see if this will rebase cleanly on master?

PR #447, opened 2014-12-19
Update rackspace plugin to support Windows distros.

akarpik:
Update rackspace plugin to support Windows distros.

### DIFF
--- a/lib/ohai/mixin/rackspace_metadata.rb
+++ b/lib/ohai/mixin/rackspace_metadata.rb
@@ -1,0 +1,51 @@
+#
+# Author:: Alexey Karpik <alexey.karpik@rightscale.com>
+# Author:: Cary Penniman (<cary@rightscale.com>)
+# Author:: Peter Schroeter <peter.schroeter@rightscale.com>
+# Author:: Stas Turlo <stanislav.turlo@rightscale.com>
+# Copyright:: Copyright (c) 2010-2014 RightScale Inc
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require "resolv"
+require "json"
+
+module ::Ohai::Mixin::RackspaceMetadata
+  def on_windows?
+    RUBY_PLATFORM =~ /windows|cygwin|mswin|mingw|bccwin|wince|emx/
+  end
+
+  def xenstore_command(command, args)
+    if on_windows?
+      xen_store_client_path = '"c:\Program Files\Citrix\XenTools\xenstore_client.exe"'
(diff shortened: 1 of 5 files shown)


lamont-granquist:
@adamedx has some concerns about shelling out to powershell and its expense

lamont-granquist:
@btm ping

akarpik:
So what is the status of this pull request?

thommay:
@btm @adamedx would you care to elucidate your concerns about powershell expense, please?

lib/ohai/mixin/rackspace_metadata.rb
```diff
@@ -0,0 +1,51 @@
...
+
+module ::Ohai::Mixin::RackspaceMetadata
+  def on_windows?
+    RUBY_PLATFORM =~ /windows|cygwin|mswin|mingw|bccwin|wince|emx/
+  end
+
+  def xenstore_command(command, args)
+    if on_windows?
+      xen_store_client_path = '"c:\Program Files\Citrix\XenTools\xenstore_client.exe"'
+      xen_store_client_alt  = "powershell -NoProfile -NonInteractive -InputFormat None -ExecutionPolicy Bypass \"#{File.join(File.dirname(__FILE__), "xenstore_client.ps1")}\""
```
adamedx: bypass execution policy is not supported prior to PS 3, so this will break on default installations of Windows Server 2008R2. You should see if you can make it work with unrestricted if you're going to use PowerShell.

lib/ohai/mixin/rackspace_metadata.rb
```diff
@@ -0,0 +1,51 @@
...
+      client = File.exists?(xen_store_client_path.tr('"', '')) ? xen_store_client_path : xen_store_client_alt
```
adamedx: Should this be `::File.exist?` instead of `File.exists?` ?

lib/ohai/mixin/rackspace_metadata.rb
```diff
@@ -0,0 +1,51 @@
...
+require "json"
+
+module ::Ohai::Mixin::RackspaceMetadata
+  def on_windows?
+    RUBY_PLATFORM =~ /windows|cygwin|mswin|mingw|bccwin|wince|emx/
+  end
+
+  def xenstore_command(command, args)
+    if on_windows?
+      xen_store_client_path = '"c:\Program Files\Citrix\XenTools\xenstore_client.exe"'
```
adamedx: The hard-coded path won't work -- for example, Windows can be (and often is, especially in datacenters) installed without a C drive. Also, any software can be installed to an arbitrary location. Administrators control that through Windows Installer / Group Policy / SCCM, etc. So there needs to be a different way of finding this command -- if it's in the path, just specifying the executable should work.
A more reliable way would be to look for a registry entry created by this software -- those are canonical and can't be configured by users. Ruby has helper code to read the registry: http://ruby-doc.org/stdlib-2.0.0/libdoc/fiddle/rdoc/Win32/Registry/API.html.

lib/ohai/mixin/rackspace_metadata.rb
adamedx: So ultimately why is the script being run? Is it to get some output such as a version? Ultimately looking in the registry may be a quicker and more reliable way to do this. Process creation overhead is not great in Windows unfortunately, and while PowerShell.exe has become faster / slimmer in recent Windows versions, in Win 2008R2 it was still a little sluggish, probably due to its reliance on .net.

adamedx:
Well, I don't see my original comment, but in general PowerShell.exe is a little slow on older versions of Windows, and in general, if you can avoid starting a process on Windows just to get some information, you should do so to avoid slowing down Ohai runs. Its not clear to me what the plugin is looking for, but use of the registry or WMI (which is being used elsewhere here) is more efficient and reliable (it's hard to find the right process to start if it's not in the PATH).

thommay:
Closing due to old age, if you'd like to finish up this plugin we'd be happy to accept it. Thanks for your contribution!

PR #448, opened 2014-12-19
Add support for Softlayer cloud

akarpik:
Add support for Softlayer cloud

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,7 +11,7 @@ gem "rspec_junit_formatter", :git => 'https://github.com/sj26/rspec_junit_format
 group :development do
   gem "chef", github: "opscode/chef", branch: "master"
 
-#  gem "sigar", :platform => "ruby"
+  gem "sigar", :platform => "ruby"
   gem 'plist'
 
   # gem 'pry-debugger'
--- a/Gemfile
+++ b/Gemfile
@@ -11,7 +11,7 @@ gem "rspec_junit_formatter", :git => 'https://github.com/sj26/rspec_junit_format
 group :development do
   gem "chef", github: "opscode/chef", branch: "master"
 
-  gem "sigar", :platform => "ruby"
+#  gem "sigar", :platform => "ruby"
   gem 'plist'
 
   # gem 'pry-debugger'
--- a/lib/ohai/mixin/softlayer_metadata.rb
+++ b/lib/ohai/mixin/softlayer_metadata.rb
@@ -57,7 +57,7 @@ module ::Ohai::Mixin::SoftlayerMetadata
       ::Ohai::Log.error("Unable to fetch item #{full_url}: status (#{res.code}) body (#{res.body})")
       nil
     end
-  rescue Exception => e
+  rescue => e
     ::Ohai::Log.error("Unable to fetch softlayer metadata from #{u}: #{e.class}: #{e.message}")
     nil
   end
--- a/lib/ohai/mixin/softlayer_metadata.rb
+++ b/lib/ohai/mixin/softlayer_metadata.rb
@@ -59,6 +59,6 @@ module ::Ohai::Mixin::SoftlayerMetadata
     end
   rescue => e
     ::Ohai::Log.error("Unable to fetch softlayer metadata from #{u}: #{e.class}: #{e.message}")
-    nil
+    raise e
   end
 end
--- a/lib/ohai/mixin/softlayer_metadata.rb
+++ b/lib/ohai/mixin/softlayer_metadata.rb
@@ -32,8 +32,6 @@ module ::Ohai::Mixin::SoftlayerMetadata
       'region'        => fetch_metadata_item("getDatacenter.txt"),
       'instance_id'   => fetch_metadata_item("getId.txt")
     }
-
-    metadata
   end
 
   # Softlayer's metadata api is only available over HTTPS.
@@ -46,23 +44,21 @@ module ::Ohai::Mixin::SoftlayerMetadata
   end
 
   def fetch_metadata_item(item)
-    begin
-      full_url = "#{SOFTLAYER_API_QUERY_URL}/#{item}"
-      u = URI(full_url)
-      net = ::Net::HTTP.new(u.hostname, u.port)
-      net.ssl_version = "TLSv1"
-      net.use_ssl = true
-      net.ca_file = ca_file_location
-      res = net.get(u.request_uri)
-      if res.code.to_i.between?(200,299)
-        res.body
-      else
-        ::Ohai::Log.error("Unable to fetch item #{full_url}: status (#{res.code}) body (#{res.body})")
-        nil
-      end
-    rescue Exception => e
-      ::Ohai::Log.error("Unable to fetch softlayer metadata from #{u}: #{e.class}: #{e.message}")
+    full_url = "#{SOFTLAYER_API_QUERY_URL}/#{item}"
+    u = URI(full_url)
+    net = ::Net::HTTP.new(u.hostname, u.port)
+    net.ssl_version = "TLSv1"
+    net.use_ssl = true
+    net.ca_file = ca_file_location
+    res = net.get(u.request_uri)
+    if res.code.to_i.between?(200,299)
+      res.body
+    else
+      ::Ohai::Log.error("Unable to fetch item #{full_url}: status (#{res.code}) body (#{res.body})")
       nil
     end
+  rescue Exception => e
+    ::Ohai::Log.error("Unable to fetch softlayer metadata from #{u}: #{e.class}: #{e.message}")
+    nil
   end
 end
--- a/lib/ohai/mixin/softlayer_metadata.rb
+++ b/lib/ohai/mixin/softlayer_metadata.rb
@@ -1,0 +1,68 @@
+#
+# Author:: Alexey Karpik <alexey.karpik@rightscale.com>
+# Author:: Peter Schroeter <peter.schroeter@rightscale.com>
+# Author:: Stas Turlo <stanislav.turlo@rightscale.com>
+# Copyright:: Copyright (c) 2010-2014 RightScale Inc
+# License:: Apache License, Version 2.0
+#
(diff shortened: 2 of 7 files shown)


lib/ohai/plugins/softlayer.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  provides 'softlayer'
+
+  # Identifies the softlayer cloud
+  #
+  # === Return
+  # true:: If the softlayer cloud can be identified
+  # false:: Otherwise
+  def looks_like_softlayer?
+    !!hint?('softlayer')
```
drag00n: I'm a little confused on the need for a !!. Looking through the hints code I see that it will either return a hash or nil(which will evaluate to false). Can you just use ```hint?('softlayer')```

lib/ohai/mixin/softlayer_metadata.rb
```diff
@@ -0,0 +1,68 @@
...
+  def fetch_metadata
+    metadata  = {
+      'public_fqdn'   => fetch_metadata_item("getFullyQualifiedDomainName.txt"),
+      'local_ipv4'    => fetch_metadata_item("getPrimaryBackendIpAddress.txt"),
+      'public_ipv4'   => fetch_metadata_item("getPrimaryIpAddress.txt"),
+      'region'        => fetch_metadata_item("getDatacenter.txt"),
+      'instance_id'   => fetch_metadata_item("getId.txt")
+    }
+
+    metadata
```
lamont-granquist: this line is useless in ruby.

lib/ohai/mixin/softlayer_metadata.rb
```diff
@@ -0,0 +1,68 @@
...
+  # Ruby by default does not link to the system's CA bundle
+  # however Chef-omnibus should set SSL_CERT_FILE to point to a valid file.
+  # Manually supply and specify a suitable CA bundle here or
+  # set the SSL_CERT_FILE file environment variable to a valid value otherwise.
+  def ca_file_location
+    ::Ohai::Config[:ca_file]
+  end
+
+  def fetch_metadata_item(item)
+    begin
```
lamont-granquist: this outer layer of begin/end can be dropped the def/end can bracket a rescue as well.

lib/ohai/plugins/softlayer.rb
lamont-granquist: Yeah, I'm not sure why this is really necessary.

spec/unit/mixin/softlayer_metadata_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    "/rest/v3.1/SoftLayer_Resource_Metadata/#{item}"
+  end
+
+  def make_res(body)
+    double("response", {:body => body, :code => "200"})
+  end
+
+
+  context 'fetch_metadata' do
+    it "creates empty softlayer hash on api query errors" do
```
lamont-granquist: wouldn't we want to not create the attributes at all if we get an error talking to the api?

spec/unit/plugins/softlayer_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require File.expand_path(File.dirname(__FILE__) + '/../../spec_helper.rb')
+
+describe Ohai::System, 'plugin softlayer' do
+
+  before(:each) do
+    @plugin = get_plugin("softlayer")
```
lamont-granquist: this would be better as a `let(:plugin) { get_plugin("softlayer") }` and get rid of the ivar

spec/unit/plugins/softlayer_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    allow(@plugin).to receive(:fetch_metadata).and_raise(Exception.new("TEST"))
+    @plugin.run
+    expect(@plugin[:softlayer]).to be_nil
+  end
+
+  it "create empty node if fetch_metadata return empty hash" do
+    allow(@plugin).to receive(:hint?).with('softlayer').and_return(true)
+    allow(@plugin).to receive(:fetch_metadata).and_return({})
+    @plugin.run
+    expect(@plugin[:softlayer]).to eq({})
```
lamont-granquist: same comment here this seems unexpected.

spec/unit/mixin/softlayer_metadata_spec.rb
akarpik: Some recipes may not use softlayer node attributes, but they need to know, that they are running on SoftLayer cloud. That is why, we are creating empty node on errors.

smurawski:
This might need a rebase due to some other previous updates to master.

cmluciano:
ping @akarpik can you rebase and resubmit?

Gemfile
```diff
@@ -11,7 +11,7 @@ gem "rspec_junit_formatter", :git => 'https://github.com/sj26/rspec_junit_format
 group :development do
   gem "chef", github: "opscode/chef", branch: "master"
 
-  gem "sigar", :platform => "ruby"
+#  gem "sigar", :platform => "ruby"
```
lamont-granquist: looks like something edited out in testing that should be dropped?

lib/ohai/mixin/softlayer_metadata.rb
```diff
@@ -0,0 +1,64 @@
...
+    net.use_ssl = true
+    net.ca_file = ca_file_location
+    res = net.get(u.request_uri)
+    if res.code.to_i.between?(200,299)
+      res.body
+    else
+      ::Ohai::Log.error("Unable to fetch item #{full_url}: status (#{res.code}) body (#{res.body})")
+      nil
+    end
+  rescue Exception => e
```
lamont-granquist: That shouldn't be `rescue Exception => e`, should just be `rescue => e`

lamont-granquist: This also may not be too useful since ohai plugins already have an exception rescuing layer around them, it can most likely be dropped entirely and just let the exception bubble up and abort the plugin.

spec/unit/mixin/softlayer_metadata_spec.rb
lamont-granquist: This is probably not a good idea.  If the softlayer metadata service returns a transient 500 then we're going to silently ignore that and set the softlayer hash and attempt to continue, but have bad (no) data in the softlayer hash.

Granted we have a larger issue in chef itself where we can't mark that certain plugins are required and should cause the run to abort, but we need to fix that issue as well.

lib/ohai/mixin/softlayer_metadata.rb
akarpik: We don't need to abort plugin. We need to have empty softlayer node, even if we get error during fetching metadata. Lets say, we have plugins that required just cloud provider, and don't care about metadata at all. So if we drop softlayer node, they will not work.

akarpik: @lamont-granquist - are you sure that we need to catch only StantardError exceptions and allow Exceptions pass through?

lamont-granquist: Yes, "rescue Exception" and suppressing the exception is bad.

http://stackoverflow.com/questions/10048173/why-is-it-bad-style-to-rescue-exception-e-in-ruby
http://daniel.fone.net.nz/blog/2013/05/28/why-you-should-never-rescue-exception-in-ruby/
https://robots.thoughtbot.com/rescue-standarderror-not-exception

There is actually a case for doing that in order to decorate the exception, but then you'd re-raise the exception in the handler.  We probably do rescue Exception and then squash the exception in other places in the chef+ohai code but those are bugs.

spec/unit/mixin/softlayer_metadata_spec.rb
akarpik: As I have explained below, even if we could not get data from metadata server, we logged error, we still need to have softlayer node, so other plugins get detect on which cloud they are running on.

lamont-granquist:
@chef/client-core @chef/client-engineers i sorta feel strongly that this is still incorrect behavior with the empty hash for softlayer attrs if the metadata service returns an error.   the kinds of conditions that i'm worried about are now we don't get our ip addresses in the metadata and the chef run continues successfully and then the node is posted and now searches for the node will not have the ip address.  it would be better to allow the error here.  it would be even better to be able to abort ohai and abort the wrapping chef run completely.

need some more thoughts on resolving this.

danielsdeleo:
I think for now it should be consistent with what the other plugins do. With Ohai 7 we wanted to implement better behavior for error handling (read: don't just swallow failures) and also some way of failing the chef run if certain attributes didn't get set in ohai, but ran out of time. We're a lot closer with the Ohai 7 architecture though.

jonlives:
:+1: for consistency, as long as it's something that does eventually get fixed

smurawski:
@akarpik We understand the use case for wanting to leave the softlayer node as a hint, but that is not the behavior of existing plugins and the direction we want to go with Ohai includes adding support to bubble up errors and deal with them in a consistent manner.  This would add an exception to the current use case and probably break moving forward.  

If you remove that behavior, we can merge in the additional softlayer metadata changes, otherwise we'll have to close the PR.

akarpik:
@smurawski, I will remove behaviour, and update tests.

smurawski:
@akarpik Thank you!  Looking forward to getting your softlayer support merged!

cmluciano:
how's this looking now @smurawski

spec/unit/mixin/softlayer_metadata_spec.rb
mcquin: @akarpik can you update spec description to match expected behavior?

thommay:
:+1: when this is merged we'll need to be careful about the changelog.

thommay:
cc @chef/client-core for one more review.

lamont-granquist:
:+1:

PR #449, opened 2014-12-19
Update Azure cloud plugin to query server for data instead of reading th...

akarpik:
Update Azure cloud plugin to query server for data instead of reading them form hint file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased:
 
+* Update Azure cloud plugin to query server for data instead of reading them form hint file
+
 ## Release 8.0.0
 
 * [**sawanoboly**](https://github.com/sawanoboly)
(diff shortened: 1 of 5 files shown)


lamont-granquist:
@adamedx had a lot of concerns about this whole approach, the dhcp server isn't necessarily a metadata server and its reponses can't be considered to be reliable.

lamont-granquist:
@btm ping

adamedx:
@jeffmendoza, @nimishas, is this approach a valid way to query Azure for metadata? There is no formal metadata service from what I can tell for Azure. This also appears to be limited to Linux and won't handle the Windows case.

jeffmendoza:
@johngossman do you want to take a look? We had discussed a solution of looking in /var/lib/waagent/SharedConfig.xml in the past. Not sure about this dhcp solution.

akarpik:
@jeffmendoza, basically /var/lib/waagent/SharedConfig.xml solution is only available for Linux. That is why we have decided to find another way to get this data.

lamont-granquist:
can we close this?

akarpik:
@lamont-granquist you want to merge and close, or just close?

akarpik:
So what is going on with this pull request?

thommay:
@akarpik it looks like the consensus is to close this, so I am doing so.

PR #450, opened 2014-12-19
Fix up incorrect cloudstack metadata

akarpik:
Fix up incorrect cloudstack metadata

### DIFF
--- a/lib/ohai/plugins/cloudstack.rb
+++ b/lib/ohai/plugins/cloudstack.rb
@@ -1,5 +1,7 @@
 #
+# Author:: Peter Schroeter <peter.schroeter@rightscale.com>
 # Author:: Olle Lundberg (<geek@nerd.sh>)
+# Copyright:: Copyright (c) 2010-2014 RightScale Inc
 # Copyright:: Copyright (c) 2014 Opscode, Inc.
 # License:: Apache License, Version 2.0
 #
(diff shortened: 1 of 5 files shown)


lamont-granquist:
my only concern would be if this changes an API and might be considered a breaking change.

@chef/client-engineers @chef/client-core needs review

mcquin:
:+1:

lamont-granquist:
:+1:

looks like it needs a rebase.

lib/ohai/plugins/cloudstack.rb
```diff
@@ -31,6 +33,11 @@
         cloudstack Mash.new
         Ohai::Log.debug("connecting to the 'cloudstack' metadata service")
         fetch_metadata.each { |k, v| cloudstack[k] = v }
+        # Note: the cloudstack public_ipv4 is the value of the NAT router (as per cloudstack documentation)
+        # and not necessarily the publicly available IP.  cloustack semi-supports floating
+        # ips in that the public ip for an instance can be an IP different from the NAT router
+        cloudstack['router_ipv4'] = cloudstack.delete('public_ipv4')
```
tyler-ball: It seems like this is a compatiability-breaking change, unless it was considered a bug before and people were using workarounds

lamont-granquist: From the description it sounds like this was a bug and we were violating the documented API for cloudstack.

btm:
rebased and merged to master.

PR #455, opened 2015-01-08
Add language elixir.

cmluciano:


lamont-granquist:
the spec could use let syntax instead of ivars

but :+1: either way.

smurawski:
@mcquin or another maintainer this looks good, can we merge?

mcquin:
LGTM.

Would prefer to see the spec updated per Lamont's comment, but that's not essential.

PR #458, opened 2015-01-13
Update WARNING for ohai 7 syntax docs page.

cmluciano:
The current path points to an outdated URL for Chef docs. This commit updates
the URL to the fancy .io address.


lamont-granquist:
:+1:

PR #461, opened 2015-01-14
Jdm/json backport

jdmundrawala:


lamont-granquist:
lol, okay, yes.  :+1:

merge this, then tag and release.

i looked into the mixlib-shellout thing and we're only pinning ~> 1.2 while rubygems is pulling in 1.6.1 even though we're on ruby 1.8.7/1.9.2 which is dumb, expected, and nothing we can do about it.  this branch supports 1.8.7/1.9.2 so we can't exclude them like mixlib-shellout 1.6.x does.

bumping rubygems in the travis config might let the tests run on the older versions of ruby.

PR #468, opened 2015-01-26
Add language rust.

cmluciano:
I chose to not trim the tag off of the semver value because I
think it is important to distinguish between RC, nightly, and alpha.

### DIFF
--- a/lib/ohai/plugins/rust.rb
+++ b/lib/ohai/plugins/rust.rb
@@ -1,0 +1,32 @@
+# Author:: Christopher M Luciano (<cmlucian@us.ibm.com>)
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+Ohai.plugin(:Rust) do
+  provides "languages/rust"
+
+  depends "languages"
+
+  collect_data do
+    output = nil
+
+    rust = Mash.new
+    so = shell_out("rustc --version")
+    if so.exitstatus == 0
+      output =  so.stdout.split
+      rust[:version] = output[1]
+      languages[:rust] = rust if rust[:version]
+    end
+  end
+end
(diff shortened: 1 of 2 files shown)


lib/ohai/plugins/rust.rb
```diff
@@ -0,0 +1,32 @@
...
+  provides "languages/rust"
+
+  depends "languages"
+
+  collect_data do
+    output = nil
+
+    rust = Mash.new
+    so = shell_out("rustc --version")
+    if so.exitstatus == 0
```
lamont-granquist: can you just `shell_out!` here and let the exception break you out so that you don't need the conditional?

lamont-granquist:
One nit, but LGTM :+1:

spec/unit/plugins/rust_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require File.expand_path(File.join(File.dirname(__FILE__), '..', '..', '/spec_helper.rb'))
+
+describe Ohai::System, "plugin rust" do
+  let(:stdout) {"rustc 1.0.0-nightly (29bd9a06e 2015-01-20 23:03:09 +0000)"}
+
+  let (:plugin) do
```
mcquin: Small whitespace fix here.

lamont-granquist:
@chef/client-core @chef/client-engineers review?

jonlives:
:+1:

lib/ohai/plugins/rust.rb
cmluciano: @lamont-granquist I tried it without the conditional and then noticed that my "should not set rust if rust command fails" test failed. When I tried to run "rustc --version" on a computer w/o rust installed I didn't get my expected output but instead "command not found: rustc" with an exit code of 127. So I still get valid text from stdout assigned to my output variable.

I think that it would still require validation of sorts with a conditional if I didn't check the exit code.

cmluciano:
Whitespace fixed

lib/ohai/plugins/rust.rb
mcquin: I believe `#shell_out!` will raise an exception if you get a non-zero exitcode. See the comments [here](https://github.com/chef/mixlib-shellout/blob/master/lib/mixlib/shellout.rb#L251-L279).

lamont-granquist: Yeah, and ohai plugins can raise and that gets ignored and ohai moves on to the next plugin.

cmluciano: Hmm. For some reason my third test keep getting true for the has.key?[:rust]. Can anything spot anything odd that I am doing in my tests?

mcquin: I think that may be coming from line 24. Try switching lines 24 and 25.

cmluciano: No dice :(

cmluciano:
@lamont-granquist Did you mean to close this? I see my elixir PR is added but not rust.

lamont-granquist:
I meant to merge it, but I certainly failed to rebase it into my merge pr

PR #473, opened 2015-02-01
Archlinux Version

### LINKED ISSUE #405: Platform version is missing on archlinux platform
When running ohai, on archlinux, platform_version is not filled.

This comes from archlinux having no "version" per se since it is a rolling release distro.
https://github.com/opscode/ohai/blob/7-stable/lib/ohai/plugins/linux/platform.rb#L84 makes it clear.

Hovewer, the comment on https://github.com/opscode/ohai/blob/7-stable/lib/ohai/plugins/linux/platform.rb#L85 suggests we could use the kernel version instead.

(ps: having no platform version makes chef test crash)

rychipman:
Solves #2843, and is a better alternative to PR #2844, since it solves the cause of the issue (lack of platform_version) instead of patching the symptoms every time it is encountered.

Ohai doesn't get a platform_version for rolling-release linux distributions. This causes problems when platform_version is assumed to exist elsewhere in chef (such as spec/spec_helper.rb)

Uses the kernel version (uname -r) as the platform version for rolling-release distros

closes #405

### DIFF
--- a/lib/ohai/plugins/linux/platform.rb
+++ b/lib/ohai/plugins/linux/platform.rb
@@ -87,6 +87,7 @@ Ohai.plugin(:Platform) do
       platform "arch"
       # no way to determine platform_version in a rolling release distribution
       # kernel release will be used - ex. 2.6.32-ARCH
+      platform_version File.read("/proc/version").scan(/Linux version (\d+\.+) /)
     elsif File.exists?('/etc/exherbo-release')
       platform "exherbo"
       # no way to determine platform_version in a rolling release distribution
--- a/lib/ohai/plugins/linux/platform.rb
+++ b/lib/ohai/plugins/linux/platform.rb
@@ -16,12 +16,9 @@
 # limitations under the License.
 #
 
-#require 'ohai/plugin/kernel'
-
 Ohai.plugin(:Platform) do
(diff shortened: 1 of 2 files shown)


lamont-granquist:
Heh there's even a comment there that someone intended to do that.

Spec would be nice if possible

rychipman:
Added spec tests

lamont-granquist:
@chef/client-core @chef/client-engineers review

tyler-ball:
:+1:

jdmundrawala:
:shipit:

danielsdeleo:
:+1:

lib/ohai/plugins/linux/platform.rb
```diff
@@ -87,10 +87,12 @@ def get_redhatish_version(contents)
       platform "arch"
       # no way to determine platform_version in a rolling release distribution
       # kernel release will be used - ex. 2.6.32-ARCH
+      platform_version `uname -r`.strip
```
mcquin: Should we be using shell_out instead of backticks here? @lamont-granquist @danielsdeleo

lamont-granquist: For consistency, yes.  I'm not sure that practically it'll matter because `uname -r` is so trivial.

PR #474, opened 2015-02-02
Removing trailing space and '\r' for windows

### LINKED ISSUE #470: DMI plugin stripping issue
Hi everybody,

There is a slice issue when using the dmi plugins on windows the parsing of dmidecode doesn't strip the reponse so every value has a '\r'.

Regards,
JM

jmauro:
Closes #470

### DIFF
--- a/lib/ohai/common/dmi.rb
+++ b/lib/ohai/common/dmi.rb
@@ -106,10 +106,11 @@ module Ohai
               next if field.to_s == 'size'
               next if field.to_s == 'record_id'
               translated = field.downcase.gsub(/[^a-z0-9]/, '_')
+              value      = value.strip
               if in_common.has_key?(translated)
-                in_common[translated] = nil unless in_common[translated] == value.strip
+                in_common[translated] = nil unless in_common[translated] == value
               else
-                in_common[translated] = value.strip
+                in_common[translated] = value
               end
             }
           }
--- a/lib/ohai/common/dmi.rb
+++ b/lib/ohai/common/dmi.rb
@@ -107,15 +107,15 @@ module Ohai
               next if field.to_s == 'record_id'
               translated = field.downcase.gsub(/[^a-z0-9]/, '_')
               if in_common.has_key?(translated)
-                in_common[translated] = nil unless in_common[translated] == value
+                in_common[translated] = nil unless in_common[translated] == value.strip
               else
-                in_common[translated] = value
+                in_common[translated] = value.strip
               end
             }
           }
           in_common.each{ |field, value|
             next if value == nil
-            dmi[type][field] = value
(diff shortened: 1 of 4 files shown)


lamont-granquist:
@chef/client-engineers @chef/client-core needs review

lib/ohai/common/dmi.rb
```diff
@@ -107,15 +107,15 @@ def convenience_keys(dmi)
               next if field.to_s == 'record_id'
               translated = field.downcase.gsub(/[^a-z0-9]/, '_')
               if in_common.has_key?(translated)
-                in_common[translated] = nil unless in_common[translated] == value
+                in_common[translated] = nil unless in_common[translated] == value.strip
```
jdmundrawala: I think you should pull `value.strip` up since it is being repeated multiple times

danielsdeleo:
Agree w/ @jdmundrawala :+1: otherwise

spec/unit/plugins/dmi_spec.rb
```diff
@@ -91,6 +91,10 @@
 	Contained Elements: 0
 EOS
 
+def convert_window_output(stdout)
```
tyler-ball: This seems useful enough to put into a `spec_helper` of some kind - I would normally only advocate this if something else already needed this functionality, but I think putting it in a helper will encourage future spec writers to leverage it if they see it being used here.

tyler-ball:
:+1: other than my minor comment and echoing Jay/Dan

jmauro:
Done

spec/spec_helper.rb
```diff
@@ -32,6 +32,10 @@ def get_plugin(plugin, ohai = Ohai::System.new, path = PLUGIN_PATH)
   loader.load_plugin(File.join(path, "#{plugin}.rb"))
 end
 
+def convert_window_output(stdout)
```
btm: convert_windows_output?

jmauro: This wasn't what I was supposed to do?

danielsdeleo: I think @btm is just saying this should be "windows" with an "s" at the end, minor typo.

jmauro:
Typo corrected.

danielsdeleo:
All the code review stuff has been addressed and this has all the :+1:s it needs, let's merge it.

tyler-ball:
appveyor tests failing - looks like the rename needs to be added in those tests

jmauro:
Fixing the method calls into the spec.

btm:
:+1:

btm:
rebased into 55ae7f38627c6eab2599fb00fc59395b9d6a60e1 and merged to master.

PR #475, opened 2015-02-03
Expand version number constraint for mime-types dependency

dan-jensen:
The mime-types version constraint has been changed from '~> 1.16' to '>= 1.16' so it will no longer cause Bundler conflicts with gems that specify a current version of the mime-types gem (e.g. CarrierWave).  (The ~> 1.16 constraint doesn't match anything more recent than June 2012.)  All tests are passing.

### DIFF
--- a/ohai.gemspec
+++ b/ohai.gemspec
@@ -15,7 +15,7 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = ">= 2.0.0"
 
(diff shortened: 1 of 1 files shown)


mcquin:
It looks like the dependency has been updated to `~> 2.0` in master: https://github.com/chef/ohai/blob/master/ohai.gemspec#L18

PR #480, opened 2015-02-09
block_device rotational key

atomic-penguin:
* Added rotational key (0/1) for differentiation of Solid-State and
  Rotational Magnetic storage.

### DIFF
--- a/lib/ohai/plugins/linux/block_device.rb
+++ b/lib/ohai/plugins/linux/block_device.rb
@@ -35,6 +35,11 @@ Ohai.plugin(:BlockDevice) do
             File.open("/sys/block/#{dir}/device/#{check}") { |f| block[dir][check] = f.read_nonblock(1024).strip }
           end
         end
+        %w{rotational}.each do |check|
+          if File.exists?("/sys/block/#{dir}/queue/#{check}")
+            File.open("/sys/block/#{dir}/queue/#{check}") { |f| block[dir][check] = f.read_nonblock(1024).strip }
+          end
+        end
       end
       block_device block
     end


lamont-granquist:
sexiness.

any chance for a spec test?

atomic-penguin:
There was no existing spec in-place for this plugin.  If someone wants to write one example for the existing unit, I would be happy to flesh the rest of it out when I have the chance.

lamont-granquist:
@chef/client-engineers @chef/client-core review?

mcquin:
:penguin:

atomic-penguin:
:metal:

tyler-ball:
:+1: 

I would like to see Windows and other systems updated to also support this, but it doesn't look like anything else has a `block_device` plugin.  So it can wait until someone requests that information for the other systems!

atomic-penguin:
@tyler-ball I wouldn't even know where to begin with Windows, 'cause I can't just read it from a file in /proc to expose it.  I imagine that involves shelling out to Powershell to get that kind of answer.

tyler-ball:
@atomic-penguin yeah, don't worry about it - I was just writing-out-loud-thoughts-in-my-head

lamont-granquist:
Yeah, this is linux-only.  Trying to standardize on a schema for pulling block dev info from all the different distros may also be a bit of madness...  As a sysadmin I'd sort of expect to have to roll my own case statements there.  Getting the virtualization schema correct so that when I move my linux servers across different clouds and virtualization technologies is probably where we need to focus our efforts.

Its not a bad idea, I just can't see us having enough human resources to focus on that given all the other things we have to do -- its gonna be hella fussy and take a long time and the payoff won't be all that large.

PR #482, opened 2015-02-11
Fixing inconsistencies between Linux and Windows (GH-481)

### LINKED ISSUE #481: Linux vs Windows filesystem plugin inconsistencies
I just noticed that the Linux filesystem plugin uses strings for a couple of fields (see below) where the Windows filesystem plugin uses numbers for these same fields. This causes issues when trying to use the Ohai info from both Linux and Windows machines by any tool written in a statically typed language.

The fields in question are:
```
KBSize
KBUsed
KBavailable
PercentUsed
```

The fix is simple, so will add a PR for that shortly...

svanharmelen:
This fixes GH-481 by making sure both the Linux and Windows filesystem plugins return the same types

Closes #481

### DIFF
--- a/lib/ohai/plugins/windows/filesystem.rb
+++ b/lib/ohai/plugins/windows/filesystem.rb
@@ -39,10 +39,19 @@ Ohai.plugin(:Filesystem) do
       disk.wmi_ole_object.properties_.each do |p|
         ld_info[filesystem][p.name.wmi_underscore.to_sym] = disk[p.name.downcase]
       end
-      fs[filesystem][:kb_size] = ld_info[filesystem][:size].to_i / 1000
-      fs[filesystem][:kb_available] = ld_info[filesystem][:free_space].to_i / 1000
-      fs[filesystem][:kb_used] = fs[filesystem][:kb_size].to_i - fs[filesystem][:kb_available].to_i
-      fs[filesystem][:percent_used]  = (fs[filesystem][:kb_size].to_i != 0 ? fs[filesystem][:kb_used].to_i * 100 / fs[filesystem][:kb_size].to_i : 0)
+
+      # Convert to integers and calculate the needed values
+      kb_size = ld_info[filesystem][:size].to_i / 1000
+      kb_available = ld_info[filesystem][:free_space].to_i / 1000
+      kb_used = fs[filesystem][:kb_size].to_i - fs[filesystem][:kb_available].to_i
+      percent_used = (fs[filesystem][:kb_size].to_i != 0 ? fs[filesystem][:kb_used].to_i * 100 / fs[filesystem][:kb_size].to_i : 0)
+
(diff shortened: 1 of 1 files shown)


lamont-granquist:
@chef/client-core @chef/client-engineers review?

technically this might be a breaking change, although i'd argue that given the discrepancy that we don't have a documented interface here and that we're fixing an undocumented behavior so a minor release would be fine.  i'm turning into a semver anarchist tho.

tyler-ball:
I think we should argue that inconsistencies between the way we treat Windows and Linux should be treated as bugs, publish lots of documentation about changes and when they are occurring, release it in a minor version change, and move towards a better world.  Together.

lamont-granquist:
This is likely to be differences not only between Windows and Linux, though, but also Linux vs Solaris vs AIX vs FreeBSD vs whatever other trouble we find ourselves in...

We should probably consider formalizing a description of the ohai data using JSON Schema or some other tool if we start going down that route...

mcquin:
:+1:

jonlives:
Yeah, I think this is minor-releasable. 

:+1:

jdmundrawala:
I'd like to urge caution here. Just because we don't have it documented does not mean it isn't documented. The fact is that the source is out there, and that can also serve as documentation.

smurawski:
@jdmundrawala so is that a thumb down from you or just a cautionary tale?  If not a thumbs down, let's get this merged.

jdmundrawala:
I don't want to hear it from people that we broke windows people again, so unless you guys are 100% this won't break people, I'm :-1: without a major version of Chef.
And we should also do what @lamont-granquist said

svanharmelen:
So... Any progress/decisions on this one? I personally see this as a bug as would really like to make things/behaviour more consistent over different operating systems...

danielsdeleo:
I agree with the goal, but we need a safe way to get there. If you have to change all of your cookbooks (some of which might be 3rd party cookbooks you don't own) at once, that needs to either go in a major version bump or we need some kind of API versioning for Ohai to let you gradually opt-in.

svanharmelen:
So... Shall I just close this PR? Seems clear this will never happen...

thommay:
@svanharmelen please leave it open. I'm going to tag it up so we include it with our next major release, which will be Chef 13/Ohai 9.

jaymzh:
Also, fwiw, I'm trying hard to get everyone away from filesystem plugin and to the filesystem2 plugin... the filesystem plugin is super buggy in a lot of other ways... I'd rather implement fs2 on windows and make it the "one true way".

tas50:
We've killed off filesystem so this is going to need to be reworked at this point since the *nix behavior changed. If you care to rebase it we can schedule it for Chef 14. Feel free to open it back up and continue the work.

PR #483, opened 2015-02-11
Workaround for hyperic/sigar#48, throw on 0 pid

### LINKED ISSUE #423: Invalid pid, name used after sigar fails to throw exception
Filed a bug with sigar (https://github.com/hyperic/sigar/issues/48), but observed this through chef/ohai, so it is susceptible to it now and can be worked around.

The rescue call in lib/ohai/plugins/network_listeners.rb is insufficient to prevent invalid pids from being used.  After sigar throws an exception for a single 0 pid (its invalid pid return) call to proc_state, all subsequent proc_state calls within the same sigar context do not throw.  This can be worked around until they correct this behavior by throwing on pid == 0:

```ruby
begin
  pid = sigar.proc_port(conn.type, port)
  throw ArgumentError.new("No such process") if pid == 0
  listeners[port][:pid] = pid
  listeners[port][:name] = sigar.proc_state(pid).name
rescue
end
```

werkt:
When proc_state is called for the second and all future times on a
single sigar instance, it fails to throw an exception, resulting in
invalid memory access during ohai reporting of process names. This
workaround causes an invalid pid (0) to be caught and prevent the
invalid access.  Closes #423 and will be unnecessary and should be
reverted when hyperic/sigar#48 is resolved.

### DIFF
--- a/lib/ohai/plugins/network_listeners.rb
+++ b/lib/ohai/plugins/network_listeners.rb
@@ -39,6 +39,11 @@ Ohai.plugin(:NetworkListeners) do
       listeners[port][:address] = addr
       begin
         pid = sigar.proc_port(conn.type, port)
+        # workaround for a failure of proc_state to throw
+        # after the first 0 has been supplied to it
+        #
+        # no longer required when hyperic/sigar#48 is fixed
+        throw ArgumentError.new("No such process") if pid == 0
         listeners[port][:pid] = pid
         listeners[port][:name] = sigar.proc_state(pid).name
       rescue


mcquin:
Would love to have specs for this plugin, at least start with a test for this case so we don't regress.

chef-supermarket:
Hi. I am an automated pull request bot named Curry. There are commits in this pull request whose authors are not yet authorized to contribute to Chef Software, Inc. projects or are using a non-GitHub verified email address. To become authorized to contribute, you will need to sign the Contributor License Agreement (CLA) as an individual or on behalf of your company. [You can read more on Chef's blog.](https://www.chef.io/blog/2014/06/23/changes-to-the-contributor-license-agreement-process)

* @werkt

PR #499, opened 2015-03-17
cleanup joyent.rb

bahamas10:
- condense code
- touch up style
- don't stat(2) before open(2)

### DIFF
--- a/lib/ohai/plugins/joyent.rb
+++ b/lib/ohai/plugins/joyent.rb
@@ -25,21 +25,18 @@ Ohai.plugin(:Joyent) do
   depends 'os', 'platform', 'virtualization'
 
   def collect_product_file
-    lines = []
-    if ::File.exists?("/etc/product")
-      ::File.open("/etc/product") do |file|
-        while line = file.gets
-          lines << line
-        end
-      end
+    data = ::File.read("/etc/product") rescue nil
+    if data
+      data.strip.split("\n")
+    else
+      nil
     end
-    lines
   end
 
   def collect_pkgsrc
-    if File.exist?('/opt/local/etc/pkg_install.conf')
-      sm_pkgsrc = ::File.read("/opt/local/etc/pkg_install.conf").split("=")
-      sm_pkgsrc[1].chomp
+    data = ::File.read("/opt/local/etc/pkg_install.conf") rescue nil
+    if data
+      data.strip.split("=", 2).last
     else
       nil
     end
@@ -58,14 +55,14 @@ Ohai.plugin(:Joyent) do
 
       # get zone id unless globalzone
       unless joyent[:sm_uuid] == "global"
-        joyent[:sm_id]            = virtualization[:guest_id]
+        joyent[:sm_id] = virtualization[:guest_id]
       end
 
       # retrieve image name and pkgsrc
       collect_product_file.each do |line|
         case line
         when /^Image/
-          sm_image = line.split(" ") 
+          sm_image = line.split(" ")
           joyent[:sm_image_id] = sm_image[1]
           joyent[:sm_image_ver] = sm_image[2]
         when /^Base Image/


thommay:
This all looks sane to me :+1: @chef/client-core

lib/ohai/plugins/joyent.rb
```diff
@@ -25,24 +25,11 @@
...
   end
 
   def collect_pkgsrc
-    if File.exist?('/opt/local/etc/pkg_install.conf')
-      sm_pkgsrc = ::File.read("/opt/local/etc/pkg_install.conf").split("=")
-      sm_pkgsrc[1].chomp
-    else
-      nil
-    end
+    ::File.read("/opt/local/etc/pkg_install.conf").strip.split("=", 2).last rescue nil
```
danielsdeleo: Can we make this rescue more specific? The way it's written now we'll rescue anything in `.strip.split("=", 2).last` when really we just want to rescue `ENOENT` and `EACCES` on the file read.

lib/ohai/plugins/joyent.rb
```diff
@@ -25,24 +25,11 @@
...
-    lines = []
-    if ::File.exists?("/etc/product")
-      ::File.open("/etc/product") do |file|
-        while line = file.gets
-          lines << line
-        end
-      end
-    end
-    lines
+    ::File.read("/etc/product").strip.split("\n") rescue []
```
danielsdeleo: Same as below, we're rescuing too much here.

lib/ohai/plugins/joyent.rb
bahamas10: @danielsdeleo is it proper for an ohai plugin to have an uncaught exception? i figured the goal was to avoid that

danielsdeleo: Ohai catches all exceptions at a higher layer, which I dislike but we can't change that now. So any uncaught exception in a plugin will be rescued and cause a debug message to be printed, so if this plugin were to fail because, say, we had unexpected content in `/opt/local/etc/pkg_install.conf`, at least you would be able to get some feedback about why you suddenly didn't get the ohai data you expected. Which illustrates why I prefer not to do "blind" rescues like this (except in special cases), it hides or obfuscates the true source of the problem when something goes wrong.

bahamas10: ah ok, i wasn't aware of `ohai` capturing all exceptions.

would it be best to remove the `rescue` statement completely then?  I think it's just as much of a problem if `/opt/local/etc/pkg_install.conf` doesn't exist, as it is if it contains garbage data.

danielsdeleo: You'd be the person to know that better than me. Are we on a fundamentally broken system if `/opt/local/etc/pkg_install.conf ` doesn't exist? Or is it a normal thing?

bahamas10: ah good point, i'll make changes and push up a new commit.

(no, the system is not fundamentally broken in that case, so i'll modify the code to reflect that)

lib/ohai/plugins/joyent.rb
```diff
@@ -25,21 +25,18 @@
...
 
   def collect_product_file
-    lines = []
-    if ::File.exists?("/etc/product")
-      ::File.open("/etc/product") do |file|
-        while line = file.gets
-          lines << line
-        end
-      end
+    data = ::File.read("/etc/product") rescue nil
```
mcquin: You'll need to stub `::File.read("/etc/product")` in the specs. The failures appear to be due to [this line](https://github.com/bahamas10/ohai/blob/dave-1426559693/lib/ohai/plugins/joyent.rb#L62) returning `nil`.

lib/ohai/plugins/joyent.rb
```diff
@@ -25,21 +25,18 @@
...
+    if data
+      data.strip.split("\n")
+    else
+      nil
```
lamont-granquist: if returns nil if the condition is false.

PR #502, opened 2015-03-17
bumping ffi-yajl to pick up 2.x

lamont-granquist:
1.x still works fine, so leaving the lower floor in place.

### DIFF
--- a/ohai.gemspec
+++ b/ohai.gemspec
@@ -17,7 +17,7 @@ Gem::Specification.new do |s|
 
   s.add_dependency "mime-types", "~> 2.0"
   s.add_dependency "systemu", "~> 2.6.4"
-  s.add_dependency "ffi-yajl", "~> 1.1"
+  s.add_dependency "ffi-yajl", ">= 1.1", "< 3.0"
   s.add_dependency "mixlib-cli"
   s.add_dependency "mixlib-config", "~> 2.0"
   s.add_dependency "mixlib-log"


lamont-granquist:
@chef/client-engineers @chef/client-core same as the chef PR

tyler-ball:
:+1: 

Can we stop supporting Ruby 1.9.3 at the next major version bump?

danielsdeleo:
Ruby 1.9.3 blowing up on travis with some encoding thing in ffi-yajl. I think Ruby 2.0 made some changes to assume UTF-8 more places (definitely for source file encodings), and we don't need to support 1.9.3, I think we can drop it now per all the version support RFCs we've gone through. So I'm :+1: on this as long as we remove 1.9.3 from travis also (my preference) or work around the encoding thing (yuk, but ok).

lamont-granquist:
1.9.3 should have been dropped here in 8.0 and chef 12 doesn't run on it anymore, so i think we can drop it in a minor rev.

the ohai thing is going to be a bug someplace in ohai where we're trying to render a non-utf8 character as json and its only blowing up at the end when ffi yajl tries to verify it.  unfortunately the error message is not useful to track down where it blew up.

i still want to patch ffi-yajl to allow forcing the encoding to utf8 and replacing any bad characters with '?' on the fly as it encodes to json, haven't gotten around to that...

PR #504, opened 2015-03-18
ec2 plugin should handle binary userdata too

sean-horn:


thommay:
This looks good to me. :+1: 
@chef/client-core

lib/ohai/plugins/ec2.rb
```diff
@@ -58,6 +59,11 @@ def looks_like_ec2?
         ec2[k] = v
       end
       ec2[:userdata] = self.fetch_userdata
+      #ASCII-8BIT is equivalent to BINARY in this case
+      if ec2[:userdata].encoding.to_s == "ASCII-8BIT"
```
ranjib: are we sure this will cover all the binary contents?

ranjib:
:+1:  i am unsure about the detection though, ASCII-8BIT should cover most. we can revisit this later if we encounter any bug (wondering mime style content, but i think this concern is far fetched :-) )

lamont-granquist:
I'm a little surprised that we get back UTF-8 when it can be transcoded and ASCII-8BIT when it can't, I would have expected to always get back ASCII-8BIT or to get an encoding error from the library when it can't be transcoded.  If the http library works that way, though, then the code looks okay.

I probably would have picked rendering the string as hex rather than Base64.

PR #505, opened 2015-03-18
remove 1.9.3 from travis

lamont-granquist:


lamont-granquist:
we're setting `required_ruby_version = ">= 2.0.0"` so i don't understand how this worked at all...

Issue #501, opened 2015-03-17
net-dhcp is GPL licensed, revert cloudstack contribution


btm:
https://github.com/chef/ohai/commit/0054eb3fdb0c71b2b4aa26bff20a97e9e7d5c129 from #354 added the net-dhcp library which is GPL licensed: https://github.com/syonbori/net-dhcp-ruby/blob/master/lib/net/dhcp.rb#L25. We'll have to revert for now and ship with the Chef 12.1.2 release that we want to come out ASAP, unless @lndbrg can come up with an alternate patch within a day.

PR #506, opened 2015-03-18
Reverting #354 because change bring in GPLv3 dependency

jdmundrawala:
Revert "Merge pull request #354 from lndbrg/cloudstack-support"

This reverts commit ca805e6952eae39bc53113d2da798b1dcd0d51de, reversing
changes made to 0779baf28aa230c134877122dbafa8eacb7b29bd.

Solves #501 

cc @chef/client-engineers


danielsdeleo:
Sad Panda, but :+1:

thommay:
:+1:

lamont-granquist:
breaking change for anyone using this, but the licensing absolutely forces this.

:+1:

jdmundrawala:
@chef/client-engineers i'd like to hold off releasing this until chef 12.2. Any objections?
cc @adamedx @charlesjohnson

lamont-granquist:
nope

thommay:
I think 12.2 is fine.

lndbrg:
I'm no longer at a company where cloudstack is used (yey!) and i do not really know what licenses are acceptable for your libraries. If MIT is acceptable, using this library might be possible:
https://github.com/astounding/dhcp

But I don't really care anymore.

adamedx:
@lndbrg thanks for the response and follow-up on an alternative -- appreciated, and our apologies for not providing this feedback until after it was merged.

btm:
@lndbrg Yes, the MIT licenses is fine.

Most of our projects are Apache-2.0 licensed. Some history of that choice is available [in this blog post](https://www.chef.io/blog/2009/08/11/why-we-chose-the-apache-license/).

The Free Software Foundation believes that linking libraries creates derivatives works, which means the whole project would then have to be GPL licensed if we included that library. Which makes the Apache-2.0 license, and many others, [incompatible](http://www.apache.org/licenses/GPL-compatibility.html). So we cannot use libraries that are GPL licensed, due to the FSF's interpretation of derivative works. 

In general, most open source licenses are compatible. The GPL is somewhat unique and pejoratively referred to as "viral" due to it's requirements on derivatives and interpretation of what is a derivative. A fairly concise list of licenses compatible with the Apache 2.0 is available here: http://apache.org/legal/resolved.html#category-a

lamont-granquist:
We also just can't ship GPLv3 code to some customers.  Period.  Full stop.  License infection or not doesn't matter since they don't want it in their infrastructure at all.  I think we can and still do ship some (L?)GPLv2 code, but prefer MIT/Apache compatible licenses.  I don't think its actually the derivative works clause which causes the biggest issues.

PR #508, opened 2015-03-23
Revert the Solaris portion of #265 to fix #420 temporarily

### LINKED ISSUE #420: Getting null domain for Solaris nodes upgraded from Ohai 6.18 to Ohai 7.x
Moving issue from Chef repo to Ohai repo: https://github.com/opscode/chef/issues/2139

Original:

We are seeing domain results of "null" when running Ohai 7.4.0. This is happening on the Solaris 10 nodes that we are attempting to upgrade from an older version of the Chef client. Here are some relevant snippets from running Ohai before and after the upgrade:

BEFORE: (using Ohai 6.18.0)

"hostname": "abc",
"fqdn": "abc.def.gh",
"domain": "def.gh",
"network": {
"interfaces": {
"lo0": {

AFTER: (Upgraded to the 11.16 client with Ohai 7.4.0) Notice that the “domain” value is null.

"machinename": "abc",
"hostname": "abc",
"fqdn": "abc",
"domain": null,
"keys": {
"ssh": {

Here is the error we see when we run "ohai domain" on a Solaris node with Ohai 7.4.0:

-bash-3.2$ ohai domain
/opt/chef/embedded/lib/ruby/gems/1.9.1/gems/ohai-7.4.0/lib/ohai/system.rb:191:in attributes_print': I cannot find an attribute named domain! (ArgumentError) from /opt/chef/embedded/lib/ruby/gems/1.9.1/gems/ohai-7.4.0/lib/ohai/application.rb:94:inblock in run_application'
from /opt/chef/embedded/lib/ruby/gems/1.9.1/gems/ohai-7.4.0/lib/ohai/application.rb:93:in each' from /opt/chef/embedded/lib/ruby/gems/1.9.1/gems/ohai-7.4.0/lib/ohai/application.rb:93:inrun_application'
from /opt/chef/embedded/lib/ruby/gems/1.9.1/gems/ohai-7.4.0/lib/ohai/application.rb:70:in run' from /opt/chef/embedded/lib/ruby/gems/1.9.1/gems/ohai-7.4.0/bin/ohai:42:in'
from /usr/bin/ohai:23:in load' from /usr/bin/ohai:23:in

This behavior produces an incomplete FQDN value consisting of just the host name and no domain. The behavior may have changed due to a recent mod to the hostname plugin for Ohai?

opscode/ohai@7e129dc

Hoping this can be corrected soon so that we can upgrade the Client and Ohai across our enterprise. Thanks!

btm:
Fixes #420 , for now.

@lamont-granquist @jdmundrawala @scotthain, this look right?

### DIFF
--- a/lib/ohai/plugins/hostname.rb
+++ b/lib/ohai/plugins/hostname.rb
@@ -138,9 +138,23 @@ Ohai.plugin(:Hostname) do
   end
 
   collect_data(:solaris2) do
+    require 'socket'
+
     machinename from_cmd("hostname")
     hostname from_cmd("hostname")
-    fqdn resolve_fqdn
+    fqdn_lookup = Socket.getaddrinfo(hostname, nil, nil, nil, nil, Socket::AI_CANONNAME).first[2]
+    if fqdn_lookup.split('.').length > 1
+      # we recieved an fqdn
+      fqdn fqdn_lookup
+    else
+      # default to assembling one
+      so = shell_out("hostname")
+      h = so.stdout.split($/)[0]
+      so = shell_out("domainname")
+      d = so.stdout.split($/)[0]
+
+      fqdn("#{h}.#{d}")
+    end
     domain collect_domain
   end
 
--- a/spec/unit/plugins/solaris2/hostname_spec.rb
+++ b/spec/unit/plugins/solaris2/hostname_spec.rb
@@ -22,14 +22,14 @@ describe Ohai::System, "Solaris2.X hostname plugin" do
   before(:each) do
     @plugin = get_plugin("hostname")
     allow(@plugin).to receive(:collect_os).and_return(:solaris2)
-    allow(@plugin).to receive(:resolve_fqdn).and_return("kitteh.inurfridge.eatinurfoodz")
     allow(@plugin).to receive(:shell_out).with("hostname").and_return(mock_shell_out(0, "kitteh\n", ""))
-#    Socket.stub(:getaddrinfo).and_return( [["AF_INET", 0, "kitteh.inurfridge.eatinurfoodz", "10.1.2.3", 2, 0, 0]] );
+    allow(Socket).to receive(:getaddrinfo).and_return( [["AF_INET", 0, "kitteh.inurfridge.eatinurfoodz", "10.1.2.3", 2, 0, 0]] );
   end
 
   it_should_check_from("solaris2::hostname", "hostname", "hostname", "kitteh")
 
-  it "should get the fqdn value from #resolve_fqdn" do
+  it "should get the fqdn value from socket getaddrinfo" do
+    expect(Socket).to receive(:getaddrinfo)
     @plugin.run
     expect(@plugin["fqdn"]).to eq("kitteh.inurfridge.eatinurfoodz")
   end


chef-supermarket:
Hi. I am an automated pull request bot named Curry. There are commits in this pull request whose authors are not yet authorized to contribute to Chef Software, Inc. projects or are using a non-GitHub verified email address. To become authorized to contribute, you will need to sign the Contributor License Agreement (CLA) as an individual or on behalf of your company. [You can read more on Chef's blog.](https://www.chef.io/blog/2014/06/23/changes-to-the-contributor-license-agreement-process)

## Non-GitHub Verified Committers

There are 1 commit author(s) whose commits are authored by a non-GitHub verified email address. Chef will have to manually verify that they are authorized to contribute.

lib/ohai/plugins/hostname.rb
```diff
@@ -138,9 +138,23 @@ def sigar_is_available?
   end
 
   collect_data(:solaris2) do
+    require 'socket'
+
     machinename from_cmd("hostname")
     hostname from_cmd("hostname")
-    fqdn resolve_fqdn
+    fqdn_lookup = Socket.getaddrinfo(hostname, nil, nil, nil, nil, Socket::AI_CANONNAME).first[2]
```
lamont-granquist: can we just re-use resolve_fqdn here?  it looks like that has a rescue clause that you probably don't want to drop.

lib/ohai/plugins/hostname.rb
```diff
@@ -138,9 +138,23 @@ def sigar_is_available?
...
+    if fqdn_lookup.split('.').length > 1
```
lamont-granquist: This needs to be protected against `nil?`

lib/ohai/plugins/hostname.rb
```diff
@@ -138,9 +138,23 @@ def sigar_is_available?
...
+      # we recieved an fqdn
+      fqdn fqdn_lookup
+    else
+      # default to assembling one
```
lamont-granquist: We should probably use `hostname` to start with and if it has dots in it, just use that.

On Solaris `hostname` will not have dots, then as a third step we can append the `domainname`.

lamont-granquist:
And this really needs an RFC.  I can work on that, but I've also got the resource_resolver patch that I haven't had any time yet today to work on.

btm:
This is just the revert portion of https://github.com/chef/ohai/issues/420#issuecomment-75631364 to fix people, not the RFC for the long term fix portion.

lamont-granquist:
Yeah, its still a potentially breaking change to the semantics of a public API in a minor version.  Anyone that has been using Chef 12 may have built code that will break when this is reverted.

btm:
@lamont-granquist @scotthain @charlesjohnson @andrewelizondo @jeremiahsnapp @Doug-MacDowell, it looked like there was agreement on #420 that we should revert the fqdn lookup behavior on Solaris for now, and come up with a long-term plan that we could share before we released it. After talking to a few people, that is not the case. I'm closing this PR and we'll look at finding a workaround for those affected by #420 in the next couple of weeks. If you have an input on that solution, please add it to #420 or contact me directly.

lamont-granquist:
I'm actually okay with this change, but with the fixes I mentioned (not just a straight git revert), but think that touching the semantics of these attributes hits the RFC bar (because clearly they're a widely used public interface which has lots of deep implications, like the use of node[:fqdn] to determine the node_name by default, etc).

btm:
Right, we should be clear what we're going to do with node[:fqdn]. We're not going to coalesce an agreement on that today, let alone get an RFC approved, which means this gets pushed past the current release.

Issue #397, opened 2014-08-24
ohai WARN: unable to detect ipaddress, macaddress


ghost:
Ohai: 7.2.4
app chef # ohai > test.ohai
[2014-08-23T05:08:39+03:00] WARN: unable to detect ipaddress
[2014-08-23T05:08:39+03:00] WARN: unable to detect macaddress
app chef # 

it did at least detect the gateway:

```
"network": {
    "interfaces": {
      "eth0:": {
        "type": "eth",
        "number": "0:"
      },
      "lo:": {

      },
      "sit0:": {
        "type": "sit",
        "number": "0:"
      }
    },
    "default_gateway": "172.16.20.1",
    "default_interface": "eth0"
  },
  "counters": {
    "network": {
      "interfaces": {

      }
    }
  },
```

network info : 
```
app chef # ip link 
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT 
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP mode DEFAULT qlen 1000
    link/ether 52:54:00:70:26:0b brd ff:ff:ff:ff:ff:ff
3: sit0: <NOARP> mtu 1480 qdisc noop state DOWN mode DEFAULT 
    link/sit 0.0.0.0 brd 0.0.0.0

app chef # ip addr      
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN 
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP qlen 1000
    link/ether 52:54:00:70:26:0b brd ff:ff:ff:ff:ff:ff
    inet 172.16.20.2/24 brd 172.16.20.255 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::5054:ff:fe70:260b/64 scope link 
       valid_lft forever preferred_lft forever
3: sit0: <NOARP> mtu 1480 qdisc noop state DOWN 
[...]

PR #510, opened 2015-03-24
Update ip binary detection logic to work on Gentoo/CoreOS

tas50:
This is a fix for #397 where any Gentoo based OS does not get IP information.  The root cause is the detection logic for the ip command.  Even though shellout to IP on those platforms would work we first check to see if IP exists by looking for ip in specific paths.  We didn't include /bin/ip where Gentoo stores it.  This simply adds that path which should magically make it all work on Gentoo.  I'm not really sure how to write a test for this, but if anyone has ideas/examples I'm open to writing something.

### DIFF
--- a/lib/ohai/plugins/linux/network.rb
+++ b/lib/ohai/plugins/linux/network.rb
@@ -34,7 +34,7 @@ Ohai.plugin(:Network) do
   end
 
   def iproute2_binary_available?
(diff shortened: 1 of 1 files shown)


tas50:
Any chance this can get reviewed?

jdmundrawala:
:+1:

thommay:
:+1:

Issue #517, opened 2015-03-31
Ohai platform discovery broken on Centos 7.1


tpowell-rms:
A new release of centos-release occurred on March 27. 
Ohai returns "derived" now for the platform value :-\

```
#  rpm -qi centos-release-7-1.1503.el7.centos.2.7.x86_64

Name        : centos-release
Version     : 7
Release     : 1.1503.el7.centos.2.7
Architecture: x86_64
Install Date: Tue 31 Mar 2015 08:12:51 PM UTC
Group       : System Environment/Base
Size        : 34409
License     : GPLv2
Signature   : RSA/SHA256, Fri 27 Mar 2015 09:53:39 PM UTC, Key ID 24c6a8a7f4a80eb5
Source RPM  : centos-release-7-1.1503.el7.centos.2.7.src.rpm
Build Date  : Fri 27 Mar 2015 09:48:00 PM UTC
Build Host  : worker1.bsys.centos.org
Relocations : (not relocatable)
Packager    : CentOS BuildSystem <http://bugs.centos.org>
Vendor      : CentOS
Summary     : CentOS Linux release file
Description :
CentOS Linux release files
```

```
# cat /etc/redhat-release 
Derived from Red Hat Enterprise Linux 7.1 (Source)
```

```
#  ohai platform
[
  "derived"
]
```

```
# ohai --version
Ohai: 7.4.0
```

nathwill:
from the [release notes](http://lists.centos.org/pipermail/centos-announce/2015-March/021006.html):

```
----------
Release file handling

This release splits the /etc/centos-release from /etc/redhat-release to
better indicate the relationship between the two distributions. There
are also changes to the /etc/os-release file to incorporate changes
needed by the new abrt stack.
```

tpowell-rms:
Kwame: Let our powers combine. Earth!
Wheeler: Fire!
Linka: Wind!
Gi: Water!
Ma-Ti: Heart!
Kwame, Wheeler, Linka, Gi, Ma-Ti: Go, Planet!
Captain Planet: By your powers combined, I am Captain Planet.

nathwill:
as noted by @bdwyertech in https://github.com/chef/opscode-omnitruck/pull/86, it looks like this change has been reverted by centos, so the associated PR may not be necessary.

that said, i'd love to see ohai use /etc/os-release as a primary source where it exists, as it's a lot more parseable and less prone to this kind of breakage.

lamont-granquist:
[...]

PR #518, opened 2015-04-01
Update ohai to use /etc/centos-release if present.

nathwill:
this fixes https://github.com/chef/ohai/issues/517 by preferentially using centos-release over redhat-release.

thanks @cmluciano for pointing out where this was at.


nathwill:
closing as the breakage was reverted by centos

PR #524, opened 2015-04-14
Add language scala

cmluciano:


thommay:
LGTM :+1:

cmluciano:
@mcquin good on your side?

mcquin:
LGTM

Should we add a `languages/scala/sbt` attribute, based on the presence of sbt? These aren't packaged together, but they tend to go hand-in-hand.

cmluciano:
Sounds great to me. Should we collect just the sbt info or also say that if sbt is found, we should label this machine as having the language scala installed?

mcquin:
Some people may have scala, but not sbt. 

I'm +1 for having sbt as a subattribute of `languages/scala` (i.e, `provides "languages/scala/sbt"`) which would contain version info (`sbt[:version] = ...`), or be `nil` if sbt is not present on the system (`scala[:sbt] = nil`).

cmluciano:
@mcquin I added a step to check for sbt.

lib/ohai/plugins/scala.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  collect_data(:default) do
+    # Check for scala
+    begin
+      output = nil
+
+      scala = Mash.new
+      so = shell_out("scala -version")
+      if so.exitstatus == 0
+        output =  so.stdout.split
```
mcquin: [nitpick] Looks like you've got two spaces after your `=` sign.

lib/ohai/plugins/scala.rb
```diff
@@ -0,0 +1,47 @@
...
+      end
+    end
+
+    # Check for sbt
+    begin
+      output = nil
+
+      so = shell_out("sbt --version")
+      if so.exitstatus == 0
+        output =  so.stdout.split
```
mcquin: https://github.com/chef/ohai/pull/524/files#r32964251

spec/unit/plugins/scala_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require File.expand_path(File.join(File.dirname(__FILE__), '..', '..', '/spec_helper.rb'))
+
+describe Ohai::System, "plugin scala" do
+
+  let (:plugin) do
```
mcquin: [nitpick] `let(:plugin)`

spec/unit/plugins/scala_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+    plugin = get_plugin("scala").tap do |plugin|
```
mcquin: I think you can drop the `plugin =`.

spec/unit/plugins/scala_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+  def setup_plugin
+    allow(plugin).to receive(:shell_out)
+      .with("scala -version")
+      .and_return(mock_shell_out(0, scala_out, ""))
+    allow(plugin).to receive(:shell_out)
+      .with("sbt --version")
+      .and_return(mock_shell_out(0, sbt_out, ""))
+  end
+
+
```
mcquin: [nitpick] Extra whitespace.

spec/unit/plugins/scala_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+  context " if scala is installed" do
+    before(:each) do
+      setup_plugin
+      plugin.run
+    end
+
+    it "should get the scala version" do
+      expect(plugin).to receive(:shell_out).with("scala -version").and_return(mock_shell_out(0, scala_out, ""))
```
mcquin: Looks like this gets stubbed in the `setup_plugin` method.

spec/unit/plugins/scala_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+      plugin.run
+    end
+
+    it "should set languages[:scala][:version]" do
```
mcquin: I'd suggest removing [the test above](https://github.com/cmluciano/ohai/blob/cml/addscala/spec/unit/plugins/scala_spec.rb#L47-L50) in favor of this one. It appears they test the same thing.

spec/unit/plugins/scala_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+      plugin.run
+    end
+
+
+    it "should get the sbt version" do
+      expect(plugin).to receive(:shell_out).with("sbt --version").and_return(mock_shell_out(0, sbt_out, ""))
+      plugin.run
+    end
+
+    it "should set languages[:sbt][:version]" do
```
mcquin: I'd suggest removing [the test above](https://github.com/cmluciano/ohai/blob/cml/addscala/spec/unit/plugins/scala_spec.rb#L65-L68)j in favor of this one. It appears they test the same thing.

spec/unit/plugins/scala_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+      allow(plugin).to receive(:shell_out)
+        .with("scala -version")
+        .and_return(mock_shell_out(1, scala_out, ""))
+      allow(plugin).to receive(:shell_out)
+        .with("sbt --version")
+        .and_return(mock_shell_out(1, sbt_out, ""))
+    end
+
+    it "should not set the languages[:scala] if scala command fails" do
+      allow(plugin).to receive(:shell_out).with("scala -version").and_return(mock_shell_out(1, scala_out, ""))
```
mcquin: Looks like this has [already been stubbed](https://github.com/cmluciano/ohai/blob/cml/addscala/spec/unit/plugins/scala_spec.rb#L78-L80)

spec/unit/plugins/scala_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+      plugin.run
+      expect(plugin.languages).not_to have_key(:scala)
+    end
+
+    it "should not set the languages[:scala] if scala command fails" do
+      allow(plugin).to receive(:shell_out).with("sbt --version").and_return(mock_shell_out(1, scala_out, ""))
```
mcquin: It looks like this has [already been stubbed](https://github.com/cmluciano/ohai/blob/cml/addscala/spec/unit/plugins/scala_spec.rb#L81-L83).

spec/unit/plugins/scala_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+  context "if scala is not installed" do
+
+    before(:each) do
+      allow(plugin).to receive(:shell_out)
+        .with("scala -version")
+        .and_return(mock_shell_out(1, scala_out, ""))
+      allow(plugin).to receive(:shell_out)
+        .with("sbt --version")
+        .and_return(mock_shell_out(1, sbt_out, ""))
+    end
```
mcquin: It looks like you could throw the `plugin.run` up here and remove it from your test blocks below.

mcquin:
Some minor things, but looking great @cmluciano ! =)

spec/unit/plugins/scala_spec.rb
cmluciano: Do you mean to drop the whole let or just a specific part?

mcquin:
:+1:

cmluciano:
Can we get a few more 👍 from @chef/client-core

mcquin:
We need another review from @chef/client-core

ranjib:
:+1: 
Q: do we know if sbt -version and scala -version will slow down ohai ?

mcquin:
@ranjib Added time appears negligible

|  | `time scala -version` | `time sbt --version` |
|---|---|---|
| **os x w/ scala, sbt installed** | 0.36s user<br>0.07s system<br>0.368 total |  0.32s user<br>0.08s system<br>0.322 total |
| **ubuntu vm w/o scala, sbt installed** | real	0m0.115s<br>user 0m0.038s<br>sys 0m0.071s | real 0m0.115s<br>user 0m0.038s<br>sys 0m0.071s|

spec/unit/plugins/scala_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+
+    it "should set languages[:sbt][:version]" do
+      expect(plugin.languages[:scala][:sbt]).to eql("0.13.8")
+    end
+  end
+
+  context "if scala is not installed" do
+
+    before(:each) do
+      allow(plugin).to receive(:shell_out)
```
danielsdeleo: Unless we've done something special in Ohai, this isn't what happens when scala isn't installed. On my machine:

```
[1] pry(main)> require 'mixlib/shellout'
=> true
[2] pry(main)> include Mixlib
=> Object
[3] pry(main)> c = ShellOut.new("scala -v")
=> <Mixlib::ShellOut#70321135567700: command: 'scala -v' process_status: nil stdout: '' stderr: '' child_pid: nil environment: {} timeout: 600 user:  group:  working_dir:  >
[4] pry(main)> c.run_command
Errno::ENOENT: No such file or directory - scala
from /Users/ddeleo/.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/mixlib-shellout-2.1.0/lib/mixlib/shellout/unix.rb:338:in `exec'
```

I think ohai still catches all failures and ignores them, but it'd be nice to rescue that instead.

mcquin: Huh. I assumed shellout caught these errors, [appears that is not the case](https://github.com/chef/mixlib-shellout/blob/master/lib/mixlib/shellout/unix.rb#L132-L136). Good to know, taking note.

cmluciano: Should I just remove this context?

mcquin: No, don't remove this context. 

Dan's comment shows `shell_out("scala -v")` throws an `Errno::ENOENT`, rather than returning what's been mocked [here](https://github.com/cmluciano/ohai/blob/cml/addscala/spec/unit/plugins/scala_spec.rb#L69). The spec should be updated to reflect that behavior.

mcquin:
LGTM

btm:
What are the two begin/end blocks for?

Does it make sense if we can't find scala to return rather than look for sbt too?

lamont-granquist:
Yeah the begin/end should go and the initial check should probably just be a shell_out! and let the exception abort the plugin.

cmluciano:
Ok begin/end have been removed. Just waiting on CI :)

cmluciano:
Failures seemed related to style, fixed up

mcquin:
:+1: @chef/client-core

lib/ohai/plugins/scala.rb
```diff
@@ -0,0 +1,43 @@
...
+Ohai.plugin(:Scala) do
+  provides "languages/scala", "languages/scala/sbt"
+
+  depends "languages"
+
+  collect_data(:default) do
+    # Check for scala
+    output = nil
+
+    scala = Mash.new
```
tas50: This needs to go after the check for the exist status.  Otherwise we get an empty hash when scala isn't around.

mcquin: Ohai won't add `attribute` to its data until you do `attribute Mash.new`.

tas50: Ignore me

PR #526, opened 2015-04-21
Move windows drivers out of kernel

jdmundrawala:
With the introduction of minimal-ohai in chef, we don't want to
grab a bunch of garbage

cc @danielsdeleo @btm @smurawski @thommay @adamedx 

This needs to go out with minimal-ohai so that it's not a breaking change

### DIFF
--- a/lib/ohai/plugins/kernel.rb
+++ b/lib/ohai/plugins/kernel.rb
@@ -183,23 +183,5 @@ Ohai.plugin(:Kernel) do
 
     kernel[:machine] = machine_lookup("#{kernel[:cs_info][:system_type]}")
 
-    kext = Mash.new
-    pnp_drivers = Mash.new
-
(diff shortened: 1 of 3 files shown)


adamedx:
Yeah, this should make Ohai twice as fast on Windows, right @jdmundrawala ?

jdmundrawala:
minimal ohai + this will be significantly faster. This patch alone shaves off a little over a second on my box.

danielsdeleo:
:+1: The change looks good to me. Can you explain "This needs to go out with minimal-ohai so that it's not a breaking change" though?

jdmundrawala:
So if we were to ship minimal-ohai without this, then removing this from kernel could break people expecting that it be loaded even with minimal-ohai.

danielsdeleo:
Gotcha.

PR #535, opened 2015-05-02
Lazy load ohai plugins

jaym:
This is totally a hack, but I wanted to talk about if we want to start doing things like this now that we have minimal-ohai in chef

Some implications here are that if you don't access an attribute, when they wont (I hope, have not tested) make it into the node save.

cc @chef/client-core

### DIFF
--- a/lib/ohai/mash.rb
+++ b/lib/ohai/mash.rb
@@ -56,7 +56,9 @@ class Mash < Hash
   # @details [Alternatives]
   #   If constructor is a Hash, a new mash will be created based on the keys of
   #   the hash and no default value will be set.
-  def initialize(constructor = {})
+  def initialize(constructor = {}, path=nil, &loader)
+    @path = path
+    @loader = loader
     if constructor.is_a?(Hash)
       super()
       update(constructor)
@@ -71,10 +73,27 @@ class Mash < Hash
   #   If key is a Symbol and it is a key in the mash, then the default value will
   #   be set to the value matching the key.
   def default(key = nil)
-    if key.is_a?(Symbol) && include?(key = key.to_s)
+    if key.is_a?(Symbol) 
+      key = key.to_s
+    end
+
+    if include?(key)
       self[key]
     else
-      super
+      if loader.nil?
+        super
+      else
+        begin
+          loader.call(path_to(key))
+        rescue
+          puts "Error"
+        end
+        if include?(key)
+          self[key]
+        else
+          super
+        end
+      end
     end
   end
 
@@ -88,7 +107,23 @@ class Mash < Hash
   # @see Mash#convert_key
   # @see Mash#convert_value
   def []=(key, value)
-    regular_writer(convert_key(key), convert_value(value))
+    regular_writer(convert_key(key), convert_value(key, value))
+  end
+
+  def path
+    @path
+  end
+
+  def path=(p)
+    @path = p
+  end
+
+  def loader
+    @loader
+  end
+
+  def loader=(l)
+    @loader = l
   end
 
   # @param other_hash<Hash>
@@ -97,7 +132,7 @@ class Mash < Hash
   #
   # @return [Mash] The updated mash.
   def update(other_hash)
-    other_hash.each_pair { |key, value| regular_writer(convert_key(key), convert_value(value)) }
+    other_hash.each_pair { |key, value| regular_writer(convert_key(key), convert_value(key, value)) }
     self
   end
 
@@ -174,7 +209,7 @@ class Mash < Hash
 
   # @return [Mash] Convert a Hash into a Mash
   # The input Hash's default value is maintained
-  def self.from_hash(hash)
+  def self.from_hash(key, hash)
     mash = Mash.new(hash)
     mash.default = hash.default
     mash
@@ -199,13 +234,26 @@ class Mash < Hash
   #   their Mash equivalents.
   #
   # @api private
-  def convert_value(value)
+  def convert_value(key=nil, value)
     if value.class == Hash
-      Mash.from_hash(value)
+      convert_value(key, Mash.from_hash(key, value))
+    elsif value.class == Mash
+      value.path = path_to(key)
+      value.loader = @loader
+      value
     elsif value.is_a?(Array)
       value.collect { |e| convert_value(e) }
     else
       value
     end
   end
+
+  def path_to(key)
+    if path.nil?
+      key
+    else
+      [path, key].join("/")
+    end
+  end
+
 end
--- a/lib/ohai/system.rb
+++ b/lib/ohai/system.rb
@@ -43,7 +43,9 @@ module Ohai
     end
 
     def reset_system
-      @data = Mash.new
+      @data = Mash.new({}, nil) do |key|
+        require_plugin(key) if key.is_a? String
+      end
       @provides_map = ProvidesMap.new
 
       @v6_dependency_solver = Hash.new


jaymzh:
Out of curiosity, if that's your goal, why not use whitelist_attributes which is part of core chef now?

jdmundrawala:
Somebody reported a bug that minimal ohai was too minimal. That's the only
thing this should solve. Not sure about whitelist attributes, will have to
take a look.

jdmundrawala:
Also, load may be the wrong word. I think lazy running of plugins might be a better desecription. Here's an example of me using it:
```ruby
jaym@wormulon ~/workspace/ohai[default]  [jdm/lazy]
± % bundle exec pry                                                                                                                                                                                                                                                         !5742
[1] pry(main)> require 'ohai'
=> true
[2] pry(main)> ohai = Ohai::System.new
=> #<Ohai::System:0x00000001e69bb8
 @data={},
 @loader=#<Ohai::Loader:0x00000001e695f0 @controller=#<Ohai::System:0x00000001e69bb8 ...>, @v6_plugin_classes=[], @v7_plugin_classes=[]>,
 @plugin_path="",
 @provides_map=#<Ohai::ProvidesMap:0x00000001e69758 @map={}>,
 @runner=#<Ohai::Runner:0x00000001e694b0 @provides_map=#<Ohai::ProvidesMap:0x00000001e69758 @map={}>, @safe_run=true>,
 @v6_dependency_solver={}>
[3] pry(main)> ohai.load_plugins
=> [Ohai::NamedPlugin::Mono,
 Ohai::NamedPlugin::Java,
 Ohai::NamedPlugin::Platform,
 Ohai::NamedPlugin::Uptime,
 Ohai::NamedPlugin::Erlang,
 Ohai::NamedPlugin::Joyent,
 Ohai::NamedPlugin::Network,
 Ohai::NamedPlugin::Filesystem,
 Ohai::NamedPlugin::Memory,
 Ohai::NamedPlugin::Virtualization,
 Ohai::NamedPlugin::CPU,
 Ohai::NamedPlugin::Eucalyptus,
 Ohai::NamedPlugin::Lua,
 Ohai::NamedPlugin::Hostname,
 Ohai::NamedPlugin::OS,
 Ohai::NamedPlugin::NetworkAddresses,
 Ohai::NamedPlugin::Go,
 Ohai::NamedPlugin::DigitalOcean,
 Ohai::NamedPlugin::Openstack,
 Ohai::NamedPlugin::Chef,
 Ohai::NamedPlugin::Languages,
 Ohai::NamedPlugin::Rust,
 Ohai::NamedPlugin::Kernel,
 Ohai::NamedPlugin::CloudV2,
 Ohai::NamedPlugin::IpScopes,
 Ohai::NamedPlugin::Command,
 Ohai::NamedPlugin::Keys,
 Ohai::NamedPlugin::PHP,
 Ohai::NamedPlugin::Drivers,
 Ohai::NamedPlugin::Powershell,
 Ohai::NamedPlugin::Passwd,
 Ohai::NamedPlugin::Python,
 Ohai::NamedPlugin::C,
 Ohai::NamedPlugin::EC2,
 Ohai::NamedPlugin::VirtualizationInfo,
 Ohai::NamedPlugin::Perl,
 Ohai::NamedPlugin::InitPackage,
 Ohai::NamedPlugin::Ohai,
 Ohai::NamedPlugin::Rackspace,
 Ohai::NamedPlugin::LSB,
 Ohai::NamedPlugin::Mdadm,
 Ohai::NamedPlugin::BlockDevice,
 Ohai::NamedPlugin::DMI,
 Ohai::NamedPlugin::Linode,
 Ohai::NamedPlugin::SSHHostKey,
 Ohai::NamedPlugin::Cloud,
 Ohai::NamedPlugin::Groovy,
 Ohai::NamedPlugin::GCE,
 Ohai::NamedPlugin::PS,
 Ohai::NamedPlugin::SystemProfile,
 Ohai::NamedPlugin::NetworkListeners,
 Ohai::NamedPlugin::RootGroup,
 Ohai::NamedPlugin::Zpools,
 Ohai::NamedPlugin::Nodejs,
 Ohai::NamedPlugin::Ruby,
 Ohai::NamedPlugin::Azure,
 Ohai::NamedPlugin::Elixir,
 Ohai::NamedPlugin::OhaiTime,
 Ohai::NamedPlugin::NetworkRoutes]
[4] pry(main)> ohai.data
=> {}
[5] pry(main)> ohai[:languages][:perl]
=> {"version"=>"5.20.2", "archname"=>"x86_64-linux-thread-multi"}
[6] pry(main)> ohai.data
=> {"languages"=>{"perl"=>{"version"=>"5.20.2", "archname"=>"x86_64-linux-thread-multi"}}}
```

jdmundrawala:
https://github.com/chef/chef/issues/3304

jaymzh:
So what we do is just run every plugin and then use `whitelist_node_attrs` in `client.rb` to prevent 99% of that going back to the server. Because we don't want it on our server, but we probably DO want to know about it for the run. I imagine most people want more of it on their chef server than we do, but limiting which Ohai plugins run to limit what gets back to your chef-server is the wrong approach... the two are different and have different knobs.

jdmundrawala:
Yea, this definetly breaks the case where you want all of stuff to go into your chef server. Breaking `whitelist_node_attrs` could be easily solved by just running any plugins before a node save that were in the whitelist that have not already run. However, the default behavior in chef is to run all the plugins at startup. This would only affect runs with `--minimal-ohai` (I think).

jaymzh:
Er. I think one of us is misunderstanding the other.

* If an Ohai plugin is causing your system problems, you shouldn't run it
* If you are sending too much data back to the server, use whitelist_node_attrs to limit that data - do **not** disable plugins
* If you are doing testing, running on embedded platforms where most of Ohai doesn't make as much sense, then use `--minimal-ohai`
* Most people probably should not use `--minimal-ohai` - it was written, afaik, to make testing Chef easier, so that a bunch of random system stuff didn't have to be mocked. If it works for you, great, but if the only reason you're running it is to limit data going back to your chef-server, you're using the wrong tool, the right tool is `whitelist_node_attrs`.

Is there a use-case I'm not understanding?

danielsdeleo:
@jaymzh You're correct about the original intent of `--minimal-ohai`. I implemented it originally just to speed up our integration tests, where the extra couple of seconds across many client runs added up to several minutes when running the entire suite. I think end users are interested in an opt-in mode of ohai for the same client-side performance reasons, but `--minimal-ohai` doesn't provide a good UX for that right now since you have to manually run each ohai plugin you use.

jaymzh:
If that's the desire, then perhaps we make an `Ohai::Config[:enabled_plugins]` which can be used in the opposite manner of `:disabled_plugins`. I don't want to just hack onto features people are mis-using, I want to understand the use-case and build the right feature. Or we go down the attribute rabbit hole. :)

jdmundrawala:
`Ohai::Config[:enable_plugins]` seems reasonable, but I guess my question is why make people tell us explicitly all the things they want (I'm sure there is a valid story about why this is important that I'm not thinking of). The ohai plugins tell us what they provide, so we should be able to find the things the user has asked for.

As an example, the powershell, dsc, and service providers use ohai data. The use of that ohai data is an implementation detail, and now we've required to person using that resource to know that. Can we/ should we provide a way for resource providers to make this explicit?

lamont-granquist:
One of the things that we've always discussed is that ideally ohai should only run the plugins that the cookbooks really need.  This goes beyond what whitelisting node attributes does in reducing the server load and also avoids ohai doing the work to collect the attributes and of having all kinds of unused node attributes in memory.  You can do this the fussy way with manually enabling/disabling plugins via config, but it would be substantially more maintainable to just have cookbooks specify their attribute dependencies.

danielsdeleo:
Should we take this to an RFC? Looks like we have 3 proposals for opt-in ohai behavior:

#### Config

Users configure (in client.rb or whatever) the plugins they want.

#### Metadata

Cookbooks declare what plugins they want explicitly in metadata.

#### Lazy/Auto Load

Cookbooks declare what plugins they want implicitly by accessing the data.

lamont-granquist:
Yeah it probably is an RFC level change.  I like Lazy/AutoLoad if we can make it work since its automatic and less typing.  We could use metadata as well if there's any gaps.

mcquin:
Closing for now, as we've decided this requires an RFC.

PR #544, opened 2015-05-21
add support for Wind River Linux and Cisco's Nexus platforms

mattray:


mattray:
@lamont-granquist the Travis test is failing on a unicode character being emitted when 'bundle exec ohai' runs on Ruby 2.0. The ohai output is causing ffi_yajl to choke, I don't think it has to do with my patch. Could you check it out please?

ryancragun:
Looks like /proc/cpuinfo is returning something funky on travis that cant be encoded

lamont-granquist:
Yeah the tests actually pass, the ohai run itself that travis invokes is failing due to UTF-8 in /proc/cpuinfo, which is a whole lotta lolwut, but i'll fix that in an incoming PR...

lib/ohai/plugins/linux/platform.rb
```diff
@@ -59,6 +59,15 @@ def get_redhatish_version(contents)
       contents = File.read("/etc/parallels-release").chomp
       platform get_redhatish_platform(contents)
       platform_version contents.match(/(\d\.\d\.\d)/)[0]
+    elsif File.exists?('/etc/os-release')
```
danielsdeleo: Do we know if this file will exist on any other platforms? Could break detection for those, if so. Probably safer to put this last.

mattray: It needs to go before the Red Hat detection because Wind River has a /etc/redhat file despite not being a Red Hat-based distro. I could make it smarter and put the logic inside the Red Hat detection.

danielsdeleo:
Travis is passing now that ffi-yajl 2.2 is out.

lib/ohai/plugins/linux/platform.rb
lamont-granquist: This is the new standard for this release information in the future, so you will break the world if you put this first and don't have detection for everyone else in here.

mattray:
@lamont-granquist @danielsdeleo updated the patch to make it safer.

lib/ohai/plugins/linux/platform.rb
```diff
@@ -60,9 +60,20 @@ def get_redhatish_version(contents)
...
       platform_version contents.match(/(\d\.\d\.\d)/)[0]
     elsif File.exists?("/etc/redhat-release")
-      contents = File.read("/etc/redhat-release").chomp
-      platform get_redhatish_platform(contents)
-      platform_version get_redhatish_version(contents)
+      if File.exists?('/etc/os-release') # check if Cisco
+      # don't clobber existing os-release properties, point to a different cisco file
+        contents = {}
+        File.read('/etc/os-release').split.collect {|x| x.split('=')}.each {|x| contents[x[0]] = x[1]}
+        if File.exists?(contents['CISCO_RELEASE_INFO'])
```
danielsdeleo:
```
[1] pry(main)> File.exists?(nil)
TypeError: no implicit conversion of nil into String
from (pry):1:in `exists?'
```

Probably should guard that case.

danielsdeleo: Also, we should use `exist?` instead of `exists?`, which is deprecated. I think it will likely never actually be removed, but it spits out a warning if you run `ruby -w`. I'm slowly trying to fix up warnings and enable them in the tests in our codebase as I can.

danielsdeleo:
:+1: once the `File.exists?(nil)` thing is taken care of.

mattray:
@danielsdeleo added the nil check

danielsdeleo:
Cool, :+1:

PR #548, opened 2015-05-26
force ohai output to UTF-8

lamont-granquist:
replaces invalid characters in the rendered JSON and avoids throwing
an exception at that stage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Ohai Changelog
 
+## Unreleased
+
+* [PR #548](https://github.com/chef/ohai/548):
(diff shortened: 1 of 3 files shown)


lamont-granquist:
@chef/client-core this will fix the travis failures.

should probably have a spec for this but i'm busy shaving other yaks today.

jdmundrawala:
:+1: 
It'd be nice if you could put what was causing this to break (if you know) into the commit message.

lamont-granquist:
None of the strings made much sense.  One was coming from /proc/cpuinfo but the non-UTF8 chars made no sense.

PR #552, opened 2015-06-05
Adds a --profile option to Ohai

pburkholder:
Simple add of millisecond profiling to ohai, as suggested by @johnbellone

``` bash
ohai -p cpu
...
{
  "real": 4,
  "total": 8,
  "mhz": 2800
}
{
  "Profile": {
    "CPU": 48
  }
}
```

or with out the attributes:

```
ohai -p
{
  "snip": true,
  "current_user": "pburkholder",
  "root_group": "wheel"
}
{
  "Profile": {
    "CPU": 91,
    "Filesystem": 16,
    "Kernel": 127,
    "Drivers": 0,
    "Memory": 12,
    "Network": 80,
    "NetworkRoutes": 10,
    "NetworkListeners": 38,
    "NetworkAddresses": 1,
    "Platform": 13,
    "Uptime": 17,
    "Virtualization": 1,
    "Joyent": 0,
    "VirtualizationInfo": 0,
    "Azure": 0,
    "C": 146,
    "Elixir": 5,
    "Erlang": 3,
    "Go": 3,
    "Groovy": 4,
    "Java": 254,
    "Languages": 0,
    "Lua": 21,
    "Mono": 4,
    "Nodejs": 20,
    "Perl": 24,
    "PHP": 121,
    "Powershell": 1,
    "Python": 55,
    "Ruby": 995,
    "Rust": 5,
    "Chef": 1,
    "Ohai": 0,
    "Cloud": 99,
    "CloudV2": 0,
    "Command": 0,
    "PS": 0,
    "SystemProfile": 821,
    "DigitalOcean": 0,
    "DMI": 6,
    "EC2": 0,
    "Eucalyptus": 0,
    "OS": 0,
    "GCE": 0,
    "Hostname": 18,
    "InitPackage": 1,
    "IpScopes": 0,
    "Keys": 0,
    "SSHHostKey": 1,
    "Linode": 0,
    "BlockDevice": 0,
    "LSB": 0,
    "Mdadm": 0,
    "OhaiTime": 0,
    "Openstack": 0,
    "Passwd": 26,
    "Rackspace": 0,
    "RootGroup": 1,
    "Zpools": 0
  }
}
```

### DIFF
--- a/lib/ohai/application.rb
+++ b/lib/ohai/application.rb
@@ -94,17 +94,15 @@ class Ohai::Application
   def run_application
     ohai = Ohai::System.new
     ohai.all_plugins(@attributes)
+    ohai.data['ohai'] = ohai.profile
 
     if @attributes
+      @attributes << 'ohai' if Ohai::Config[:profiling]
       @attributes.each do |a|
         puts ohai.attributes_print(a)
       end
     else
       puts ohai.json_pretty_print
-    end
-    if Ohai::Config[:profiling]
-      puts ohai.json_pretty_print(
-        { Elapsed_millisecs_by_plugin: ohai.profile} )
     end
   end
 
--- a/lib/ohai/application.rb
+++ b/lib/ohai/application.rb
@@ -56,6 +56,11 @@ class Ohai::Application
     :proc         => lambda {|v| puts "Ohai: #{::Ohai::VERSION}"},
     :exit         => 0
 
+  option :profiling,
+    :short        => "-p",
+    :long         => "--profiling",
+    :boolean      => true
+
   def initialize
     super
 
@@ -95,6 +100,9 @@ class Ohai::Application
       end
     else
       puts ohai.json_pretty_print
+    end
+    if Ohai::Config[:profiling]
+      puts ohai.json_pretty_print({ Profile: ohai.profile} )
     end
   end
 
--- a/lib/ohai/application.rb
+++ b/lib/ohai/application.rb
@@ -59,6 +59,7 @@ class Ohai::Application
   option :profiling,
     :short        => "-p",
     :long         => "--profiling",
+    :description  => "Show elapsed time per plugin",
     :boolean      => true
 
   def initialize
@@ -102,7 +103,8 @@ class Ohai::Application
       puts ohai.json_pretty_print
     end
     if Ohai::Config[:profiling]
-      puts ohai.json_pretty_print({ Profile: ohai.profile} )
+      puts ohai.json_pretty_print(
+        { Elapsed_millisecs_by_plugin: ohai.profile} )
     end
   end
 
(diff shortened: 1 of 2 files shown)


pburkholder:
Current output is not valid JSON:

```
$ bin/ohai -p root_group
[
  "wheel"
]
{
  "Elapsed_millisecs_by_plugin": {
    "RootGroup": 1
  }
}
```

Not sure if anyone cares.

lib/ohai/application.rb
```diff
@@ -96,6 +102,10 @@ def run_application
     else
       puts ohai.json_pretty_print
     end
+    if Ohai::Config[:profiling]
+      puts ohai.json_pretty_print(
```
thommay: I'd like to see this go to stderr, in the same way we (well, Dan) changed knife to output statistics to stderr and "real" output to stdout.

pburkholder:
I've changed the sense of this update to add ohai profile data to `ohai.data` since these timings are, in a sense, automatic attributes of a node. So if you need to, you could add Chef recipe code like 

```ruby
if node['ohai']['Passwd'] > 1000 
  # do something because passwd plugin takes 'too long'
end
```

So, the profile info will be reported at the end of every run by default. Further, these timing will be stored with the node in case you need to access them later.

The `-p` option controls only if the the timing are emitted when you specify attributes on the CLI.  E.g.:

```
$ bundle exec ohai current_user
[
  "pburkholder"
]
```
```
$ bundle exec ohai current_user -p
  "pburkholder"
]
{
  "Passwd": 16
}
```

pburkholder:
@thommay So thanks for the comment -- I took your idea and ran a different direction, and decided the statistics are 'real' data. What do you think?

pburkholder:
@thommay Anything here besides adding tests? Or do we need to bring others into the conversation?

thommay:
yeah, tests are essential. We also need to make sure this doesn't result in any unanticipated stomping on plugin data

thommay:
hey @pburkholder, any joy with adding those tests?

pburkholder:
@thommay It fell of my radar. Back on now.

lamont-granquist:
My only concern is that there might be many ways to profile ohai --

- ruby profiling with ruby-prof
- memory profiling
- these kinds of timing tests
- instrumentation linked into ranjib's RFC

So `--profile` might be a bit too generic of a flag, which is why i went with `--profile-ruby` for the ruby profiling in chef-client...

jaym:
How about instead of --profile, we just check if we're in debug mode, and if we are, we print the timings after the plugins have been loaded

lamont-granquist:
makes sense to me...

thommay:
@jaym sounds like a plan to me

tas50:
I'm closing this out in favor of simple timing data in debug mode

tas50:
https://github.com/chef/ohai/pull/1056 implements the suggested timing in debug mode from this PR

PR #555, opened 2015-06-10
drop ruby 2.0 on travis in line with chef

thommay:
@chef/client-core


lamont-granquist:
You can drop 7-stable as well.  7.4-stable is chef 11.x and master is chef 12.x

danielsdeleo:
:+1:

lamont-granquist:
yeah :+1: as well

PR #559, opened 2015-06-17
Provide a new and improved linux filesystem plugin

jaymzh:
Benefits:
* Can handle multiple virtual FSes with the same "device" (none)
* Can handle a device mounted multiple places
* Is resilient to mtab being a symlink to /proc/mounts
* Provides multiple views for ease of use
* Provides a mechanism that has gauranteed lack of overwritten data
* Forks far fewer times than it's predecessor

How it works:

* It provides 3 views:
  * by_pair is the primary one and what most users are expected to use. It
   gaurantees no loss of data from conflict and has an entry in the hash for
   each pair of $device,$mountpoint (or "$device," for unmounted devices).
  * by_device a similar layout to the original filesystem plugin which is
   indexed by device, except that 'mount' entry is now 'mounts' and is an array.
   While this solves many of the problems users can encounter by having
   /etc/mtab be a symlink to /proc/mounts it can still have data loss due to
   different mount options, or multiple virtualfs mounts with the same fake
   device name.
  * by_mount similar to the above but indexed by mountpoint. Won't include
   unmounted filesystems, of course.

### DIFF
--- a/lib/ohai/plugins/linux/filesystem2.rb
+++ b/lib/ohai/plugins/linux/filesystem2.rb
@@ -1,5 +1,7 @@
 #
 # Author:: Phil Dibowitz <phil@ipom.com>
+# Author:: Adam Jacob <adam@chef.io>
+# Copyright:: Copyright (c) 2008 Opscode, Inc.
 # Copyright:: Copyright (c) 2015 Facebook, Inc.
 # License:: Apache License, Version 2.0
 #
@@ -85,7 +87,6 @@ Ohai.plugin(:Filesystem2) do
 
   collect_data(:linux) do
     fs = Mash.new
-    have_lsblk = File.executable?('/bin/lsblk')
 
     # Grab filesystem data from df
     so = shell_out("df -P")
@@ -137,10 +138,11 @@ Ohai.plugin(:Filesystem2) do
     end
 
     have_lsblk = File.exist?('/bin/lsblk')
-    if have_lsblk = File.exist?('/bin/lsblk')
+    if have_lsblk
       cmd = 'lsblk -n -P -o NAME,UUID,LABEL,FSTYPE'
     else
-      cmd = '"blkid'
+      # CentOS5 and other platforms don't have lsblk
+      cmd = 'blkid'
     end
 
     so = shell_out(cmd)
--- a/lib/ohai/plugins/linux/filesystem2.rb
+++ b/lib/ohai/plugins/linux/filesystem2.rb
@@ -90,7 +90,7 @@ Ohai.plugin(:Filesystem2) do
 
     # Grab filesystem data from df
     so = shell_out("df -P")
-    so.stdout.lines do |line|
+    so.stdout.each_line do |line|
       case line
       when /^Filesystem\s+1024-blocks/
         next
@@ -108,7 +108,7 @@ Ohai.plugin(:Filesystem2) do
     
     # Grab filesystem inode data from df
     so = shell_out("df -iP")
-    so.stdout.lines do |line|
+    so.stdout.each_line do |line|
       case line
       when /^Filesystem\s+Inodes/
         next
@@ -126,7 +126,7 @@ Ohai.plugin(:Filesystem2) do
 
     # Grab mount information from /bin/mount
     so = shell_out("mount")
-    so.stdout.lines do |line|
+    so.stdout.each_line do |line|
       if line =~ /^(.+?) on (.+?) type (.+?) \((.+?)\)$/
         key = "#{$1},#{$2}"
         fs[key] = Mash.new unless fs.has_key?(key)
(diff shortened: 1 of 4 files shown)


jaymzh:
BTW I made me the copyright owner since it's ostensibly new... there's tons of common code... I'm fine leaving it with Adam/Chef if people feel that's the right call. It was sorta on the border.

jaymzh:

danielsdeleo:
Your goals all sound reasonable to me. Needs tests, of course.

As for the copyright issue, I think what we normally do is have both names on the file.

lib/ohai/plugins/linux/filesystem2.rb
```diff
@@ -0,0 +1,216 @@
...
+        fs[key] = Mash.new unless fs.has_key?(key)
+        fs[key][:device] = $1
+        fs[key][:mount] = $2
+        fs[key][:fs_type] = $3
+        fs[key][:mount_options] = $4.split(",")
+      end
+    end
+
+    have_lsblk = File.exist?('/bin/lsblk')
+    if have_lsblk = File.exist?('/bin/lsblk')
```
jdmundrawala: you have `have_lsblk` multiple times:
line 140, 139, and 88

lib/ohai/plugins/linux/filesystem2.rb
```diff
@@ -0,0 +1,216 @@
...
+      cmd = 'lsblk -n -P -o NAME,UUID,LABEL,FSTYPE'
+    else
+      cmd = '"blkid'
```
jdmundrawala: Are there any known cases where this fallback is needed? Maybe a comment if so.

lib/ohai/plugins/linux/filesystem2.rb
```diff
@@ -0,0 +1,216 @@
...
+          end
+        end
+      end
+    end
+
+    # Grab any missing mount information from /proc/mounts
+    if File.exist?('/proc/mounts')
+      mounts = ''
+      # Due to https://tickets.opscode.com/browse/OHAI-196
+      # we have to non-block read dev files. Ew.
```
jdmundrawala: does this apply to

```ruby
File.open('/proc/mounts') do
  ...
end
```
as well?

lib/ohai/plugins/linux/filesystem2.rb
jaymzh: yeah, centos5 doesn't have lsblk.

lib/ohai/plugins/linux/filesystem2.rb
jaymzh: whoops

lib/ohai/plugins/linux/filesystem2.rb
jaymzh: That can block - we must call `read_nonblock()` - this code is copied wholesale from the original filesystem plugin, fwiw.

jaymzh:
OK, updated. @danielsdeleo - two authors makes sense, not sure about two copyrights, is this what you had in mind?
I'll add tests next.

jdmundrawala:
:+1:

jaymzh:
OK, tests. Which revealed several bugs - yay for testing!!

I think this is good to go now.

danielsdeleo:
> two authors makes sense, not sure about two copyrights, is this what you had in mind?
Yep, I'm not a super expert on this, but the idea is you'd need consent from both copyright holders if you were to hypothetically relicense as GPL or something. tl;dr, this is good.

jaymzh:
That makes sense @danielsdeleo, thanks. You good with the code too?

jaymzh:
Oh I suppose I should add some docs... will do now.

lib/ohai/plugins/linux/filesystem2.rb
```diff
@@ -0,0 +1,218 @@
...
+    end
+    view
+  end
+
+  collect_data(:linux) do
+    fs = Mash.new
+
+    # Grab filesystem data from df
+    so = shell_out("df -P")
+    so.stdout.lines do |line|
```
danielsdeleo: I've been looking at enabling warnings for our tests and this usage emits a warning for being deprecated; it's suggested to use `each_line` instead. That said, there's a fair bit of cleanup to do to make the code warning free (including fixing the same usage elsewhere) so it's no big deal to leave it in.

danielsdeleo:
:+1: the warning thing would be nice to have fixed up front but it won't be difficult to clean it up later.

jaymzh:
Updated per warnings. With 2 +1's I'll merge once tests pass. Thanks!

jdmundrawala:
I don't think the merge for the changelog did what you want

jaymzh:
Wow, git your drunk, go home. LOL

Thanks @jdmundrawala - I'll push a fix now.

jaymzh:
Doc merge fixed in #563

PR #564, opened 2015-06-24
Port new filesystem2 plugin to darwin.

jaymzh:
This is just a port of #559 to Darwin. See that PR for the details.

### DIFF
--- a/lib/ohai/plugins/darwin/filesystem2.rb
+++ b/lib/ohai/plugins/darwin/filesystem2.rb
@@ -1,0 +1,103 @@
+#
+# Author:: Benjamin Black (<bb@opscode.com>)
+# Copyright:: Copyright (c) 2009 Opscode, Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+# 
+#     http://www.apache.org/licenses/LICENSE-2.0
+# 
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+Ohai.plugin(:Filesystem2) do
+  provides "filesystem2"
+
+  def generate_device_view(fs)
+    view = {}
+    fs.each_value do |entry|
+      view[entry[:device]] = Mash.new unless view[entry[:device]]
+      entry.each do |key, val|
+        next if ['device', 'mount'].include?(key)
+        view[entry[:device]][key] = val
+      end
+      if entry[:mount]
+        view[entry[:device]][:mounts] = [] unless view[entry[:device]][:mounts]
+        view[entry[:device]][:mounts] << entry[:mount]
+      end
+    end
+    view
+  end
+
+  def generate_mountpoint_view(fs)
+    view = {}
+    fs.each_value do |entry|
+      next unless entry[:mount]
+      view[entry[:mount]] = Mash.new unless view[entry[:mount]]
+      entry.each do |key, val|
+        next if key == 'mount'
+        view[entry[:mount]][key] = val
+      end
+    end
+    view
+  end
+
+  collect_data(:darwin) do
+    fs = Mash.new
+
+
+    block_size = 0
+    # on new versions of OSX, -i is default, on old versions it's not, so
+    # specifying it gets consistent output
+    so = shell_out("df -i")
+    so.stdout.each_line do |line|
+      case line
+      when /^Filesystem\s+(\d+)-/
+        block_size = $1.to_i
+        next
+      when /^(.+?)\s+(\d+)\s+(\d+)\s+(\d+)\s+(\d+\%)\s+(\d+)\s+(\d+)\s+(\d+%)\s+(.+)$/
+        key = "#{$1},#{$9}"
+        fs[key] = Mash.new
+        fs[key][:block_size] = block_size
+        fs[key][:device] = $1
+        fs[key][:kb_size] = ($2.to_i / (1024 / block_size)).to_s
+        fs[key][:kb_used] = ($3.to_i / (1024 / block_size)).to_s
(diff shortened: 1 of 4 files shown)


jaymzh:
CC @chef/client-core and @chef/client-maintainers

tyler-ball:
:+1:

lib/ohai/plugins/darwin/filesystem2.rb
```diff
@@ -0,0 +1,103 @@
+#
+# Author:: Benjamin Black (<bb@opscode.com>)
```
mcquin: @jaymzh is this your alter ego? ;)

lib/ohai/plugins/darwin/filesystem2.rb
```diff
@@ -0,0 +1,103 @@
...
+        view[entry[:mount]][key] = val
+      end
+    end
+    view
+  end
+
+  collect_data(:darwin) do
+    fs = Mash.new
+
+
```
mcquin: [nit] Got some extra whitespace here.

lib/ohai/plugins/darwin/filesystem2.rb
```diff
@@ -0,0 +1,103 @@
...
+Ohai.plugin(:Filesystem2) do
+  provides "filesystem2"
+
+  def generate_device_view(fs)
+    view = {}
+    fs.each_value do |entry|
+      view[entry[:device]] = Mash.new unless view[entry[:device]]
+      entry.each do |key, val|
+        next if ['device', 'mount'].include?(key)
```
mcquin: I'm curious: Why do you skip if `key == 'device'` in this method, when you skip if `key == 'mount'` in the `generate_mountpoint_view` method?

spec/unit/plugins/darwin/filesystem2_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require File.expand_path(File.dirname(__FILE__) + '/../../../spec_helper.rb')
+
+describe Ohai::System, "darwin filesystem2 plugin" do
+  before(:each) do
+    @plugin = get_plugin("darwin/filesystem2")
```
mcquin: Can you update to use `:let` syntax? (e.g., `let(:plugin) { get_plugin("darwin/filesystem2") }`)

lib/ohai/plugins/darwin/filesystem2.rb
jaymzh: because I have to make 'mount' an array here when making this view. :)

lib/ohai/plugins/darwin/filesystem2.rb
jaymzh: copypasta... thanks, will fix.

spec/unit/plugins/darwin/filesystem2_spec.rb
jaymzh: will do

lib/ohai/plugins/darwin/filesystem2.rb
mcquin: Ah, gotcha.

jaymzh:
@mcquin your syntax is flawed:

```
     Failure/Error: let(:plugin) { get_plugin("darwin/filesystem2") }
       `let` is not available from within an example (e.g. an `it` block) or from constructs that run in the scope of an example (e.g. `before`, `let`, etc). It is only available on an example group (e.g. a `describe` or `context` block).
```

jaymzh:
So you can do that in in the not-before block, but then it's not clear you can do mocks in the same way...

jaymzh:
OK got 1 :+1: , do a I hear another?

spec/unit/plugins/darwin/filesystem2_spec.rb
tyler-ball: It should look like
```ruby
describe Ohai::System, "darwin filesystem2 plugin" do
  let(:plugin) { get_plugin("darwin/filesystem2") }
  before do
    allow(plugin).to receive(:collect_os).and_return(:darwin)
    allow(plugin).to receive(:shell_out).with("df -i").and_return(mock_shell_out(0, "", ""))
    allow(plugin).to receive(:shell_out).with("mount").and_return(mock_shell_out(0, "", ""))
  end
  ...
```

You are right, mocks are typically not done in a `let` block.  You can do a `let` for `stdout` as well and then it doesn't need to be in a `before` block either.

jaymzh: That makes sense.
However, almost nothing else in the tree does this (almost everything has plugin be a classvar) - and this is a mostly-copy of the linux version which works the way this one does (on a classvar). Rather than change all this away from that, I'd rather get it in as-is... if there's not a strong objection.

tyler-ball: ¯\_(ツ)_/¯

mcquin: Here are some good reasons for using let:
http://stackoverflow.com/questions/5359558/when-to-use-rspec-let/5359979#5359979

I'm trying to encourage Ohai specs to follow this pattern (most core chef
specs already do). It's okay if you'd rather not update existing specs,
though this will have to be done at some point. Writing new specs with let
syntax will make updating easier.

jaymzh: Tell you what, if we get this and #566 merged, I'll go back and changed both the linux and mac specs for both fs and fs2 to use let. :)

mcquin: Deal! :)

jaymzh:
Alright. @mcquin is on board. But she's not a maintainer anymore (even though she probably knows Ohai better than anyone else... I'm trying to follow the rules). I need one more :+1:

mcquin:
I am too a maintainer! ;)

jaymzh:
So you are. I thought you stepped down. I remember the PR. Um. Welcome back? LOL. Ok, that's two. Merging. One more PR to go.

mcquin:
I did. I'm back on a more client facing team now so I'm able to be a more
active  maintainer again.

jaymzh:
WOOO! A huge benefit for the client community, for sure!

PR #565, opened 2015-06-24
Fix darwin filesystem plugin on newer OSX.

jaymzh:
Recent versions of Darwin change default 'df' output to have '-i' by default,
which broke parsing by the filesystem plugin. This passes -i in all the time so
we always get consistent output.

There were no tests for this... and even though I'm trying deprecate this plugin
with filesystem2, I still wrote some for ya. :)


jaymzh:

tyler-ball:
:+1:

danielsdeleo:
:+1:

jaymzh:
rebased to fix docs... will merge when tests pass.

PR #566, opened 2015-06-24
Forward port fix for #469

jaymzh:
This is a forwardport of 2cb36823b173aa45736d32194db2ae8365fdad53 (which fixes
some lsblk parsing in filesystem) to filesystem2.

This also fixes a bug in the related unittests for filesystem2

PR #567, opened 2015-06-25
Fix my_mountpoint view in filesystem2; add tests

jaymzh:
The same trick for by_device to list multiple pointpoints should be in
by_mounpoint for multiple devices.

Add tests for both alternative views.


jaymzh:
wanna get this in before we release so no one starts relying on the API before it's fixed. :)

CC @chef/client-core  @chef/client-maintainers

thommay:
:+1:

jaymzh:
Thanks @thommay !! 
I have one other fs-related fix laying around: https://github.com/chef/ohai/pull/566 if you have a moment.

PR #568, opened 2015-06-25
Move filesystem-related specs to use `let`

jaymzh:
Only because @mcquin is so awesome.

Currently based ontop of #567


thommay:
I'm :+1: and the tests are happy on OS X.

PR #570, opened 2015-06-29
Add packages plugin

cmluciano:
This packages plugin allows individuals to get a list of packages that
they have installed on the system.
It works on Debian,Fedora, and Windows systems.


cmluciano:
@btm @smurawski Could you help me out with the wmi-lite testing? I'm drawing a blank on what to stub since there isn't an easy shellout return type of thing with the wmi-lite gem.

mcquin:
@cmluciano You likely need to stub the `WmiLite::Wmi` class itself. One way to do this is to use `allow_any_instance_of` and `expect_any_instance_of`. For example, to stub [this line](https://github.com/cmluciano/ohai/blob/cml/addpackages/lib/ohai/plugins/packages.rb#L54) you might write

```ruby
allow_any_instance_of(WmiLite::Wmi).to receive(:instances_of).
  with('Win32_Product').
  and_return(win32_product)
```

thommay:
Closing due to old age and being superceded.

PR #571, opened 2015-06-29
[filesystem2] Minor linux bug

jaymzh:
We were dropping 'device' for entries only in /proc/mounts


thommay:
:+1:

jaymzh:
Thanks!

OK, I technically need uno mas before I can merge.

CC: @chef/client-core @chef/client-maintainers

jaym:
:+1:

PR #572, opened 2015-07-01
we need to use our strings here

lamont-granquist:


lamont-granquist:
@dan @mattray @glennmatthews we really need to do something better here and not blindly return the os-release values.  slaving platform_family to the 'ID_LIKE' field is likely to end very badly and not do what you want at all, particularly if it does like the spec suggests and returns ID_LIKE="rhel fedora" which is nonsense as far as the Chef ecosystem is concerned -- it probably needs to be "fedora"

no idea what the platform is supposed to be, and don't have a system to play with.

thommay:
I think this is a good band aid but we need to go through and make sure we are using `/etc/os-release` properly for all the platforms that are gonna expose it.

thommay:
( :+1: )

glennmatthews:
I guess I'm not clear on the purpose of this pull. This code is cisco-specific (hence the `os_release_file_is_cisco?` check) so why does it need to be changed? What's the purpose of marking our platform as Fedora family?

glennmatthews:
@sanjaychoudhury Adding Sanjay to the conversation.

lamont-granquist:
We need to not blindly trust the fields from os-release.  I don't trust os-release and don't trust the vendors that put information into os-release -- including cisco -- we've had too many accidents over the years.  The platform and platform_family are Chef-defined values that make sense in a Chef ecosystem.

The use of 'wrlinux' as a platform_family is probably bad.  If you are fedora-derived then you should use 'fedora' there.  You'll immediately pick up some cookbooks that way which will just work.  I should also probably code more defensively around the Cisco-detection logic so that only ID=nexus is mapped to nexus.

thommay:
@lamont-granquist is this still necessary now 585 is merged?

thommay:
closing since #585 / #597 got merged.

PR #573, opened 2015-07-03
filesystem2 bugfix

jaymzh:
This makes 'mounts' in by_device and 'devices' in by_mount always get
initialized as an empty array instead of not existing. if the data doesn't
exist. This prevents callers from having to do

```
if foo and foo['mounts'] and foo['mounts'].emtpy?
```

to determine if something is mounted.

### DIFF
--- a/lib/ohai/plugins/linux/filesystem2.rb
+++ b/lib/ohai/plugins/linux/filesystem2.rb
@@ -64,8 +64,8 @@ Ohai.plugin(:Filesystem2) do
         next if ['device', 'mount'].include?(key)
         view[entry[:device]][key] = val
       end
+      view[entry[:device]][:mounts] = [] unless view[entry[:device]][:mounts]
       if entry[:mount]
-        view[entry[:device]][:mounts] = [] unless view[entry[:device]][:mounts]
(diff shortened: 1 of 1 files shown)


jaymzh:

lib/ohai/plugins/linux/filesystem2.rb
```diff
@@ -64,8 +64,8 @@ def generate_device_view(fs)
         next if ['device', 'mount'].include?(key)
         view[entry[:device]][key] = val
       end
+      view[entry[:device]][:mounts] = [] unless view[entry[:device]][:mounts]
```
thommay: `||= []` ?

lib/ohai/plugins/linux/filesystem2.rb
```diff
@@ -81,8 +81,8 @@ def generate_mountpoint_view(fs)
         next if ['mount', 'device'].include?(key)
         view[entry[:mount]][key] = val
       end
+      view[entry[:mount]][:devices] = [] unless view[entry[:mount]][:devices]
```
thommay: here too

jaymzh:
picky picky. =P

thommay:
do it :+1:

PR #574, opened 2015-07-09
Deprecate Ohai::Config in favor of Ohai::Config.ohai

mcquin:
Implements the following from [rfc053](https://github.com/chef/chef-rfc/blob/master/rfc053-ohai-config.md):

- Create config context `Ohai::Config.ohai`
- Create shortcut `Ohai.config` for `Ohai::Config.ohai`
- Deprecates top-level attributes of `Ohai::Config`

@chef/client-maintainers @chef/client-core

### DIFF
--- a/lib/ohai/config.rb
+++ b/lib/ohai/config.rb
@@ -17,6 +17,7 @@
 #
 
 require 'mixlib/config'
+require 'ohai/log'
 
 module Ohai
   class Config
@@ -32,7 +33,6 @@ module Ohai
       end
       path
     end
-
 
     default :log_level, :info
     default :log_location, STDERR
--- a/lib/ohai/config.rb
+++ b/lib/ohai/config.rb
@@ -1,6 +1,7 @@
 #
 # Author:: Adam Jacob (<adam@opscode.com>)
-# Copyright:: Copyright (c) 2008 Opscode, Inc.
+# Author:: Claire McQuin (<claire@chef.io>)
+# Copyright:: Copyright (c) 2008-2015 Chef Software, Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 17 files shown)


mcquin:
Working on resolving the EC2 failure on Windows. Fix shouldn't impact the code at large.

mcquin:
This will require changes to chef client. I'm working on those currently.

lib/ohai/config.rb
```diff
@@ -16,28 +18,91 @@
...
-require 'mixlib/config'
+require 'ohai/log'
+require 'chef-config/logger'
+
+ChefConfig::Logger = Ohai::Log
+
+require 'chef-config/config'
 
 module Ohai
+  Config = ChefConfig::Config
```
jaym: This is a little confusing. Is this about to monkeypatch ChefConfig::Config?
A comment might be helpful here.

lib/ohai/config.rb
```diff
@@ -16,28 +18,91 @@
...
+Ohai::Config[:#{option}] is set. Ohai::Config[:#{option}] is deprecated and will
+be removed in future releases of ohai. Use ohai.#{option} in your configuration
+file to configure :#{option} for ohai.
+EOM
+      Ohai::Log.warn(msg)
+    end
+  end
+
+  # Shortcut for Ohai::Config.ohai
+  def self.config
```
jaym: I might be missing something. How is this a shortcut? `Ohai::Config.config` vs `Ohai::Config.ohai`?

Also available through ChefConfig::Config.config?

lib/ohai/config.rb
mcquin: Adds these settings to ChefConfig::Config. See: https://github.com/chef/chef/blob/master/lib/chef/config.rb

I will add a comment.

ohai.gemspec
```diff
@@ -25,6 +25,7 @@ Gem::Specification.new do |s|
   s.add_dependency "ipaddress"
   s.add_dependency "wmi-lite", "~> 1.0"
   s.add_dependency "ffi", "~> 1.9"
+  s.add_dependency "chef-config", "~> 12.4.1"
```
jaym: It's going to be annoying having to bump the version here all the time. Any objections to letting be '~> 12.4`?

jaym:
:+1:

ohai.gemspec
lamont-granquist: I could see an argument actually for chef-config "~> 12.4" since I may want to upgrade ohai or chef entirely independently of each other provided that they stay on the same major version number.  We should ideally be able to still replicate omnibus builds since we should pin the version of the chef and ohai gems and the chef-config version will then just get depsolved.  I don't know about the implications for gem-installed chef and ohai though (although clearly version pinning there would also solve this problem).

So I think I'm leaning towards a much less restrictive pin here.

lamont-granquist:
I think @btm is on vacay?  It'd be nice to hold off on merging until he gets back, unless that's going to be awhile...

jaym:
@btm is gone for a few months

lib/ohai/config.rb
```diff
@@ -16,28 +18,93 @@
...
+    ].each do |option|
+      # https://docs.chef.io/config_rb_client.html#ohai-settings
+      # hints_path and plugin_path are intentionally excluded here; warnings for
+      # setting these attributes are generated in merge_deprecated_config since
+      # append (<<) operations bypass the config writer.
+      config_attr_writer option do |value|
+        warn_deprecated(option)
+        value
+      end
+    end
```
mcquin: When the config file is loaded, there is no differentiation between configuration settings `:option` and `Ohai::Config[:option]`. The settings `:log_level` and `:log_location` are in every configuration file auto-generated by chef. False deprecation warnings could be annoying and cause confusion for some users. Parsing configuration files for lines containing `Ohai::Config[:option]` and warning on their presence is the only workaround I have come up with so far. Any thoughts?

mcquin: There could also be a modified deprecation warning for `:log_level` and `:log_location`. Perhaps something along the lines of:

> Ohai::Config[:log_location] is deprecated and will be removed in future releases of ohai. Use ohai.log_location in your configuration file to configure :log_location for ohai. If your configuration file is used with other applications which configure :log_location, and you have not configured Ohai::Config[:log_location], you may disregard this warning.

mcquin: We will need to ensure that ohai config can process `:log_level` and `:log_location` as it does now (e.g., sets `Ohai.config[:log_level]` if `:log_level` is configured)

lib/ohai/config.rb
```diff
@@ -1,6 +1,8 @@
 #
 # Author:: Adam Jacob (<adam@opscode.com>)
+# Author:: Claire McQuin (<claire@chef.io>)
 # Copyright:: Copyright (c) 2008 Opscode, Inc.
+# Copyright:: Copyright (c) 2015 Chef Software, Inc.
```
jkeiser: I'd check around and make sure this is the right way to do this (two lines), but it seems right offhand ...

jkeiser:
Hanger-on :+1: :)  Nothing jumps out at me as problematic.

lib/ohai/config.rb
mcquin: [comment added](https://github.com/chef/ohai/blob/df94046b5e9d60d8699041f396ce6246501cf32e/lib/ohai/config.rb#L30-L31)

lib/ohai/log.rb
```diff
@@ -16,15 +16,14 @@
...
 
 module Ohai
   class Log
     extend Mixlib::Log
-    
-    init(Ohai::Config[:log_location])
-    level = Ohai::Config[:log_level]
-    
+
+    init(STDERR)
```
thommay: what happens if chef is configured to log using windows event logger or syslog?

thommay:
LGTM :+1:

PR #583, opened 2015-07-18
remove "warning: already initialized constant ETHERNET_ENCAPS"

juliandunn:
Remove definition of `ETHERNET_ENCAPS` as a constant and move it into the plugin body as a frozen array to avoid annoying warnings during test runs.

### DIFF
--- a/lib/ohai/plugins/solaris2/network.rb
+++ b/lib/ohai/plugins/solaris2/network.rb
@@ -53,18 +53,19 @@
 # srcof qfe1
 # inet6 fe80::203:baff:fe17:4444/128
 
+# Extracted from http://illumos.org/hcl/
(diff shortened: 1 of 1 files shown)


mcquin:
@juliandunn I don't see these warnings when running `bundle exec rspec` off of master.

juliandunn:
@mcquin Try running the specs in Chef proper.

mcquin:
Gotcha.

Technically this is a breaking change, though I doubt (m)any rely on that constant.

jaym:
could do `unless defined?`

juliandunn:
@jaym Could do that -- are you that worried about breaking backcompat though? Happy to change it if it's your preference though.

mcquin:
Let's not be lazy about backcompat, even for small things.

mcquin:
:+1:

@chef/client-core

PR #585, opened 2015-07-20
Correct platform, platform_family and version detection on Cisco's Nexus platforms.

mattray:
Cisco's upcoming Nexus platform has a native Wind River Linux userland and a
CentOS-based guestshell that runs in a container over the host operating system.

This patch supports the following platform/platform_family combinations:
CentOS 7.1: centos/rhel (unchanged)
guestshell: nexus/rhel
WRL: nexus/cisco-wrlinux

### DIFF
--- a/lib/ohai/plugins/linux/platform.rb
+++ b/lib/ohai/plugins/linux/platform.rb
@@ -92,7 +92,7 @@ Ohai.plugin(:Platform) do
     elsif File.exist?("/etc/redhat-release")
       contents = File.read("/etc/redhat-release").chomp
       if os_release_file_is_cisco? # Cisco guestshell
-        platform 'nexus_guestshell'
+        platform 'nexus_centos'
         platform_version parse_os_release_for_cisco['VERSION']
       else
         platform get_redhatish_platform(contents)
@@ -129,10 +129,12 @@ Ohai.plugin(:Platform) do
       # kernel release will be used - ex. 3.13
       platform_version `uname -r`.strip
     elsif os_release_file_is_cisco?
-      # Cisco platform not based on known distro above
-      platform 'nexus'
-      platform_family 'wrlinux'
-      platform_version parse_os_release_for_cisco['VERSION']
+      # Cisco platform not based on known distro above, normalize platform and
+      # platform_family between product lines
+      cisco_platform = parse_os_release_for_cisco
+      platform 'nexus' if cisco_platform['ID'].include?('nexus')
+      platform_family 'wrlinux' if cisco_platform['ID_LIKE'].include?('wrlinux')
+      platform_version cisco_platform['VERSION']
     elsif lsb[:id] =~ /RedHat/i
       platform "redhat"
       platform_version lsb[:release]
@@ -155,7 +157,7 @@ Ohai.plugin(:Platform) do
       platform_family "debian"
     when /fedora/, /pidora/
       platform_family "fedora"
-    when /oracle/, /centos/, /redhat/, /scientific/, /enterpriseenterprise/, /amazon/, /xenserver/, /cloudlinux/, /ibm_powerkvm/, /parallels/, /nexus_guestshell/ # Note that 'enterpriseenterprise' is oracle's LSB "distributor ID"
+    when /oracle/, /centos/, /redhat/, /scientific/, /enterpriseenterprise/, /amazon/, /xenserver/, /cloudlinux/, /ibm_powerkvm/, /parallels/, /nexus_centos/ # Note that 'enterpriseenterprise' is oracle's LSB "distributor ID"
       platform_family "rhel"
     when /suse/
       platform_family "suse"
--- a/lib/ohai/plugins/linux/platform.rb
+++ b/lib/ohai/plugins/linux/platform.rb
@@ -28,16 +28,33 @@ Ohai.plugin(:Platform) do
     contents[/Rawhide/i] ? contents[/((\d+) \(Rawhide\))/i, 1].downcase : contents[/release ([\d\.]+)/, 1]
   end
 
+  # Cisco's Nexus 3/9ks are based on Wind River Linux 5. They also have a
+  # guestshell based on CentOS 7. The file /etc/os-release is read to determine
+  # the platform_family and platform, it may point to another
+  # /etc/shared/os-release file if we are within the guestshell.
   def os_release_file_is_cisco?
     return false unless File.exist?('/etc/os-release')
+    # parse /etc/os-release into a hash
     os_release_info = File.read('/etc/os-release').split.inject({}) do |map, key_value_line|
       key, _separator, value = key_value_line.partition('=')
-      map[key] = value
+      map[key] = value.gsub(/\A"|"\Z/, '') # strip leading/trailing quotes
       map
     end
-    if os_release_info['CISCO_RELEASE_INFO'] && File.exist?(os_release_info['CISCO_RELEASE_INFO'])
-      os_release_info
+    # if it has a CISCO_RELEASE_INFO value, must be Cisco
+    cisco_release_info = os_release_info['CISCO_RELEASE_INFO']
+    if cisco_release_info && File.exist?(cisco_release_info)
+      if cisco_release_info == '/etc/os-release'
+        return os_release_info
+      end
+      # read /etc/shared/os-release to override /etc/os-release info.
+      cisco_release_info = File.read(cisco_release_info).split.inject({}) do |map, key_value_line|
+        key, _separator, value = key_value_line.partition('=')
+        map[key] = value.gsub(/\A"|"\Z/, '')
+        map
+      end
+      os_release_info.merge!(cisco_release_info)
     else
+      # has /etc/os-release, but not a Cisco box
       false
     end
   end
@@ -74,12 +91,11 @@ Ohai.plugin(:Platform) do
       platform get_redhatish_platform(contents)
       platform_version contents.match(/(\d\.\d\.\d)/)[0]
     elsif File.exist?("/etc/redhat-release")
-      if File.exist?('/etc/os-release') && (os_release_info = os_release_file_is_cisco? ) # check if Cisco
-        platform os_release_info['ID']
-        platform_family os_release_info['ID_LIKE']
-        platform_version os_release_info['VERSION'] || ""
+      contents = File.read("/etc/redhat-release").chomp
+      if (os_release_info = os_release_file_is_cisco? ) # Cisco guestshell
+        platform 'nexus_guestshell'
+        platform_version os_release_info['VERSION']
       else
-        contents = File.read("/etc/redhat-release").chomp
         platform get_redhatish_platform(contents)
         platform_version get_redhatish_version(contents)
       end
@@ -113,6 +129,11 @@ Ohai.plugin(:Platform) do
       # no way to determine platform_version in a rolling release distribution
       # kernel release will be used - ex. 3.13
       platform_version `uname -r`.strip
+    elsif (os_release_info = os_release_file_is_cisco?)
+      # Cisco platform not based on known distro above
+      platform 'nexus'
+      platform_family 'wrlinux'
+      platform_version os_release_info['VERSION']
     elsif lsb[:id] =~ /RedHat/i
       platform "redhat"
       platform_version lsb[:release]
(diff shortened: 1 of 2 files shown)


mcquin:
LGTM

@chef/client-core

mattray:
If someone from @chef/client-core wants to merge, we (Cisco and I) would greatly appreciate it.

lib/ohai/plugins/linux/platform.rb
```diff
@@ -32,12 +32,24 @@ def os_release_file_is_cisco?
...
     os_release_info = File.read('/etc/os-release').split.inject({}) do |map, key_value_line|
       key, _separator, value = key_value_line.partition('=')
-      map[key] = value
+      # strip leading/trailing quotes
+      map[key] = value.gsub(/\A"|"\Z/, '')
       map
     end
-    if os_release_info['CISCO_RELEASE_INFO'] && File.exist?(os_release_info['CISCO_RELEASE_INFO'])
-      os_release_info
+    cisco_release_info = os_release_info['CISCO_RELEASE_INFO']
```
btm: If I'm reading this right, this would be a lot easier to understand if cisco_release_info was something like cisco_release_file.

lib/ohai/plugins/linux/platform.rb
```diff
@@ -74,12 +86,16 @@ def os_release_file_is_cisco?
       platform get_redhatish_platform(contents)
       platform_version contents.match(/(\d\.\d\.\d)/)[0]
     elsif File.exist?("/etc/redhat-release")
-      if File.exist?('/etc/os-release') && (os_release_info = os_release_file_is_cisco? ) # check if Cisco
-        platform os_release_info['ID']
-        platform_family os_release_info['ID_LIKE']
-        platform_version os_release_info['VERSION'] || ""
+      contents = File.read("/etc/redhat-release").chomp
+      if (os_release_info = os_release_file_is_cisco? ) # check if Cisco
```
btm: I think we could simplify this set of nested conditionals.

First, os_release_file_is_cisco? Is doing a confusing double duty as both checking for a file and also returning that data. I'd rather see something like a #get_cisco_release as the pattern of a method ending with a ? Indicates it would give use true/false. This set of conditionals here might make more sense if that was two methods. One to tell us if we are on Cisco and another to fetch the data.

lib/ohai/plugins/linux/platform.rb
```diff
@@ -74,12 +86,16 @@ def os_release_file_is_cisco?
...
+        if os_release_info['ID'] == 'nexus'
+          platform 'nexus'
+        else
+          platform get_redhatish_platform(contents)
```
btm: So this is where I was having a hard time. While I get now how we would end up here, we shouldn't have both this else and the next one. 

If we are on Cisco, and it's nexus, set blah, else were just on red hat, set foo. If we need this many forks, we are going to have a hard time ensuring our test coverage has them all.

lib/ohai/plugins/linux/platform.rb
```diff
@@ -113,6 +129,15 @@ def os_release_file_is_cisco?
       # no way to determine platform_version in a rolling release distribution
       # kernel release will be used - ex. 3.13
       platform_version `uname -r`.strip
+    elsif (os_release_info = os_release_file_is_cisco?)
+      # Cisco platform not based on known distro above
+      if os_release_info['ID'] == 'nexus'
+        platform 'nexus'
+        platform_family 'cisco-wrlinux'
+      elsif os_release_info['ID_LIKE'] =~ /cisco-wrlinux/
+        platform_family 'cisco-wrlinux'
```
btm: Is this a breaking change? It looks like we used to get wrlinux before, now we're explicitly setting Cisco-wrlinux.

Is this platform family so unique that it needs to be its own family? Can it stay wrlinux or whatever wrlinux is based on?

lib/ohai/plugins/linux/platform.rb
```diff
@@ -113,6 +129,15 @@ def os_release_file_is_cisco?
...
+      end
+      platform_version os_release_info['VERSION'] || ""
```
btm: Why the || ""?

lib/ohai/plugins/linux/platform.rb
```diff
@@ -113,6 +129,15 @@ def os_release_file_is_cisco?
       # no way to determine platform_version in a rolling release distribution
       # kernel release will be used - ex. 3.13
       platform_version `uname -r`.strip
+    elsif (os_release_info = os_release_file_is_cisco?)
+      # Cisco platform not based on known distro above
+      if os_release_info['ID'] == 'nexus'
+        platform 'nexus'
+        platform_family 'cisco-wrlinux'
+      elsif os_release_info['ID_LIKE'] =~ /cisco-wrlinux/
```
btm: I don't see a test for this case. Is this legit? Can we get a test and a comment with what platform this might catch?

lib/ohai/plugins/linux/platform.rb
glennmatthews: Agreed that it's not obvious, however...

```
[root@n9k-108 guestshell]# cat /etc/os-release
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
PRETTY_NAME="CentOS Linux 7 (Core)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:7"
HOME_URL="https://www.centos.org/"
BUG_REPORT_URL="https://bugs.centos.org/"

CENTOS_MANTISBT_PROJECT="CentOS-7"
CENTOS_MANTISBT_PROJECT_VERSION="7"
REDHAT_SUPPORT_PRODUCT="centos"
REDHAT_SUPPORT_PRODUCT_VERSION="7"

CISCO_RELEASE_INFO=/etc/shared/os-release
```

`'CISCO_RELEASE_INFO'` is the key whose value points to a file where the Cisco release info can be found. It's probably too late to change this string in the image.

`cisco_release_info` is a hash of keys to values as loaded from this file. I don't think it makes sense to change this name, personally.

lib/ohai/plugins/linux/platform.rb
glennmatthews: We were asked to change from 'wrlinux' to 'cisco-wrlinux' just in case anyone else is using their own wrlinux variant that doesn't have the same assumptions as Cisco's wrlinux variants. I don't know what the risk is but we were asked to make this change.

lib/ohai/plugins/linux/platform.rb
glennmatthews: In case this key isn't in the file, seemed better to set platform_version to an empty string instead of nil. If that's an incorrect design, it can be removed.

lib/ohai/plugins/linux/platform.rb
glennmatthews: This is for cisco-wrlinux platforms that aren't nexus. We have at least one coming down the pipe that I know of, but I'm not sure what its platform designator will be yet. Personally I'd prefer to just set `platform = os_release_info['ID']` so as to be future-proof, but this seemed to be discouraged - see #572.

lib/ohai/plugins/linux/platform.rb
glennmatthews: Seems reasonable. We were trying to work within the existing paradigm of the code but I agree that your proposed change may be more sensible.

lamont-granquist:
It really seems like we're creating a mess here yet again, similar to including 'amazon' in platform_family of 'rhel'.

Can we get access to some of these devices to play around at all?

lib/ohai/plugins/linux/platform.rb
lamont-granquist: That rationale seems entirely wrong.  The platform_family is designed to cover different operating systems that come from a common base where the distros DO make different assumptions.  For example both "ubuntu" and "debian" have a platform_family of "debian".

If we start going down the road of requiring platform_family values to all make identical assumptions, then it dilutes the utility of platform_family down to only being useful for 'rhel' which makes the entire concept useless.

mattray:
Based off a call with @mcquin I'll refactor for clarity and comments and update this PR.

lamont-granquist:
Can someone answer a simple question for me?

How do you set the hostname of a wind river linux host?  Is it /etc/sysconfig/network like redhat/fedora?

lamont-granquist:
And 'cisco-wrlinux' looks very wrong to me, unless someone can explain that rationale, then i'm very :-1: on all of these patches.

mattray:
@lamont-granquist the "cisco-wrlinux" is because Wind River Linux is an embedded, highly customizable Linux that is very likely to have changes making it quite divergent from the next Wind River Linux build. We're calling it "cisco-wrlinux" because this is Cisco's custom build based off of WRL 5 and we expect to eventually add support for Cisco's WRL 3 and 7-based builds.

lamont-granquist:
That is still not a compelling enough reason for me.

The whole purpose of the platform_family attribute is to allow cookbook authors to be lazy when distributions have similarities.  The question to ask is not "how is it different?" but "is it at all similar?"

And this is important to get correct, because if you go down this road then you pretty much ensure that very few community cookbooks will ever work with cisco linux and the cookbook reusability for cisco linux will likely remain bad.

If you *can* pick platform_family of 'fedora' and have some community cookbooks work, then you should do so.

mattray:
Updated the patch to accomodate @mcquin and @lamont-granquist's feedback. The platform_family for WRL is now "wrlinux", "cisco-wrlinux" was considered a premature optimization. The guestshell now reports "nexus_guestshell" as the platform to differentiate it from the WRL platform. If you need to key off the platform, you can match with startswith("nexus")

lib/ohai/plugins/linux/platform.rb
```diff
@@ -28,16 +28,33 @@ def get_redhatish_version(contents)
...
+        key, _separator, value = key_value_line.partition('=')
+        map[key] = value.gsub(/\A"|"\Z/, '')
+        map
+      end
+      os_release_info.merge!(cisco_release_info)
     else
+      # has /etc/os-release, but not a Cisco box
       false
     end
   end
```
lamont-granquist: So this method is hella ugly.

First thing you probably want is a single method which reads an os-release file and renders it into a hash.

Then you probably want to write another method which does the read-os-release; find-cisco-release-info, read cisco-release-info; merge-with-os-release-info dance and then memoizes the result.  This will then be a reusable method that other stuff that inspects os-release-info can use.

Then the `os_release_file_is_cisco?` check should be vastly simplified into something like just checking `os_release_info.key?('CISCO_RELEASE_INFO')`

I'd also suggest not worrying about special casing when cisco_release_info is also '/etc/os-release' since the result of merge!'ing /etc/os-release onto /etc/os-release is fine.

lib/ohai/plugins/linux/platform.rb
```diff
@@ -74,12 +91,11 @@ def os_release_file_is_cisco?
       platform get_redhatish_platform(contents)
       platform_version contents.match(/(\d\.\d\.\d)/)[0]
     elsif File.exist?("/etc/redhat-release")
-      if File.exist?('/etc/os-release') && (os_release_info = os_release_file_is_cisco? ) # check if Cisco
-        platform os_release_info['ID']
-        platform_family os_release_info['ID_LIKE']
-        platform_version os_release_info['VERSION'] || ""
+      contents = File.read("/etc/redhat-release").chomp
```
lamont-granquist: this line probably needs to get moved back?

mattray: It gets set below at line 158 with the new /nexus_guestshell/ match

mattray:
After all the feedback from Cisco, @mcquin and @lamont-granquist the final behavior for Wind River Linux and the CentOS 7.1 guestshell on Cisco's 3/9k platform is now this:

WRL: platform "nexus" platform_family "wrlinux"

guestshell: platform "nexus_centos" platform_family "rhel"

glennmatthews:
:+1:

mcquin:
:rocket:

mattray:
We can close this one once #597 is merged.

lamont-granquist:
since i pulled your commits, merging #597 merged this one...

lamont-granquist:
For posterity (mostly to restate in my words so I remember when this comes up in a year or two...),

We chose "wrlinux" over "cisco-wrlinux" as the platform_family for the embedded linux since it was felt that was premature optimization.  We don't even have other "wrlinux"es to look at to determine if we need to be creating unique platform_families for them all or not.  Without anything concrete, and with a bias towards fewer platform_families (which makes platform_families _as_a_fundamental_concept_ more usable), we kept it "wrlinux", which is what other wind river linuxes also SHOULD use until such a time as we have enough information that this looks like a poor decision, but we then actually understand why -- otherwise YAGNI.

We chose "nexus" as the platform for the embedded o/s because that is what cisco calls it and seems a decent enough choice.

We threw out the idea of using "nexus" as the platform for both the host and guest o/s and having different platform_families since that inverted our heirarchy and seemed very broken.  We chose "nexus-centos" as the platform name so that /^nexus/ would match both and because it is centos-based (the fact that platform =~ /centos/ works also seems like a nice bonus).  We are anticipating having "nexus-ubuntu" show up as well.  The platform_family is "rhel" since for most purposes it should act exactly like centos/rhel.

mattray:
Thanks!

glennmatthews:
:+1: Thank you Matt and Lamont.

PR #586, opened 2015-07-22
Update the chef-config dependency version to be compatible with chef

mcquin:
Fixing things that I broke...

Chef will always depend on [an unreleased version of `chef-config`](https://github.com/chef/chef/blob/master/chef.gemspec#L18). Chef also depends on `ohai`. To prevent dependency conflicts we have to ensure the version of `chef-config` that Ohai depends on is compatible with whatever the current version of Chef is on master. This will be fine and good until we start working on Chef 13.

Yay.

Apparently, the desire is to keep the `chef-config` version the same as the Chef version, so we can't relax the constraint there.

@chef/client-core 

PS: Working on omnibus changes to get this to build correctly.

### DIFF
--- a/ohai.gemspec
+++ b/ohai.gemspec
@@ -25,7 +25,15 @@ Gem::Specification.new do |s|
   s.add_dependency "ipaddress"
   s.add_dependency "wmi-lite", "~> 1.0"
   s.add_dependency "ffi", "~> 1.9"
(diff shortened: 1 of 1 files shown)


ksubrama:
:+1:

danielsdeleo:
:+1: :shipit:

PR #591, opened 2015-07-30
Fix memory plugin output on solaris

sh9189:
Memory plugin does not return correct output.
Running ohai memory on a solaris machine gives output as
{
  "total": 0
}
This is because memory plugin on solaris uses "prtconf -m" to get memory which does not seem to be valid.
This pull request attempts to fix this. (Reference http://docs.oracle.com/cd/E23824_01/html/821-1451/gkkwk.html#enflm on how to get memory on a solaris machine)

### DIFF
--- a/lib/ohai/plugins/solaris2/memory.rb
+++ b/lib/ohai/plugins/solaris2/memory.rb
@@ -19,6 +19,7 @@ Ohai.plugin(:Memory) do
 
   collect_data(:solaris2) do
     memory Mash.new
-    memory[:total] = shell_out("prtconf -m").stdout.to_i
+    meminfo =  shell_out("prtconf | grep Memory").stdout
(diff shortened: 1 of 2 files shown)


mcquin:
Makes sense.

You'll need to update [this method stub](https://github.com/chef/ohai/blob/master/spec/unit/plugins/solaris2/memory_spec.rb#L23) to `shell_out` with the new command and return as stdout `"Memory size: 8194 Megabytes\n"`.

sh9189:
@mcquin Thanks for pointing out where to update the spec.  I have made the suggested changes, please take a look.

mcquin:
Thanks for the update!

:+1:

@chef/client-core

thommay:
yep, tested on my illumos box and this works far better! :+1:

PR #593, opened 2015-08-06
Remove config logging, resolve :auto log-level.

mcquin:
Two things I discovered while plugging the new ohai config system into chef (https://github.com/chef/chef/pull/3689):

1. It's unnecessary to set `ChefConfig.logger = Ohai::Log`. Chef sets this logger to `Chef::Log` for [reasons which aren't applicable to ohai](https://github.com/chef/chef/blob/master/lib/chef/config.rb#L25-L27), IMO. Ohai will only use the configuration settings in its config context, none of which use a logger. Besides, setting `ChefConfig.logger` in ohai will override the setting by chef. Removing this setting reduces headache.

2. Chef allows an `:auto` log-level, which gets [resolved when `chef-client` runs](https://github.com/chef/chef/blob/master/lib/chef/application.rb#L141). Test Kitchen [explicitly configures log_level to `:auto`](https://github.com/test-kitchen/test-kitchen/blob/dp_chefignore/lib/kitchen/provisioner/chef_solo.rb#L50). This causes `'log_level' => :auto` to appear in the [configuration hash which is merged into the ohai configuration context](https://github.com/chef/ohai/blob/31d6b0e30659557080dcfd66ec6223aef4fecc1f/lib/ohai/config.rb#L49). Ohai does not currently support `:auto`, and so Test Kitchen converges would fail. I've added config support for `log_level :auto` and `ohai.log_level :auto`, which resolves to `:info`.

@chef/client-core

### DIFF
--- a/lib/ohai/config.rb
+++ b/lib/ohai/config.rb
@@ -17,12 +17,8 @@
 # limitations under the License.
 #
 
-require 'ohai/log'
-require 'chef-config/logger'
-
-ChefConfig::Logger = Ohai::Log
-
 require 'chef-config/config'
+require 'ohai/log'
 
 module Ohai
   Config = ChefConfig::Config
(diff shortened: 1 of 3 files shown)


ksubrama:
:+1:

thommay:
LGTM :+1:

PR #596, opened 2015-08-10
Relax restriction on chef dependency.

ksubrama:
During omnibus builds, bundler --without development still downloads all
dependencies specified so that it can resolve versions consistently
regardless of which install groups are specified.  If ohai pins a
chef dependency to chef master, then any changes to chef master that
pull in a different version of ohai will impact our ability to bundle
install our dependencies.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,8 +9,6 @@ gem "rspec_junit_formatter", :git => 'https://github.com/sj26/rspec_junit_format
                              :ref => "147836c41fab23ff7b92806f34122c8e5f2ddcad"
 
 group :development do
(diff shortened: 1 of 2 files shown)


mcquin:
I'm :+1:, as `chef` is a development dependency in the gemspec anyway.

tyler-ball:
:+1:

schisamo:
:+1: assuming you fix the CI/merge failures above

ksubrama:
I'm not going to merge at all :)  This was just a convenient way to show the diff.  The version bumps and stuff don't need to go into master.

thommay:
@ksubrama if you don't want to merge this can we close it now?

PR #598, opened 2015-08-11
Fix aix memory plugin output

sh9189:
Memory plugin does not return any memory attributes on aix due to a minor bug in definition of memory Mash. This pull request is to fix that and also adds unit tests for aix memory plugin

### DIFF
--- a/lib/ohai/plugins/aix/memory.rb
+++ b/lib/ohai/plugins/aix/memory.rb
@@ -20,7 +20,7 @@ Ohai.plugin(:Memory) do
   provides "memory"
 
   collect_data(:aix) do
-    memory = Mash.new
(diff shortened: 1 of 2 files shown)


johnbellone:
:+1: from our end here.

irvingpop:
Paging @scotthain for review

irvingpop:
The reason this was broken in ohai was because there's no memory test for AIX - this PR adds it.

scotthain:
:+1:

irvingpop:
Travis is acting funny, but I can confirm via rspec that the new memory test works:

```
Ohai::System AIX memory plugin
  should get total memory
  should get free memory
```

PR #599, opened 2015-08-13
Add swap space attributes for solaris memory plugin

sh9189:
This PR adds swap space attributes for solaris memory plugin. Swap space is computed using "swap -s" command and converted to megabytes to be consistent with memory[total] which is also in megabytes.


johnbellone:
:+1: /cc @irvingpop

thommay:
LGTM :+1: cc @lamont-granquist

lamont-granquist:
:+1:

PR #600, opened 2015-08-13
Only run ifconfig against active interfaces

acaiafa:
Currently on AIX if ifconfig is run against interfaces that are down it will bring that interface up. This fix will only run ifconfig against available interfaces.

### DIFF
--- a/lib/ohai/plugins/aix/network.rb
+++ b/lib/ohai/plugins/aix/network.rb
@@ -56,7 +56,7 @@ Ohai.plugin(:Network) do
     end
 
     # List the interfaces in system.
-    so = shell_out("lsdev -Cc if")
+    so = shell_out("lsdev -Cc if | grep Available")
     so.stdout.lines.each do |line|
       if line =~ /(\S+) (\S+)\s+(.+)/
         interface = $1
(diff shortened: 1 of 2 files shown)


irvingpop:
Paging @chef/client-core for review.   It appears that travis doesn't recognize this project as `active` and in Appveyor all of the new checks failed - need some kind of guard for windows?

danielsdeleo:
Ohai doesn't have integration tests, only unit tests. In general, tests should be written so that they'll pass everywhere, rather than turning them off and on. This change may have broken a method stub in existing tests?

thommay:
hey @acaiafa can you review the test failures in travis please? 

They all look the same: it should just be a matter of updating https://github.com/chef/ohai/blob/master/spec/unit/plugins/aix/network_spec.rb#L70 to match the command you're now issuing.

acaiafa:
@thommay ok tests are fixed and passing.

thommay:
:+1: LGTM

jaym:
I think this needs a test for when an interface is not available

acaiafa:
@jaym not sure what you mean? 

We don't want to run ifconfig against a defined and not available interface.

jaym:
nvm, https://github.com/chef/ohai/pull/600/files#diff-9e442ba64a57db9e411b53c990c8d17cR29 is indeed correct because of the grep.
:+1:

PR #601, opened 2015-08-13
Add support for sparc based processors in solaris cpu plugin

sh9189:
Solaris cpu plugin does not return correct per processor data on systems that have sparc processor as the current plugin is targeted towards systems that have x86 processor. From solaris 11.2 documentation (https://docs.oracle.com/cd/E36784_01/html/E36819/frcbd.html#ADSYSeqbug), output of  "psrinfo -p -v"  is different based on sparc vs x86 systems. This PR adds support for sparc based systems.

### DIFF
--- a/lib/ohai/plugins/solaris2/cpu.rb
+++ b/lib/ohai/plugins/solaris2/cpu.rb
@@ -26,27 +26,27 @@ Ohai.plugin(:CPU) do
 
       index = i.to_s
       cpu[index] = Mash.new
-      cpu[index]['vendor_id'] = cpu_info[1]
-      cpu[index]['family'] = cpu_info[4]
-      cpu[index]['model'] = cpu_info[6]
-      cpu[index]['stepping'] = cpu_info[8]
-      cpu[index]['model_name'] = model_name.strip
-      cpu[index]['mhz'] = cpu_info[10]
+      cpu[index]["vendor_id"] = cpu_info[1]
+      cpu[index]["family"] = cpu_info[4]
+      cpu[index]["model"] = cpu_info[6]
+      cpu[index]["stepping"] = cpu_info[8]
+      cpu[index]["model_name"] = model_name.strip
+      cpu[index]["mhz"] = cpu_info[10]
     end
   end
 
   def get_sparc_processor_info
     i = 0
     cores = 0
-    shell_out('psrinfo -v -p').stdout.lines.each do |line|
+    shell_out("psrinfo -v -p").stdout.lines.each do |line|
       case  line.strip
         when /(\d+)\s+cores/
-          cores += "#{$1}".to_i
+          cores += $1.to_i
         when /^(\S+).*\b(\d+)\s+MHz\)$/
           index = i.to_s
           cpu[index] = Mash.new
-          cpu[index]['model_name'] = "#{$1}"
-          cpu[index]['mhz'] = "#{$2}"
+          cpu[index]["model_name"] = $1
+          cpu[index]["mhz"] = $2
           i += 1
       end
     end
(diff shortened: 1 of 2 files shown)


johnbellone:
:+1: /cc @irvingpop

lib/ohai/plugins/solaris2/cpu.rb
```diff
@@ -38,4 +34,35 @@
...
     end
   end
+
+  def get_sparc_processor_info
+    i = 0
+    cores = 0
+    shell_out('psrinfo -v -p').stdout.lines.each do |line|
+      case  line.strip
+        when /(\d+)\s+cores/
+          cores += "#{$1}".to_i
```
thommay: just `$1.to_i`; no need to cast it to a string first.

lib/ohai/plugins/solaris2/cpu.rb
```diff
@@ -38,4 +34,35 @@
...
+        when /^(\S+).*\b(\d+)\s+MHz\)$/
+          index = i.to_s
+          cpu[index] = Mash.new
+          cpu[index]['model_name'] = "#{$1}"
```
thommay: here too; just `[model_name] = $1`. Also, please consistently use either single quotes or double quotes. My preference would be to use doubles.

lib/ohai/plugins/solaris2/cpu.rb
```diff
@@ -38,4 +34,35 @@
       cpu[index]['mhz'] = cpu_info[10]
     end
   end
+
+  def get_sparc_processor_info
+    i = 0
+    cores = 0
+    shell_out('psrinfo -v -p').stdout.lines.each do |line|
```
thommay: use `each_with_index`:
```
shell_out('psrinfo -v -p').stdout.lines.each_with_index do |line, i|
```

lib/ohai/plugins/solaris2/cpu.rb
```diff
@@ -38,4 +34,35 @@
...
+    cpu[:cores] = cores
+  end
+
+  collect_data(:solaris2) do
+    cpu Mash.new
+    cpu[:total] = shell_out("psrinfo | wc -l").stdout.to_i
+    cpu[:real] = shell_out("psrinfo -p").stdout.to_i
+    
+    processor_type = shell_out("uname -p").stdout.strip
+    if processor_type.eql?('sparc')
```
thommay: use `==` in general rather than `eql?` Also quotation consistency, please.

thommay:
Logic looks good, code needs some tweaks :)

sh9189:
@thommay Thanks for the quick review :) , I made the suggested changes regarding quotation consistency and avoid unnecessary string conversion. However, I can't use "each with index" while looping as you mentioned as the output of "psrinfo -v -p" does not give the exact number of physical processors on the system and the index can only be incremented every time there is a line containing 'Mhz' in the output. Please review the latest changes.

sh9189:
I missed a test case for getting total number of cores on sparc processors. Added that now.

thommay:
Great, thanks. :+1: cc @chef/client-core

spec/unit/plugins/solaris2/cpu_spec.rb
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 
-require File.expand_path(File.dirname(__FILE__) + '/../../../spec_helper.rb')
+require File.expand_path(File.dirname(__FILE__) + "/../../../spec_helper.rb")
```
jaym: Don't think this should be part of this PR

lib/ohai/plugins/solaris2/cpu.rb
```diff
@@ -16,12 +16,8 @@
...
 Ohai.plugin(:CPU) do
   provides "cpu"
-
-  collect_data(:solaris2) do
-    cpu Mash.new
-    cpu[:total] = shell_out("psrinfo | wc -l").stdout.to_i
-    cpu[:real] = shell_out("psrinfo -p").stdout.to_i
-
+  
+  def get_x86_processor_info
```
jaym: pedantic: getters usually don't change state. perhaps set is a better name?

sh9189:
@jaym I made the suggested changes. Please review.

jaym:
:+1:

PR #602, opened 2015-08-13
Add swap space attributes for AIX

sh9189:
This PR adds swap space attributes for AIX as found in memory plugins for other platforms

### DIFF
--- a/lib/ohai/plugins/aix/memory.rb
+++ b/lib/ohai/plugins/aix/memory.rb
@@ -21,8 +21,13 @@ Ohai.plugin(:Memory) do
 
   collect_data(:aix) do
     memory Mash.new
+    memory[:swap] = Mash.new
 
     meminfo = shell_out("svmon -G -O unit=MB,summary=longreal | grep '[0-9]'").stdout
     memory[:total], u, memory[:free] = meminfo.split
+    
+    swapinfo = shell_out("swap -s").stdout.split #returns swap info in 4K blocks
+    memory[:swap]['total'] = (swapinfo[2].to_i / 1024.0) * 4
+    memory[:swap]['free'] = (swapinfo[10].to_i / 1024.0) * 4
   end
 end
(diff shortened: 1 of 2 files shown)


johnbellone:
:+1:

johnbellone:
@irvingpop

liutc:
Is there any special reason to return the swap size in 4K blocks? Why not in MB or KB?

thommay:
@lamont-granquist can you take a look? The 4k blocks thing does seem weird but maybe that's an AIXism?

lamont-granquist:
looks like its a JFS2-ism.  it looks like it can be tuned differently but i doubt anyone does for swap partitions?

i think i'm okay with merging as-is and letting bug reports find any edge conditions with different filesystems and different tunings since i suspect those edge conditions won't occur enough and worrying about dynamically determining the fs blocksize of the swap partition will be YAGNI.

lamont-granquist:
so :+1:

liutc:
Why shouldn't we use the "size" in KB or MB? Other platforms use KB (at
least for Linux).

We do not need to handle that by ourself the block size thing. See below
output from one AIX box I manage. If we use "swap -l", we just get the size
and do a sum will be all good.

swap -l
device              maj,min        total       free
/dev/paging01         38,  3      1024MB      1023MB
/dev/paging00         37,  4      1280MB      1279MB
/dev/paging02         39,  3       128MB       124MB
/dev/hd6              10,  2      2560MB      2551MB

swap -s
allocated  =    1277952 blocks    used  =   3432 blocks      free  =
1274520 blocks



Best Regards,

Dr. Tiancheng Liu (?$BE7@.(B)

Cloud Operations Technology
IBM Research - China

Phone: (86-10)58748521
Fax:   (86-10)58748731
Email: liutc@cn.ibm.com

Address: Building 19 Zhongguancun Software Park, 8 Dongbeiwang WestRoad,
Haidian District, Beijing, P.R. China
Beijing 100193

lamont-granquist:
@liutc can you submit a patch for that?

PR #604, opened 2015-08-14
use bundle rake tasks

thommay:
cc @chef/client-core

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,17 +1,6 @@
-require 'rubygems'
-require 'rubygems/package_task'
-require 'rubygems/specification'
+require "bundler/gem_tasks"
 require 'date'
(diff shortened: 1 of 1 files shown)


jaym:
:+1:

danielsdeleo:
@thommay does this change the names of the tasks? Could have an impact on Ci if so.

thommay:
It changes the `:gem` task to `:build` and removes `:package`. Omnibus uses `gem build` directly, travis uses `rake spec` and Appveyor uses `:default` which is an alias to `:spec`. So I don't think it's a problem, but I'll keep an eye out for failing Jenkins.

danielsdeleo:
ok, cool.

PR #606, opened 2015-08-14
Windows memory plugin

sh9189:
This PR builds on top of #452 and adds tests. All memory attributes are converted to MB to be consistent with other platforms like Solaris,Unix. Thanks to @webframp for the initial pull request.

PR #609, opened 2015-08-18
Make aix, solaris memory output consistent to linux, add spec test for linux memory

sh9189:
1. Make linux memory plugin output to be in megabytes, consistent with other platforms such as Solaris, AIX, Windows. This helps with writing any tool that read memory plugin attributes regardless of platforms. 
2. Add spec tests for the linux memory plugin


thommay:
we probably can't take this until Chef 13 at the earliest, and it would break the entire world even then. :(

sh9189:
@thommay, In that case, do you think that it is better to change the output of other platform's memory plugin to match that of Linux. That is less of a breaking change compared to this and I can change the pull request to do that and to add spec tests for linux memory plugin. For example in #606, I am changing kB to MB for Windows memory to be consistent with other platforms. and only recently did we fix memory plugin output on solaris through #591, #599 and on aix through #598 and #602. These changes are not released yet so I doubt anyone is relying on them and memory plugins on other platforms (AIX,Solaris,Windows) were not available before these pull requests. Let me know your thoughts.

thommay:
yeah, i think KB across the board might be better. Sorry for making you do additional work.

sh9189:
@thommay  I have made changes so that memory plugin output is consistent across AIX,Solaris and Linux. Please review

sh9189:
@thommay With the latest changes, can we consider this as a patch to be merged to master instead of "Accepted Major" milestone?

thommay:
LGTM :+1: - @chef/client-maintainers

mcquin:
:+1:

sh9189:
@mcquin @thommay  Can this be marked "Ready for Merge" or does it need more review?

btm:
:+1:

I've reviewed the changes and I'm pretty sure they're not breaking any existing release, so this could go out in a minor.

I'm a little sad that we're pushing out the unit with the data, such that you've got to strip it if you want to do any arithmetic with the data. OTOH, it makes it a little "not our problem" if some values are in different units and we're letting people know what the units are.

PR #610, opened 2015-08-20
Convert Solaris OHAI CPU detection to kstat from psrinfo

cachamber:
# Issue:
* psrinfo -p does not work across all platforms.  Oracle T4-1 platforms

# Update:
* Converted the CPU detection to use kstat, tested/validated on Solaris 8,9,10,11, and across SPARC and x86 architectures 
* Generalized CPU attribute support - SPARC and X86 are now consistent
* Added cpustate counters to count CPU states - useful for software license restrictions for on-line CPUs 
* Kept all previous Key/Value pairs to align with other OSs

### DIFF
--- a/lib/ohai/plugins/solaris2/cpu.rb
+++ b/lib/ohai/plugins/solaris2/cpu.rb
@@ -50,7 +50,7 @@ Ohai.plugin(:CPU) do
            cpu[instance]["mhz"] = value
         when /brand/
            cpu[instance]["model_name"] = value.sub(/\s+/," ")
-        when /state$/
+        when /^state$/
            cpu[instance]["state"] = value
            cpu["cpustates"][value] = 0 if cpu["cpustates"][value].nil?
            cpu["cpustates"][value] += 1 
--- a/lib/ohai/plugins/solaris2/cpu.rb
+++ b/lib/ohai/plugins/solaris2/cpu.rb
@@ -32,7 +32,7 @@ Ohai.plugin(:CPU) do
     cpusockets = Array.new
     processor_info.each_with_index do |processor, i|
       desc,instance,record,keyvalue = processor.split(":")
-      cpu[instance] = Mash.new if cpu[instance].nil?
+      cpu[instance] ||= Mash.new
       if (currentcpu !=  instance)
          cpu["total"] += 1
          currentcpu = instance
@@ -52,7 +52,7 @@ Ohai.plugin(:CPU) do
            cpu[instance]["model_name"] = value.sub(/\s+/," ")
         when /^state$/
            cpu[instance]["state"] = value
-           cpu["cpustates"][value] = 0 if cpu["cpustates"][value].nil?
+           cpu["cpustates"][value] ||= 0 
            cpu["cpustates"][value] += 1 
         when /core_id/
            cpu[instance]["core_id"] = value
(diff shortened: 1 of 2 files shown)


cachamber:
I've tested the following platforms:
Sun 480R,V240/440,T3-1,T4-1,T5120 as well as various x86 based hardware 
Tested on various releases of Solaris 10 and 11

I don't have access to any newer SPARC based hardware and would appreciate some validations on those platforms.

thommay:
@sh9189 can you take a look at this since you did the work to make SPARC work originally. I've tested this on illumos and it seems happy there.

thommay:
:+1:

thommay:
@chef/client-maintainers

lib/ohai/plugins/solaris2/cpu.rb
```diff
@@ -17,52 +17,54 @@
...
+        when /core_id/
+           cpu[instance]["core_id"] = value
+           # Detect hyperthreading/multithreading
+           cpucores.push(value) if cpucores.index(value).nil?
+        when /family|fpu_type|model|stepping|vendor_id/
+           cpu[instance][key] = value
+      end
     end
+    cpu["cores"] = cpucores.size
+    cpu["corethreads"] = (cpu["total"] / cpucores.size)
```
sh9189: I don't think corethreads attribute is needed since it can be calculated from cpu["total" and cpu["cores"]. But I will let the chef engineering team make the final call.

lib/ohai/plugins/solaris2/cpu.rb
```diff
@@ -17,52 +17,54 @@
...
+    cpu["sockets"] = cpusockets.size
```
sh9189: Same as above, cpu["sockets"] is the same as cpu["real"]

sh9189:
LGTM except for my comments. Tested on our solaris sparc machines and works great.

lib/ohai/plugins/solaris2/cpu.rb
```diff
@@ -17,52 +17,54 @@
...
+    cpucores = Array.new
+    cpusockets = Array.new
+    processor_info.each_with_index do |processor, i|
+      desc,instance,record,keyvalue = processor.split(":")
+      cpu[instance] ||= Mash.new
+      if (currentcpu !=  instance)
+         cpu["total"] += 1
+         currentcpu = instance
+      end 
+      kv = keyvalue.gsub(/\s+/,"=").split(/=/)
```
mcquin: Ruby supports splitting on a regex delimiter. You can split on `/\s+/`:

```
[16] pry(main)> keyvalue
=> "brand      Crazy(r) Argon(r) CPU           Y5570  @ 1.93GHz\n"
[17] pry(main)> kv = keyvalue.gsub(/\s+/,"=").split(/=/)
=> ["brand", "Crazy(r)", "Argon(r)", "CPU", "Y5570", "@", "1.93GHz"]
[18] pry(main)> keyvalue.split(/\s+/)
=> ["brand", "Crazy(r)", "Argon(r)", "CPU", "Y5570", "@", "1.93GHz"]
```

cachamber: Thanks, much cleaner that way.

lib/ohai/plugins/solaris2/cpu.rb
cachamber: Agreed.  I would prefer to use the cores/sockets instead of total/real, as it's more descriptive.  But I kept the real/total to make sure that anything written to use those values wouldn't break.  I also understand that corethreads is a computed value, but I like to have the value set for easy comparison, such as determining if hyperthreading is turned on for an x86 based machine.

mcquin:
:+1:

PR #612, opened 2015-08-20
Add total cores to linux cpu plugin

sh9189:
Add a cores attributes to get total number of cores on linux machines.

### DIFF
--- a/lib/ohai/plugins/linux/cpu.rb
+++ b/lib/ohai/plugins/linux/cpu.rb
@@ -80,5 +80,6 @@ Ohai.plugin(:CPU) do
     cpu cpuinfo
     cpu[:total] = cpu_number
     cpu[:real] = real_cpu.keys.length
+    cpu[:cores] = real_cpu.keys.length * cpu["0"]["cores"].to_i
   end
 end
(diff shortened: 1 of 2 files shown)


lib/ohai/plugins/linux/cpu.rb
```diff
@@ -80,5 +80,6 @@
     cpu cpuinfo
     cpu[:total] = cpu_number
     cpu[:real] = real_cpu.keys.length
+    cpu[:cores] = real_cpu.keys.length * cpu["0"]["cores"].to_i
```
mcquin: Are all cpus guaranteed to have the same number of cores?

sh9189: I believe so. At least that's what I see on all of our Linux machines.

thommay:
i think this is fine, :+1:

lib/ohai/plugins/linux/cpu.rb
lamont-granquist: Yeah I think its fairly unlikely for a Linux box to have a mix of say dual core and quad core in different sockets.  That may, in fact, make lots of MBs sad if the CPUs aren't reasonably identicalish.  Certainly, we can fix it if it ever becomes a problem, YAGNI.

lamont-granquist:
:+1:

PR #617, opened 2015-09-05
initial Cisco IOS-XR support

mattray:
This is for adding support for Cisco's IOS-XR platform on Wind River Linux 7 (Cisco's Nexus platform is on WRL5). This reads the /etc/os-release, checks for CISCO_RELEASE_INFO and if the ID_LIKE=wrlinux and the ID=eXR then the platform is 'ios-xr'. 

ios-xr's guestshell will be different from nexus_centos and added once we have access to that platform.

### DIFF
--- a/lib/ohai/plugins/linux/platform.rb
+++ b/lib/ohai/plugins/linux/platform.rb
@@ -143,9 +143,13 @@ Ohai.plugin(:Platform) do
       # kernel release will be used - ex. 3.13
       platform_version `uname -r`.strip
     elsif os_release_file_is_cisco?
-      raise "unknown Cisco /etc/os-release ID field" unless os_release_info['ID'].include?('nexus')
+      raise "unknown Cisco /etc/os-release ID field" unless os_release_info['ID'] =~ /nexus|exr/i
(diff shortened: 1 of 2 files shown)


glennmatthews:
1. Ohai convention for platform names seems to be underscores rather than dashes (`ibm_powerkvm`, `nexus_centos`) so probably preferable to use `ios_xr` rather than `ios-xr`.

2. I will be cleaning up the os-release file in the very near future (hopefully this week) as follows:
    * `ID=eXR` => `ID=ios_xr`
    * `ID_LIKE=wrlinux` => `ID_LIKE=cisco-wrlinux`
    * `NAME=IOS-XR` => `NAME="IOS XR"`
    * `BUILD_ID=TBD` => (removed)

    Can you update the logic and the spec accordingly?

adamleff:
I agree that the platform should be `ios_xr` instead of `ios-xr` (underscore instead of hyphen) if anything for the reason that the chef-sugar helper method will need to have underscores instead of hyphens and we should remain consistent.

mattray:
Cisco has let us know that more changes are coming, invalidating this PR.

PR #624, opened 2015-09-24
WIP: Adds the 'root_user' Ohai attribute.

johnbellone:
This abstracts out the root user information just as the root group
information is. It makes the appropriate name available via the
root_user attribute.

### DIFF
--- a/lib/ohai/plugins/root_user.rb
+++ b/lib/ohai/plugins/root_user.rb
@@ -1,0 +1,28 @@
+#
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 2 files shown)


lib/ohai/plugins/root_user.rb
```diff
@@ -0,0 +1,28 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+Ohai.plugin(:RootUser) do
+  provides 'root_user'
+
+  collect_data do
+    case ::RbConfig::CONFIG['host_os']
```
mcquin: It would be more resilient to have this plugin declare a dependency on the `'os'` attribute and then `case os`. On Windows, `os` returns `'windows'` (see: https://github.com/chef/ohai/blob/master/lib/ohai/mixin/os.rb#L50).

thommay:
superseded by #698

PR #627, opened 2015-09-29
Removed chef as development_dependency in gemspec as it is already li…

tylercloke:
…sted in Gemfile development block.


mcquin:
I believe this is going to cause problems once the version in chef/chef master is 13, because of [this dependency](https://github.com/chef/ohai/blob/master/ohai.gemspec#L36).

lamont-granquist:
I think the only difference between the two is that dev gems listed in the gemspec show up on rubygems, while dev gems in Gemfile's give you the ability to pull from github and such.  We probably do want to have it only in one place, and most likely want it in the Gemfile since we need the flexibility to pull from GH.

smurawski:
:+1:

thommay:
:+1:

lamont-granquist:
:+1:

Issue #623, opened 2015-09-22
wrong CPU count 


Roviluca:
ohai,

i'm working in a virtual machine with:
  OS: windows server 2012R2
  Powershell: 5 (production preview)
  Chef-client: 12.4.2
  "real" CPU 4
  "total" CPU 4
 
and cpu infromations are not correct in ohai:
```json
{
  "cpu": {
    "0": {
      "vendor_id": "GenuineIntel",
      "family": "2",
      "model": "*****",
      "stepping": null,
      "physical_id": "CPU0",
      "cores": 1,
      "model_name": "Intel64 Family 6 Model 45 Stepping 7",
      "mhz": "****",
      "cache_size": "0 KB"
    },
    "total": 1,
    "real": 1
  },
  ...
 }
```
if you need some further tests to help you identify the issue let me know.
Thank you

thommay:
please can you run `chef-shell` on that node and then do the following:
```
require 'wmi-lite/wmi'
wmi = WmiLite::Wmi.new
wmi.instances_of 'Win32_Processor'
```

I just tried this on a fresh 2012r2 instance and got back the following:
```
 => [#<WmiLite::Wmi::Instance:0x26d5568 @wmi_ole_object=#<WIN32OLE:0x26d5868>, @property_map={"addresswidth"=>64, "architecture"=>9, "availability"=>3, "caption"=>"Intel64 Family 6 Model 62 Stepping 4", "configmanagererrorcode"=>nil, "configmanageruserconfig"=>nil, "cpustatus"=>1, "creationclassname"=>"Win32_Processor", "currentclockspeed"=>2500, "currentvoltage"=>nil, "datawidth"=>64, "description"=>"Intel64 Family 6 Model 62 Stepping 4", "deviceid"=>"CPU0", "errorcleared"=>nil, "errordescription"=>nil, "extclock"=>nil, "family"=>1, "installdate"=>nil, "l2cachesize"=>nil, "l2cachespeed"=>nil, "l3cachesize"=>0, "l3cachespeed"=>0, "lasterrorcode"=>nil, "level"=>6, "loadpercentage"=>0, "manufacturer"=>"GenuineIntel", "maxclockspeed"=>2500, "name"=>"Intel(R) Xeon(R) CPU E5-2670 v2 @ 2.50GHz",
[...]

PR #628, opened 2015-10-02
Fix Windows cpu enumeration, add tests

chefsalim:
Windows CPU enumeration was only looking at the first CPU. Hence the core and CPU counts were incorrect. This change also adds unit tests for the Windows CPU plug-in. This fixes issue#623 (https://github.com/chef/ohai/issues/623)

@adamedx @mcquin @thommay @chef/client-core

### DIFF
--- a/lib/ohai/plugins/windows/cpu.rb
+++ b/lib/ohai/plugins/windows/cpu.rb
@@ -29,7 +29,7 @@ Ohai.plugin(:CPU) do
     wmi = WmiLite::Wmi.new
     processors = wmi.instances_of('Win32_Processor')
 
-    processors.find(:all).each do |processor|
+    processors.each do |processor|
       #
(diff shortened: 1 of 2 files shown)


adamedx:
:+1: . 

Original error was introduced with this change to convert to wmi-lite: https://github.com/chef/ohai/commit/6f8fd202b5d8ef36aad52c85ce45f47bd5c327e7#diff-035c76e96e522d7868f256be066b5d50. Basically the original `find(:all)` syntax before wmi-lite was retained with it, but that was not actually valid, so `find` didn't do what we expected. There were no unit or functional tests to validate this, and the `find` method does not seem to validate its input (hard to see how passing the symbol `:all` should be allowed).

I checked for similar instances in Ohai and in Chef and found none.

smurawski:
:+1:

thommay:
Awesome, thanks @chefsalim

Issue #629, opened 2015-10-05
ohai filesystem plugin output needs adjustment for solaris11


SolverDude:
With solaris 11, each user on a server has their own filesystem mount for their home, using zfs.  This isn't optional.

The present solaris2 filesystem plugin [unchanged in 2 years] in ohai iterates over each zfs filesystem's attributes with this command:
"zfs get -p -H all"

This generates a [very large] chunk of text data for ea zfs mount, On our sample system it's 55 attributes and 2535 bytes of text.  Most of it is default parameters.

When there are many users, the above expands to exceed the 1,000,000 byte limit the Chef server imposes, and thus fails to converge the node.

I would suggest 'all' is a bit too much data.

An alternative would be to select the minimal desired attributes, like so:
"zfs get -p -H mountpoint,creation,available,used"

The above seems the minimal we see as needed for Ohai data, we aren't saying that's the case for all solaris users.

SolverDude:
@hornbeck @steversmith @craigadams Here is the summary of our Ohai issue re solaris11

PR #613, opened 2015-08-24
Plugin configuration

mcquin:
Implement the remaining tasks of [rfc053](https://github.com/chef/chef-rfc/blob/master/rfc053-ohai-config.md), namely:
* Add support for `ohai.plugin` in configuration files.
* Add `configuration` method to plugin DSL.

@chef/client-core

PR #630, opened 2015-10-05
Configurable filesystem properties collected for Solaris2.x

mcquin:
Resolves https://github.com/chef/ohai/issues/629.

@chef/client-core

### DIFF
--- a/lib/ohai/plugins/solaris2/filesystem.rb
+++ b/lib/ohai/plugins/solaris2/filesystem.rb
@@ -63,7 +63,15 @@ Ohai.plugin(:Filesystem) do
 
     # Grab any zfs data from "zfs get"
     zfs = Mash.new
-    so = shell_out("zfs get -p -H all")
+    # ohai.plugin[:filesystem][:zfs_properties] = 'used'
+    # ohai.plugin[:filesystem][:zfs_properties] = ['mountpoint', 'creation', 'available', 'used']
+    zfs_get = "zfs get -p -H "
+    if configuration(:zfs_properties).nil? || configuration(:zfs_properties).empty?
+      zfs_get << "all"
+    else
+      zfs_get << configuration(:zfs_properties).join(',')
+    end
+    so = shell_out(zfs_get)
     so.stdout.lines do |line|
       next unless (line =~ /^([^\t]+)\t([^\t]+)\t([^\t]+)\t([^\t]+)$/)
       filesystem = $1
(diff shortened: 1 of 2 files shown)


btm:
@mcquin, How would this patch be used, would you put something like this in your client.rb?

```
ohai.plugin[:filesystem][:zfs_properties] = [ "used", "available" ]
```

It looks like the recently released ChefDK 0.9.0 is the first release with the plugin system from #613 (Ohai 8.6.0), so that's required for this patch.

@bdha, would this patch solve an issue for you?

lib/ohai/plugins/solaris2/filesystem.rb
```diff
@@ -63,7 +63,15 @@
...
     # Grab any zfs data from "zfs get"
     zfs = Mash.new
-    so = shell_out("zfs get -p -H all")
+    # ohai.plugin[:filesystem][:zfs_properties] = 'used'
+    # ohai.plugin[:filesystem][:zfs_properties] = ['mountpoint', 'creation', 'available', 'used']
+    zfs_get = "zfs get -p -H "
+    if configuration(:zfs_properties).nil? || configuration(:zfs_properties).empty?
+      zfs_get << "all"
+    else
+      zfs_get << configuration(:zfs_properties).join(',')
```
btm: Is there magic elsewhere that would allow zfs_properties to be a string?

mcquin:
[The comment here](https://github.com/chef/ohai/blob/mcquin/ohai-629/lib/ohai/plugins/solaris2/filesystem.rb#L66-L67) is a little non-obvious, but it demonstrates what to place in your client.rb

lib/ohai/plugins/solaris2/filesystem.rb
mcquin: There is now ;)

btm:
Great, which appears congruent with the example I came up with.

:+1:

bdha:
Thanks for looking at this! I've been rolling my own Chef dists for a few years now and this is one of the big reasons why.

I'd argue for swapping what you collect -- be conservative first. Otherwise with the default behavior someone with a lot of snapshots is still going to have a bad experience. `chef-client` will take an excessively long time to run, or simply never complete. (Which is what happens on my systems where I have many snapshots.)

Whereas if they're wondering why they aren't getting all the properties they need, they'll look at the code.

Alternatively, if you only collected `-t filesystem` and `-t volume` (this would require two execs) and avoided snapshots by default, it would also -- potentially -- guard against submitting excessive data to the Chef Server.

btm:
@bdha, I agree on being conservative but we only recently designed [RFC53](https://github.com/chef/chef-rfc/blob/master/rfc053-ohai-config.md) and implemented the ability to easily send configuration values to Ohai. Historically noisy Ohai plugins can move toward more conservative defaults and users that need more can be choosy via configuration.

Switching the filesystem plugin to be conservative on zfs properties for the next major release of Ohai would be a welcome patch.

ksubrama:
:+1:

PR #632, opened 2015-10-09
Add gemspec files to allow bundler to run from the gem

ksubrama:
/cc @jkeiser


ksubrama:
Using jkeiser's +1 in https://github.com/chef/chef-dk/pull/559

PR #640, opened 2015-10-29
Fix and extend CPU detection on FreeBSD

tas50:
The comment block and the tests had an incorrect sample of what
dmesg.boot looked like.  I got 68 different people to send me samples of
what the output looks like, with about 100 samples total.  The Origin
line is always indented and the spacing around the equals
signs differ.  On 10.2+ there are not spaces, but on 10.1 and earlier there are 
spaces.  I updated the regex to properly detect the existing fields and
also added family/model values, which we already gathered on Linux.

### DIFF
--- a/lib/ohai/plugins/freebsd/cpu.rb
+++ b/lib/ohai/plugins/freebsd/cpu.rb
@@ -41,7 +41,7 @@ Ohai.plugin(:CPU) do
       when /CPU:\s+(.+) \(([\d.]+).+\)/
         cpuinfo["model_name"] = $1
         cpuinfo["mhz"] = $2
-      when /\s+Origin="(.*)"\s.*Family=(?:0x)?(\S+)\s.*Model=(?:0x)?(\S+)\s+Stepping=(?:0x)?(\S+)/
+      when /\s+Origin.*"(.*)".*Family.*0x(\S+).*Model.*0x(\S+).*Stepping.*(\S+)/
         cpuinfo["vendor_id"] = $1
         cpuinfo["family"] = $2
         cpuinfo["model"] = $3
--- a/lib/ohai/plugins/freebsd/cpu.rb
+++ b/lib/ohai/plugins/freebsd/cpu.rb
@@ -1,5 +1,6 @@
 #
 # Author:: Bryan McLellan (btm@loftninjas.org)
+# Author:: Tim Smith (tsmith@chef.io)
 # Copyright:: Copyright (c) 2008 Bryan McLellan
 # License:: Apache License, Version 2.0
 #
@@ -43,8 +44,9 @@ Ohai.plugin(:CPU) do
         cpuinfo["mhz"] = $2
       when /Origin.*"(.*)".*Family.*0x(\S+).*Model.*0x(\S+).*Stepping.*(\S+)/
         cpuinfo["vendor_id"] = $1
-        cpuinfo["family"] = $2
-        cpuinfo["model"] = $3
+        # convert from hex value to int, but keep a string to match Linux ohai
+        cpuinfo["family"] = $2.to_i(16).to_s
+        cpuinfo["model"] = $3.to_i(16).to_s
         cpuinfo["stepping"] = $4
         # These _should_ match /AMD Features2?/ lines as well
       when /Features=.+<(.+)>/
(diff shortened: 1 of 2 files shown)


lib/ohai/plugins/freebsd/cpu.rb
```diff
@@ -27,21 +27,25 @@
...
+    #   Structured Extended Features=0x2000<NFPUSG>
+    #   TSC: P-state invariant
 
     File.open("/var/run/dmesg.boot").each do |line|
       case line
       when /CPU:\s+(.+) \(([\d.]+).+\)/
         cpuinfo["model_name"] = $1
         cpuinfo["mhz"] = $2
-      when /Origin = "(.+)"\s+Id = (.+)\s+Stepping = (.+)/
+      when /\s+Origin.*"(.*)".*Family.*0x(\S+).*Model.*0x(\S+).*Stepping.*(\S+)/
```
mcquin: I don't believe checking for the leading whitespace is necessary since we aren't matching from the beginning of the line.

tas50: Good catch.  I removed that portion

mcquin:
:+1:

@chef/client-core

lib/ohai/plugins/freebsd/cpu.rb
```diff
@@ -27,21 +27,25 @@
...
+    #   Structured Extended Features=0x2000<NFPUSG>
+    #   TSC: P-state invariant
 
     File.open("/var/run/dmesg.boot").each do |line|
       case line
       when /CPU:\s+(.+) \(([\d.]+).+\)/
         cpuinfo["model_name"] = $1
         cpuinfo["mhz"] = $2
-      when /Origin = "(.+)"\s+Id = (.+)\s+Stepping = (.+)/
+      when /Origin.*"(.*)".*Family.*0x(\S+).*Model.*0x(\S+).*Stepping.*(\S+)/
```
btm: We shouldn't strip the 0x off these values unless we're going to convert them. 

FreeBSD 10.1-RELEASE VM:
```
CPU: Intel(R) Core(TM) i7-3615QM CPU @ 2.30GHz (2294.87-MHz 686-class CPU)
  Origin = "GenuineIntel"  Id = 0x306a9  Family = 0x6  Model = 0x3a  Stepping = 9
```

Ubuntu 14.04 LTS VM:
```
processor	: 0
vendor_id	: GenuineIntel
cpu family	: 6
model		: 58
model name	: Intel(R) Core(TM) i7-3615QM CPU @ 2.30GHz
stepping	: 9
```

On Ubuntu we're displaying and capturing the family and model as decimal, but on FreeBSD it's being displayed as hex. We should convert that to decimal so their equivalent.

tas50: We probably should just convert the values.  Let me see about adding that.

spec/unit/plugins/freebsd/cpu_spec.rb
```diff
@@ -18,51 +18,96 @@
...
 end
+
+describe Ohai::System, "FreeBSD cpu plugin on FreeBSD <=10.1" do
+  before(:each) do
+    @plugin = get_plugin("freebsd/cpu")
+    allow(@plugin).to receive(:collect_os).and_return(:freebsd)
+    allow(@plugin).to receive(:shell_out).with("sysctl -n hw.ncpu").and_return(mock_shell_out(0, "2", ""))
+    @double_file = double("/var/run/dmesg.boot")
+    allow(@double_file).to receive(:each).
+      and_yield('CPU: Intel(R) Atom(TM) CPU N270   @ 1.60GHz (1596.03-MHz 686-class CPU)').
```
btm: I realize this is basically the same line as in the 10.2+ test, but we should make sure we get model_name and mhz out of this string as well. It's already there, regexes are hard, and here's an opportunity to test against any future modifications someone might make.

tas50: I think we can safely avoid this.  The regex that performs the parsing of this text hasn't been modified and the text doesn't change between 9 and 11.  Our existing test of the regex with the input text catches that just fine.

lib/ohai/plugins/freebsd/cpu.rb
```diff
@@ -1,5 +1,6 @@
 #
 # Author:: Bryan McLellan (btm@loftninjas.org)
+# Author:: Tim Smith (tsmith@chef.io)
 # Copyright:: Copyright (c) 2008 Bryan McLellan
```
btm: lol 2008.

btm:
:+1:

PR #641, opened 2015-10-29
Collect inode data on FreeBSD and add filesystem specs

tas50:
Modified version of the inode collection code from the Linux plugin.
Also add a filesystem spec for FreeBSD which is about 1/3 the specs from
the Linux filesystem spec, but modified for command output on FreeBSD

### DIFF
--- a/lib/ohai/plugins/freebsd/filesystem.rb
+++ b/lib/ohai/plugins/freebsd/filesystem.rb
@@ -44,9 +44,9 @@ Ohai.plugin(:Filesystem) do
     so = shell_out("df -iP")
     so.stdout.lines do |line|
       case line
-      when /^Filesystem\s+Size/
+      when /^Filesystem/
         next
-      when /^(.+?)\s.*\d+\%\s+(\d+)\s+(\d+)\s+(\d+\%)\s+(.+)$/
+      when /^(\S+)\s.+%\s+(\d+)\s+(\d+)\s+(\d+\%)\s+(\S+)/
         filesystem = $1
         fs[filesystem] ||= Mash.new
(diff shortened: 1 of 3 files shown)


mcquin:
:+1:

mcquin:
\cc @chef/client-core

lib/ohai/plugins/freebsd/filesystem.rb
```diff
@@ -39,6 +40,23 @@
       end
     end
 
+    # Grab filesystem inode data from df
+    so = shell_out("df -iP")
+    so.stdout.lines do |line|
+      case line
+      when /^Filesystem\s+Size/
```
btm: I don't see why we have this pattern everywhere, but you want `when /^Filesystem/` here.

lib/ohai/plugins/freebsd/filesystem.rb
```diff
@@ -39,6 +40,23 @@
...
+        next
+      when /^(.+?)\s.*\d+\%\s+(\d+)\s+(\d+)\s+(\d+\%)\s+(.+)$/
```
btm: greedy regex is greedy. :p

btm:
:+1:

lib/ohai/plugins/freebsd/filesystem.rb
tas50: Total cargo cult from Linux here. I'll fix it

lib/ohai/plugins/freebsd/filesystem.rb
tas50: Fixed

PR #642, opened 2015-10-30
Detect vbox/vmware/kvm on Windows and speed up runs

tas50:
Fetch WMI data within Ruby vs. shelling out to Powershell.  This should
be WAY faster.

### DIFF
--- a/spec/unit/plugins/windows/virtualization_spec.rb
+++ b/spec/unit/plugins/windows/virtualization_spec.rb
@@ -1,6 +1,8 @@
 #
 # Author:: Pavel Yudin (<pyudin@parallels.com>)
+# Author:: Tim Smith (<tsmith@chef.io>)
 # Copyright:: Copyright (c) 2015 Pavel Yudin
+# Copyright:: Copyright (c) 2015 Chef Software, Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


tas50:
Going to rework the specs for virtualization so hold off on this one for now

btm:
@tas50 this is cool. you should finish it. :)

also hyperv. See #647, although that's about linux guests.

tas50:
I need to sit down and beef up my rspec skillset to test this.

btm:
VMware Fusion: https://gist.github.com/btm/f044387275ffa277d745

tas50:
@btm This is ready now.  Also it speeds up the Ohai runs by 6% and works if PS isn't installed.

spec/unit/plugins/windows/virtualization_spec.rb
```diff
@@ -18,47 +20,159 @@
...
-SMBIOSBIOSVersion : 10.2.0 (28956) rev 0
-Manufacturer      : Parallels Software International Inc.
-Name              : Default System BIOS
-SerialNumber      : Parallels-92 05 B4 56 97 11 4F FA B1 95 1A FF 8E F9 DD CE
-Version           : PRLS   - 1
-      BIOS
-      shellout = double("shellout")
-      allow(shellout).to receive(:stdout).and_return(bios)
-      allow(plugin).to receive(:shell_out).with('powershell.exe "Get-WmiObject -Class Win32_BIOS"').and_return(shellout)
+  context 'when running on vmware fusion' do
```
btm: vmware fusion or parallels? it looks like two tests got combined by accident here?

tas50: oh looks like I did a select and delete on part of the block before I did the commit :(  Let me push it with the correct test block.

btm:
woot :+1: presuming it all passes. I was hoping you'd shoot for 7 or 8%, but 6% will do. ;)

thommay:
:+1: thanks Tim!

PR #648, opened 2015-11-12
"hostname/fqdn" returns nil for CentOS 6

patcon:
Seems that on a fresh centos 6.7 server on digital ocean, when someone names the server a full FQDN, there is some quirky behavior.

```
[root@jira ~]# hostname
example.blendive.com
[root@jira ~]# hostname --short
example
[root@jira ~]# hostname --fqdn
hostname: Unknown host
[root@jira ~]# hostname --all-fqdns
example.blendive.com
```

It probably has to the with the fact that nothing beyond localhost is set in the hosts file:

```
[root@jira ~]# cat /etc/hosts
127.0.0.1   localhost localhost.localdomain localhost4 localhost4.localdomain4
::1         localhost localhost.localdomain localhost6 localhost6.localdomain6
```

(Setting the server name in the hosts file itself resolves the error. Not sure if that's an issue with how DigitalOcean provisions its images, or if that's expected and done similarly elsewhere.)

The error these circumstances, ohai chooses not to set the fqdn on the node object. I think it would be better to, in case where `hostname --fqdn` fails, that we use the output of `hostname --all-fqdns` which would make for a good last resort.

Thoughts?

### DIFF
--- a/lib/ohai/plugins/hostname.rb
+++ b/lib/ohai/plugins/hostname.rb
@@ -124,6 +124,15 @@ Ohai.plugin(:Hostname) do
         ourfqdn = from_cmd("hostname --fqdn")
       end
 
+      # Some IaaS providers (ie. DigitalOcean) don't set the server name in
+      # /etc/hosts, and so 'hostname --fqdn` fails. However, `hostname
+      # --all-fqdns` might still succeed.
+      if ourfqdn.nil? || ourfqdn.empty?
+        Ohai::Log.debug("hostname --fqdn returned an empty string, retrying " +
+                        "with --all-fqdns flag.")
+        ourfqdn = from_cmd("hostname --all-fqdns")
+      end
+
       if ourfqdn.nil? || ourfqdn.empty?
         Ohai::Log.debug("hostname --fqdn returned an empty string twice and " +
                         "will not be set.")
--- a/lib/ohai/plugins/hostname.rb
+++ b/lib/ohai/plugins/hostname.rb
@@ -124,6 +124,11 @@ Ohai.plugin(:Hostname) do
         ourfqdn = from_cmd("hostname --fqdn")
       end
 
+      if ourfqdn.nil? || ourfqdn.empty?
+        Ohai::Log.debug("hostname --fqdn returned an empty string twice and " +
+                        "will not be set.")
+      end
+
       # Some IaaS providers (ie. DigitalOcean) don't set the server name in
       # /etc/hosts, and so 'hostname --fqdn` fails. However, `hostname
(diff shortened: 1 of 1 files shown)


patcon:
Should be ready for review

patcon:
Hm. Ok, on closer inspection, it seemed my issue stemmed from something larger. Apparently, under certain conditions, fqdn wasn't set at all on the node object. At least in most other places, a failed key results in `nil`. The lack of setting an fqdn key at all seemed to lead to an issue where the fqdn couldn't be refreshed. I need to confirm that.

I could pull these into two separate PRs, if that's preferred.

patcon:
Sorry, I'll split this out. I shouldn't have mashed both together

thommay:
this seems sensible, thanks. Can you whip up a test for when `hostname --fqdn` is empty to confirm the change works, please.

lamont-granquist:
Nope.

This is addressed by the RFC:  https://github.com/chef/chef-rfc/pull/108

FQDN is defined as the DNS lookup name and must not change by default (there is a long argument I had with dan about this somewhere in a PR here).

We /can/ introduce the ability for users to tweak this behavior.  Fundamentally what is being worked around though is bad cookbook code using the `fqdn` attribute when it should use `machinename` or some other attribute which does not fail when DNS is used.  The /right/ solution to this is to fix the broken cookbook code.  The /workaround/ is in the discussion of that RFC which is to let the user set ohai config to override their own behavior of the FQDN attribute without changing it for everyone else in the universe.

patcon:
OK, sounds like this is final. But just to confirm we're throwing all of it out (as it seemed there was a bugfix), `node['fqdn']` is intended not to exist on a fresh image of CentOS?

lamont-granquist:
Yep, node[:fqdn] is defined as the forward+reverse DNS lookup thingy because of unfortunate $HISTORY, which means it is _accurately_ broken on new installs that have broken DNS.

Nobody should use that in cookbooks because its brittle (but everyone does because reaching for 'fqdn' just seems obvious to authors).

We can't change the behavior because of $HISTORY.

The RFC is the compromise way around this all.

At some point we may RFC changing the default behavior during a major version bump, but we definitely can't do it outside of a major version bump.

patcon:
Thanks so much for the context @lamont-granquist :)

lamont-granquist:
Yep, I will completely agree that it all sucks as well, and that RFC turned into way too long of a yak shave and we still haven't shipped it...   I'm not very happy about it all either...   There is a way to dig out of it but its not as simple as a single pull request for $HISTORY and $REASONS...

PR #653, opened 2015-11-25
Add a plugin for Virtualbox details

tas50:
Grabs data from the guest additions if the CLI tool works.  That provides us with the guest additions version/revision and vbox version/revision/lang on the host.

### DIFF
--- a/lib/ohai/plugins/virtualbox.rb
+++ b/lib/ohai/plugins/virtualbox.rb
@@ -1,0 +1,44 @@
+#
+# Author:: "Tim Smith" <tsmith@chef.io>
+# Copyright:: Copyright (c) 2015 Chef Software, Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+Ohai.plugin(:Virtualbox) do
+  provides 'virtualbox'
+
+  collect_data do
+    so = shell_out('VBoxControl guestproperty enumerate')
+    if so.exitstatus == 0
+      virtualbox Mash.new
+      virtualbox[:host] = Mash.new
+      virtualbox[:guest] = Mash.new
+      so.stdout.lines.each do |line|
+        case line
+        when /LanguageID, value: (\S*),/
+          virtualbox[:host][:language] = Regexp.last_match(1)
+        when /VBoxVer, value: (\S*),/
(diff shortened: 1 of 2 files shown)


lib/ohai/plugins/virtualbox.rb
```diff
@@ -0,0 +1,44 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+Ohai.plugin(:Virtualbox) do
+  provides 'virtualbox'
+
+  collect_data do
+    so = shell_out('VBoxControl guestproperty enumerate')
+    if so.exitstatus == 0
```
lamont-granquist: usually i'd do `return unless so.exitstatus == 0`.  avoids indenting all the code.  violates the principle that 'there should be only one return' but i always violate that for sanity checks at the top of a method.

i think you may also be able to just shell_out! and let ohai fail the plugin.

lib/ohai/plugins/virtualbox.rb
```diff
@@ -0,0 +1,44 @@
...
+      virtualbox Mash.new
+      virtualbox[:host] = Mash.new
+      virtualbox[:guest] = Mash.new
+      so.stdout.lines.each do |line|
+        case line
+        when /LanguageID, value: (\S*),/
```
lamont-granquist: can we anchor these with a `^`or a `\A`?

lamont-granquist: Ah, looks like not from the specs...

spec/unit/plugins/virtualbox.rb
```diff
@@ -0,0 +1,87 @@
...
+      plugin = get_plugin('virtualbox')
+      allow(plugin).to receive(:shell_out).with('VBoxControl guestproperty enumerate').and_return(mock_shell_out(1, '', ''))
+      plugin.run
+      expect(plugin).not_to have_key(:virtualbox)
+    end
+  end
+
+  context 'when VBoxControl shellout succeeds' do
+    before(:each) do
+      @plugin = get_plugin('virtualbox')
```
lamont-granquist: should really be `let(:plugin) { get_plugin('virtualbox') }` and kill off the `@` signs

lamont-granquist:
couple of nits that i don't super care about, :+1: overall

lib/ohai/plugins/virtualbox.rb
tas50: when I switched to return unless so.existstatus == 0 the specs failed:

     Failure/Error: return unless so.exitstatus == 0

     LocalJumpError:
       unexpected return
     # ./lib/ohai/plugins/virtualbox.rb:24:in `block (2 levels) in <main>'
     # ./lib/ohai/dsl/plugin/versionvii.rb:90:in `instance_eval'
     # ./lib/ohai/dsl/plugin/versionvii.rb:90:in `run_plugin'
     # ./lib/ohai/dsl/plugin.rb:98:in `run'
     # ./spec/unit/plugins/virtualbox.rb:55:in `block (3 levels) in <top (required)>'

tas50:
@btm or @mcquin thoughts on this?

lib/ohai/plugins/virtualbox.rb
lamont-granquist: what you've got is probably fine then

thommay:
Hit it! :+1:

PR #655, opened 2015-11-26
Dedupe BSD Plugins, create Mixin for parsing DMI Decode, add HyperV detection

tas50:
We have a large amount of duplication in the BSD platforms.  Due to this we haven't applied changes to all BSD platforms where we should have.  This collapses the filesystem and virtualization plugins into individual plugins.  It reduces duplicate code and improves functionality on netbsd/freebsd/dragonflybsd.  Additionally the DMI output on BSD and Linux hosts is the same, but we parse it differently on each platform.  This breakout out the dmidecode parsing into a mixin that we can use on both platforms.

Virtualization improvements:
- Bring BSD virtualization detection in line with Linux.  This adds Openstack, KVM, and Vbox detection
- Add the "systems" format used in the Linux/Darwin plugisn where you can have multiple host / guests
- Only create the virtualization mash if it hasn't been created already so we don't overwrite data from the platform agnostic virtualization plugin
- Improves detection of VirtualPC guests and adds detection of guests on Hyper-V


tas50:
@btm I'd love some feedback on moving the dmidecode logic out of linux and freebsd as well as how to test the collapsed BSD code.

lamont-granquist:
Yeah I was gonna say that the only *BSD-specific code in the parsing-dmidecode function might be in the specific output formation of dmidecode on *BSD.  The values that you pull out of dmidecode should be the same on *BSD and Linux.  I think I'm okay with doing the cleanup iteratively though and shipping this for *BSD and then having the work to merge *BSD and Linux happen.

lamont-granquist:
:+1:

thommay:
:+1:

smurawski:
:+1: This won't catch all the hyper-v versions, but we can add more later.

PR #660, opened 2015-11-30
Add automated changelog, bump version

chefsalim:
This PR is getting up ready to release Ohai 8.8.0
1. It adds infra to start using an automated process to generate change logs
2. It bumps up the Ohai release version to 8.8.0
3. Note: The CHANGELOG.md file has a bunch of re-writes due to being auto-generated. In the future, the changes will only be the ones that have been made since the last release.  The original change log has been presevered in CHANGELOG_OLD.md.

@lamont-granquist @tas50 @btm @jkeiser

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,254 +1,849 @@
-# Ohai Changelog
+# Change Log
 
-## Unreleased
-* [**James Flemer, NDP LLC**](https://github.com/jflemer-ndp):
-  - Add detection for RHEV (on Linux guests) to virtualization plugin
-* [**Shahul Khajamohideen**](https://github.com/sh9189):
-  - Fixes Windows :CPU plugin inconsistencies with other platforms: modifies
-  `cpu[:total]` to return total number of logical processors, adds `cpu[:cores]`
-  to return total number of cores.
-* [**clewis**](https://github.com/clewis):
-  - Don't constrain the width of `ps` output.
-* [**George Gensure**](https://github.com/werkt):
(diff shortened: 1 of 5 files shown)


lamont-granquist:
HISTORY.md is the conventional old changlog in github-changelog-generator.

tas50:
Is there a way to turn off the closed issues and is there a guide somewhere on how to get things to show up as enhancements vs. plain old merges?

lamont-granquist:
Yeah `config.issues = false` will remove the closed issues which we should definitely use.

And to expand on my earlier comment, we should preserve the old changelog in the new changelog by using HISTORY.md and setting `config.since_tag = 'v8.7.0'`

The only problem there is that you'll immediately run into https://github.com/skywinder/github-changelog-generator/issues/304

lamont-granquist:
@tas50 if you tag things as 'bug' or 'enhancement' and re-run the rake task it'll regenerate the changelog with the merges moved into the correct category.

chefsalim:
@lamont-granquist Yeah, the wierdness with using the since tag is the reason why we removed it for now, and used CHANGELOG_OLD.md instead.  I can rename it to HISTORY.md (and config.issues = false) if that helps getting this approved.

lamont-granquist:
I'd really like to see that bug fixed and do it properly.

tas50:
I renamed some PRs and added labels.  If you regenerate the changelog it should look better now

lamont-granquist:
to be clear, :-1: until we do this right.

PR #662, opened 2015-12-01
Only poll vmware attributes when on vmware

tas50:
This will prevent Azure guests that contain the vmware toolbox CLI from being identified as vmware hosts.  We shouldn't even try to grab vmware values unless dmidecode has identified the host as vmware.  Added specs for when the CLI is missing and when we're not running on vmware.

### DIFF
--- a/lib/ohai/plugins/vmware.rb
+++ b/lib/ohai/plugins/vmware.rb
@@ -1,7 +1,7 @@
 #
-# Author:: "Dan Robinson" <drobinson@getchef.com>
+# Author:: "Dan Robinson" <drobinson@chef.io>
 # Author:: "Christopher M. Luciano" <cmlucian@us.ibm.com>
-# Copyright:: Copyright (c) 2014 Chef Software, Inc.
+# Copyright:: Copyright (c) 2014-2015 Chef Software, Inc.
 # Copyright (C) 2015 IBM Corp.
 # License:: Apache License, Version 2.0
 #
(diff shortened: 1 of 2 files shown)


tas50:
Tested on Azure and we no longer get a pile of empty attributes.  Tested on vmware fusion with both the compiled vmware tools and the open tools available via apt and both do what you'd expect.

lib/ohai/plugins/vmware.rb
```diff
@@ -40,28 +41,30 @@ def from_cmd(cmd)
...
-      vmware Mash.new
-      begin
-        # vmware-toolbox-cmd stat <param> commands
-        # Iterate through each parameter supported by the "vwware-toolbox-cmd stat" command, assign value
-        # to attribute "vmware[:<parameter>]"
-        [ "hosttime", "speed", "sessionid", "balloon", "swap", "memlimit", "memres", "cpures", "cpulimit" ].each do |param|
-          vmware[param] = from_cmd("#{vmtools_path} stat #{param}")
-          if vmware[param] =~ /UpdateInfo failed/
-            vmware[param] = nil
+    if virtualization[:systems][:vmware]
```
mcquin: You can avoid indenting the entirety of this method by pairing this check with the call to the method:

```ruby
collect_data(:linux) do
  get_vm_attributes("/usr/bin/vmware-toolbox-cmd") if virtualization[:systems][:vmware]
end
```

mcquin:
LGTM. (https://github.com/chef/ohai/pull/662#discussion_r46704444 is only a suggestion)

tas50:
Much simpler.  Rebased with that @mcquin

mcquin:
Awesome. Looks good!

thommay:
:+1:

PR #670, opened 2015-12-04
Added 2 plugins we've written to collect

knightorc:
Installed windows features (installed_features). Valid for Win7,Win2K8,Win2K12
Installed programs (installed_programs).  Valid for Win7,Win2K8,Win2K12

### DIFF
--- a/lib/ohai/plugins/windows/features.rb
+++ b/lib/ohai/plugins/windows/features.rb
@@ -1,0 +1,20 @@
+Ohai.plugin(:Features) do
+  provides "installed_features"
+  collect_data(:default) do
+    installed_features Mash.new
+    # Grab raw feature information
+    raw_list_of_features = shell_out("C:\\Windows\\sysnative\\dism.exe /Get-Features /Online /Format:Table").stdout
+    # Remove Quotes and Split into an array
+    features_list = raw_list_of_features.split("\r\n")
+    features_list.each do | feature_details_raw |
+    # Check for Enabled or Enable Pending (Reboot Needed)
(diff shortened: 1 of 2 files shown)


lib/ohai/plugins/windows/features.rb
```diff
@@ -0,0 +1,20 @@
+Ohai.plugin(:Features) do
+  provides "installed_features"
+  collect_data(:default) do
```
mcquin: It's best practice to restrict plugins which collect platform-specific data to the specific platform. You can tell Ohai to collect this data only on Windows by replacing `:default` (runs on all platforms) with `:windows`.

smurawski:
@knightorc If you restrict the platform as described by mccquin and drop the installed_packages part (we just merged something to cover that same use case), we can get the installed_features added.

A spec or two wouldn't hurt.

lib/ohai/plugins/windows/features.rb
```diff
@@ -0,0 +1,20 @@
...
+    installed_features Mash.new
+    # Grab raw feature information
+    raw_list_of_features = shell_out("C:\\Windows\\sysnative\\dism.exe /Get-Features /Online /Format:Table").stdout
```
smurawski: You should probably grab the Windows directory location from %WINDIR% in the case where Windows is installed to a non "c" drive.

tas50:
We recently merged an optional plugin that shipped with 12.7.2 that reports installed applications on Windows.  I'm going to close this out since I think it's redundant at this point.  Feel free to open it back up if I'm missing something.

Issue #671, opened 2015-12-05
CPU counts on OS X are wrong


tas50:
On OS X we report the following aggregate CPU information on a Retina MBP:
    "real": 4,
    "total": 8,

The problem is this system has 1 real cpu, 4 total (real cores), and 8 cores (hyperthreaded cores).

we use sysctl -n hw.physicalcpu for real CPUs which is actually real cores not CPUs.

If we want the number of physical CPUs we want hw.packages

tas50:
Turns out we're not doing this wrong.  We're just not reporting the real CPU.

PR #672, opened 2015-12-05
Correctly count real vs. total vs. cores on darwin

tas50:
We were counting CPUs wrong.  See #671 for details

### DIFF
--- a/lib/ohai/plugins/darwin/cpu.rb
+++ b/lib/ohai/plugins/darwin/cpu.rb
@@ -1,7 +1,7 @@
 #
 # Author:: Nathan L Smith (<nlloyds@gmail.com>)
-# Author:: Tim Smith (<tsmith@limelight.com>)
-# Copyright:: Copyright (c) 2013 Opscode, Inc.
+# Author:: Tim Smith (<tsmith@chef.io>)
+# Copyright:: Copyright (c) 2013-2015 Chef Software, Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


mcquin:
:+1: 

ping @chef/client-core

lib/ohai/plugins/darwin/cpu.rb
```diff
@@ -22,10 +22,12 @@
...
   collect_data(:darwin) do
     cpu Mash.new
-    so = shell_out("sysctl -n hw.physicalcpu")
+    so = shell_out("sysctl -n hw.packages")
     cpu[:real] = so.stdout.to_i
-    so = shell_out("sysctl -n hw.logicalcpu")
+    so = shell_out("sysctl -n hw.physicalcpu")
     cpu[:total] = so.stdout.to_i
+    so = shell_out("sysctl -n hw.logicalcpu")
+    cpu[:cores] = so.stdout.to_i
```
lamont-granquist: well, this is confusing.  at first glance it seems backwards from what i'd expect.

"cores" should be physical cores -- i would expect '4' here
"total" should be the logical cores things like the scheduler sees - i would expect '8' here
"real" should be how many packages i haven plugged into the motherboard - i would expect '1' here

i haven't looked at how we define these on linux or what the APIs return on darwin, but your cores and total read backwards from what i would expect.

(and definitely include some comments if things really need to be confusing).

lamont-granquist: checked the darwin APIs and they're the way i assumed they be from the names...  so the only remaining question is if our ohai APIs are backwards from what i think they should mean...

lamont-granquist:
So on linux the 'total' is the thing that the scheduler sees, so all the hyperthreaded cores aka all the entries in /proc/cpuinfo.  So i'm pretty certain 'total' here should match and be `hw.logicalcpu`

lamont-granquist:
And linux only defines 'real' as the sum of the unique physical_id's which should be the number of sockets.

So i'm pretty sure linux behaves the way that I outlined above.

tas50:
Updated to swap the cores vs. total logic

lamont-granquist:
:+1:

PR #674, opened 2015-12-05
Determine real, total, and cores for BSD

tas50:
Use dmesg to correctly determine the number of real CPUs, the number of
real cores, and the number of HT cores.  This aligns CPU count data with
what we collect on Linux.

### DIFF
--- a/lib/ohai/plugins/freebsd/cpu.rb
+++ b/lib/ohai/plugins/freebsd/cpu.rb
@@ -36,6 +36,9 @@ Ohai.plugin(:CPU) do
     #   AMD Features2=0x21<LAHF,ABM>
     #   Structured Extended Features=0x2000<NFPUSG>
     #   TSC: P-state invariant
+    #   ...
+    #   FreeBSD/SMP: Multiprocessor System Detected: 8 CPUs
+    #   FreeBSD/SMP: 1 package(s) x 4 core(s) x 2 SMT threads
 
     File.open("/var/run/dmesg.boot").each do |line|
       case line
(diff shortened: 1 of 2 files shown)


arachnist:
dmesg contents are volatile.
Relying on them is bound to produce errors or undefined behaviour sooner or later.

kev009:
It should be fairly safe between majors

lamont-granquist:
@arachnist i raised that question when we saw the same patch for dragonflybsd, but bsd users couldn't come up with a better API to query than dmesg...

lib/ohai/plugins/freebsd/cpu.rb
```diff
@@ -54,13 +57,14 @@
         # Features2=0x80000001<SSE3,<b31>>
       when /Features2=[a-f\dx]+<(.+)>/
         cpuinfo["flags"].concat($1.downcase.split(','))
-      when /Logical CPUs per core: (\d+)/
-        cpuinfo["cores"] = $1
+      when /FreeBSD\/SMP: Multiprocessor System Detected: (\d*) CPUs/
+        cpuinfo["total"] = $1.to_i
+      when /FreeBSD\/SMP: (\d*) package\(s\) x (\d*) core\(s\)/
+        cpuinfo["real"] = $1.to_i
+        cpuinfo["cores"] = $2.to_i
```
lamont-granquist: doesn't this need to be $1.to_i * $2.to_i?

tas50: Ruby did the right thing, but I should add that incase it doesn't

tas50: @lamont-granquist do you mean the total should be real * cores?  The line in dmesg above actually gives the number of cores including HT cores so we don't need to do the math.

lamont-granquist: No, I'm talking about the 'cores' computation.  If this is a dual-socket quad-proc shouldn't that look like:

```
FreeBSD/SMP: Multiprocessor System Detected: 16 CPUs
FreeBSD/SMP: 2 package(s) x 4 core(s) x 2 SMT threads
```

and cpuinfo["cores"] should be 8 (2 x 4)

tas50: yep.  let me fix that one too

lamont-granquist:
:+1:

mcquin:
LGTM

PR #675, opened 2015-12-05
WIP: Expand Windows VM detection to match Linux/BSD

tas50:
Copy in the product names and manufacturers that we've already verified
on the *nix side


lamont-granquist:
:+1: modulo whatever is making travis/appveyor angry

tas50:
This is entirely broken since DMI data in windows is broken into a whole pile of different WMI calls.  We can't do the simple shotgun blast regex sorta logic we do in Linux.  I'm going to need to find hosts on each of these platforms and get WMI data.  It needs to happen, but it's a larger project and this PR doesn't do it right.

PR #677, opened 2015-12-07
The mime must die

jaym:
We don't use this gem and pinning it is causing problems
with other things (chef-server). The issue is that
chef-server-ctl things are using rest-client, which requires
mime-types < 3.

cc @chef/client-core

### DIFF
--- a/ohai.gemspec
+++ b/ohai.gemspec
@@ -15,7 +15,6 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = ">= 2.0.0"
 
-  s.add_dependency "mime-types", "~> 3.0"
   s.add_dependency "systemu", "~> 2.6.4"
   s.add_dependency "ffi-yajl", "~> 2.2"
   s.add_dependency "mixlib-cli"


tylercloke:
This will unblock the server :+1:

coderanger:
mime

lamont-granquist:
Someone out there /might/ be using mime types in their own cookbook code and accidentally getting it pulled in via this line, and it might then break and they might have to add chef_gem and/or a require.

IMO we need to break that person, apologize for breaking them in a minor release and explain how they can fix their issue, rather than continue to cause gem conflict chaos from pinning to ancient versions of mime-types.

Expecting us to continue to maintain gem deps that we don't even use anymore is a perverse reading of SemVer.  The root bug here is that the code which this could break is buggy because it does not correctly declare its deps.  We also have not and would not document our inclusion of mime-types as a public facing API of chef or ohai.

So :shipit:

btm:
Poor little mimes. Another casualty in the war against rubygems dependency trees.

danielsdeleo:
Yup, we shoulda removed this in Chef 12, it was never a direct dep, we just needed to pin it because of SemVer shenanigans. :+1: 

Also I think the risk of breakage is really low, since you need to be doing some kind of network communication for the gem to be useful, and you'd probably be using a gem that actually implements your protocol in that case.

PR #678, opened 2015-12-08
better tests for /proc/cpuinfo

lamont-granquist:
actually tests more than a single socket/core/threaded scenario


tas50:
Yes.  Do it

PR #682, opened 2015-12-10
ipaddress on Linux - default route pointing to unaddressed interface, with route src

glennmatthews:
### Scenario:
```
bash-4.3# ifconfig
Gi0_0_0_0 Link encap:Ethernet  HWaddr 52:46:0b:98:d2:51  
          inet addr:192.168.122.222  Mask:255.255.255.0
          inet6 addr: fe80::5046:bff:fe98:d251/64 Scope:Link
          UP RUNNING NOARP MULTICAST  MTU:1514  Metric:1
          RX packets:30965 errors:0 dropped:0 overruns:0 frame:0
          TX packets:3 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000 
          RX bytes:45928110 (43.8 MiB)  TX bytes:210 (210.0 B)

Mg0_RP0_CPU0_0 Link encap:Ethernet  HWaddr 52:46:07:10:c6:14  
          inet addr:10.1.1.1  Mask:255.255.255.0
          inet6 addr: fe80::5046:7ff:fe10:c614/64 Scope:Link
          UP RUNNING NOARP MULTICAST  MTU:1514  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:3 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000 
          RX bytes:0 (0.0 B)  TX bytes:210 (210.0 B)

fwd_ew    Link encap:Ethernet  HWaddr 00:00:00:00:00:0b  
          inet6 addr: fe80::200:ff:fe00:b/64 Scope:Link
          UP RUNNING NOARP MULTICAST  MTU:1500  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:2 errors:0 dropped:1 overruns:0 carrier:0
          collisions:0 txqueuelen:1000 
          RX bytes:0 (0.0 B)  TX bytes:140 (140.0 B)

fwdintf   Link encap:Ethernet  HWaddr 00:00:00:00:00:0a  
          inet6 addr: fe80::200:ff:fe00:a/64 Scope:Link
          UP RUNNING NOARP MULTICAST  MTU:1496  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:8916 errors:0 dropped:1 overruns:0 carrier:0
          collisions:0 txqueuelen:1000 
          RX bytes:0 (0.0 B)  TX bytes:631885 (617.0 KiB)

lo        Link encap:Local Loopback  
          inet addr:127.0.0.1  Mask:255.0.0.0
          inet6 addr: ::1/128 Scope:Host
          UP LOOPBACK RUNNING  MTU:65536  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:0 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:0 
          RX bytes:0 (0.0 B)  TX bytes:0 (0.0 B)

lo:0      Link encap:Local Loopback  
          inet addr:1.1.1.1  Mask:255.255.255.255
          UP LOOPBACK RUNNING  MTU:65536  Metric:1

lo:2      Link encap:Local Loopback  
          inet addr:2.2.2.2  Mask:255.255.255.255
          UP LOOPBACK RUNNING  MTU:65536  Metric:1
```

```
bash-4.3# ip route show
default dev fwdintf  scope link  src 192.168.122.222 
10.1.1.0/24 dev Mg0_RP0_CPU0_0  proto kernel  scope link  src 10.1.1.1 
```

### Expected Result:
`ohai ipaddress` should report 192.168.122.222 (the configured `src` for the default route) and `ohai network` should include `"default_interface": "fwdintf"`

### Actual Result:
`ohai ipaddress` ignores the default route (because of existing logic that says 'ignore default route if its src is not an address configured on the default interface') and reports 1.1.1.1 (lowest global scope IP address found). No value is reported for `default_interface`.

### Fix
If the default route src is not a configured address on the default interface, but the default interface has *no* addresses of this type, then do not ignore the default route when determining the ipaddress.

### DIFF
--- a/lib/ohai/plugins/linux/network.rb
+++ b/lib/ohai/plugins/linux/network.rb
@@ -319,20 +319,44 @@ Ohai.plugin(:Network) do
 
   # returns the macaddress for interface from a hash of interfaces (iface elsewhere in this file)
   def get_mac_for_interface(interfaces, interface)
-<<<<<<< ac9f5a2ae4ada9f2e6e821418ed3a9039bda9a91
-    interfaces[interface][:addresses].select { |k, v| v["family"] == "lladdr" }.first.first unless interfaces[interface][:flags].include? "NOARP"
-=======
-    interfaces[interface][:addresses].select{|k,v| v["family"]=="lladdr"}.first.first unless interfaces[interface][:addresses].nil? || interfaces[interface][:flags].include?("NOARP")
->>>>>>> Report an ipaddress in a Linux corner case observed on Cisco IOS XR
+    interfaces[interface][:addresses].select { |k, v| v["family"] == "lladdr" }.first.first unless interfaces[interface][:addresses].nil? || interfaces[interface][:flags].include?("NOARP")
   end
 
   # returns the default route with the lowest metric (unspecified metric is 0)
   def choose_default_route(routes)
-    default_route = routes.select do |r|
+    routes.select do |r|
       r[:destination] == "default"
     end.sort do |x, y|
       (x[:metric].nil? ? 0 : x[:metric].to_i) <=> (y[:metric].nil? ? 0 : y[:metric].to_i)
     end.first
+  end
+
+  def interface_has_no_addresses_in_family?(iface, family)
+    return true if iface[:addresses].nil?
+    iface[:addresses].values.all? { |addr| addr['family'] != family }
+  end
+
+  def interface_have_address?(iface, address)
+    return false if iface[:addresses].nil?
+    iface[:addresses].key?(address)
+  end
+
+  def interface_address_not_link_level?(iface, address)
+    iface[:addresses][address][:scope].downcase != "link"
+  end
+
+  def interface_valid_for_route?(iface, address, family)
+    return true if interface_has_no_addresses_in_family?(iface, family)
+
+    interface_have_address?(iface, address) && interface_address_not_link_level?(iface, address)
+  end
+
+  def route_is_valid_default_route?(route, default_route)
+    # if the route destination is a default route, it's good
+    return true if route[:destination] == "default"
+
+    # the default route has a gateway and the route matches the gateway
+    !default_route[:via].nil? && IPAddress(route[:destination]).include?(IPAddress(default_route[:via]))
   end
 
   # ipv4/ipv6 routes are different enough that having a single algorithm to select the favored route for both creates unnecessary complexity
@@ -341,35 +365,19 @@ Ohai.plugin(:Network) do
   def favored_default_route(routes, iface, default_route, family)
     routes.select do |r|
       if family[:name] == "inet"
-        # selecting routes for ipv4
-        # using the source field when it's specified :
-        # 1) in the default route
-        # 2) in the route entry used to reach the default gateway
-        r[:src] && # it has a src field
-          iface[r[:dev]] && # the iface exists
-          (
-            iface[r[:dev]][:addresses].nil? || # this int has no addresses
-            iface[r[:dev]][:addresses].values.all? { |addr| addr['family'] != family[:name] } || # this int has no ip
-            (
-              iface[r[:dev]][:addresses].has_key?(r[:src]) && # the src ip is set on the node
-              iface[r[:dev]][:addresses][r[:src]][:scope].downcase != "link" # this isn't a link level address
-            )
-          ) && (
-            r[:destination] == "default" ||
-            (
-              default_route[:via] && # the default route has a gateway
-              IPAddress(r[:destination]).include?(IPAddress(default_route[:via])) # the route matches the gateway
-            )
-          )
+        # the route must have a source field
+        return false unless defined?(r[:src])
+
+        route_interface = iface[r[:dev]]
+
+        # the interface specified in the route must exist
+        return false unless defined?(route_interface) # the interface specified in the route must exist
+
+        interface_valid_for_route?(route_interface, r[:src], "inet") && route_is_valid_default_route?(r, default_route)
       elsif family[:name] == "inet6"
-        # selecting routes for ipv6
-        iface[r[:dev]] and # the iface exists
-          iface[r[:dev]][:state] == "up" and # the iface is up
-          ( r[:destination] == "default" or
-            ( default_route[:via] and # the default route has a gateway
-              IPAddress(r[:destination]).include? IPAddress(default_route[:via]) # the route matches the gateway
-              )
-            )
+        iface[r[:dev]] &&
+          iface[r[:dev]][:state] == "up" &&
+          route_is_valid_default_route?(r, default_route)
       end
     end.sort_by do |r|
       # sorting the selected routes:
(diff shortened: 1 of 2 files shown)


lib/ohai/plugins/linux/network.rb
```diff
@@ -109,7 +109,15 @@ def check_routing_table(family, iface)
 
         # a sanity check, especially for Linux-VServer, OpenVZ and LXC:
         # don't report the route entry if the src address isn't set on the node
-        next if route_entry[:src] and not iface[route_int][:addresses].has_key? route_entry[:src]
+        # unless the interface has no addresses of this type at all
+        if route_entry[:src]
+          addr = iface[route_int][:addresses]
+          unless addr.nil? or addr.has_key? route_entry[:src] or
+                 not addr.values.any? { |a| a['family'] == family[:name] }
```
mcquin: Please use symbolic operators `||` and `!` for boolean calculations.

lib/ohai/plugins/linux/network.rb
```diff
@@ -355,8 +364,13 @@ def parse_ip_addr_inet6_line(cint, iface, line)
...
               iface[r[:dev]] and # the iface exists
-              iface[r[:dev]][:addresses].has_key? r[:src] and # the src ip is set on the node
-              iface[r[:dev]][:addresses][r[:src]][:scope].downcase != "link" and # this isn't a link level addresse
+              ( iface[r[:dev]][:addresses].nil? or # this int has no addresses
+                (
+                  ( iface[r[:dev]][:addresses].has_key? r[:src] and # the src ip is set on the node
+                    iface[r[:dev]][:addresses][r[:src]][:scope].downcase != "link" # this isn't a link level addresse
+                  ) or not iface[r[:dev]][:addresses].values.any? { |addr| addr['family'] == family[:name] } # this interface has no ip
+                )
+              ) and
```
mcquin: Please use symbolic operators here, as well.

lib/ohai/plugins/linux/network.rb
glennmatthews: Sorry - I was following the convention of the existing code. Will rework my diffs.

mcquin: Thank you! =)

(I'll likely submit a rework of this code once your changes go through to remove the remaining instances.)

glennmatthews:
Changed to boolean operators as requested.

glennmatthews:
Hi, is anything needed to move this along?

mcquin:
Sorry for the delay @glennmatthews . It looks like there have been some recent changes to these files and we cannot cleanly merge. Would you please rebase this branch on latest master?

glennmatthews:
Yup looks like a conflict with the refactoring done by 58605b2a8f7044e1a3d9a3cd3b82eaf0e0dad716. I'll see about getting it fixed.

glennmatthews:
@mcquin Rebased and reworked the diffs; tests are again passing. Thanks!

thommay:
@btm you touched it last ;)

lib/ohai/plugins/linux/network.rb
```diff
@@ -332,15 +341,22 @@ def favored_default_route(routes, iface, default_route, family)
...
+              iface[r[:dev]][:addresses].has_key?(r[:src]) && # the src ip is set on the node
+              iface[r[:dev]][:addresses][r[:src]][:scope].downcase != "link" # this isn't a link level address
+            )
+          ) && (
+            r[:destination] == "default" ||
+            (
+              default_route[:via] && # the default route has a gateway
+              IPAddress(r[:destination]).include?(IPAddress(default_route[:via])) # the route matches the gateway
             )
+          )
```
lamont-granquist: this logic is kind of horrible to read.  the comments are useful.

could we extract some helpers out of this though in order to make the logic more readable?

a trivial example:

```ruby
def iface_has_no_addresses?(iface)
  iface[iface][:addresses].nil?
end
```

glennmatthews:
@lamont-granquist Sorry - were you asking me to refactor the code for readability? Or was that a general observation that this logic is quite messy (which I agree with, but predates my changes)?

lamont-granquist:
yeah refactoring the code for readability would be good.

lamont-granquist:
essentially, if you look where you have comments, if you could make the code read like the comments via helpers...  

then the helpers become bite-size chunks that you can look at and validate that they do what it says on the tin.  then you look at the overall algorithm and make sure it reads correctly...

glennmatthews:
OK, let me ask a different way - can this code be merged as-is, or will it not be accepted until I find the time to refactor it and resubmit? :smile:

lamont-granquist:
yeah, i'd really like to see it more readable before accepting.  otherwise i strongly suspect it won't get cleaned up, and then since this is the linux networking plugin, in 2-6 months after i've forgotten it, i'll have to waste a lot of time deciphering it when the next major PR comes in against it to fix some bug in the code...

btm:
@thom fwiw, I left it better than I got it. :)

I'll put this on the side of my desk and will help if it's still here when I find some time.

glennmatthews:
@adamleff will also be looking into the refactor later this week. Thanks Adam!

adamleff:
@mcquin / @lamont-granquist : would you mind giving this another look? I cleaned up `#favored_default_route` and think it's a whole bunch cleaner now. Also added tests for the new methods I added, and rebased against master.

/cc: @glennmatthews

glennmatthews:
:+1: Nice work Adam!

adamleff:
Rebased on master again due to the conflict created by the chefstyle work yesterday. Should be good to go for another review, @mcquin and @lamont-granquist -- thank you!

lib/ohai/plugins/linux/network.rb
```diff
@@ -310,46 +319,65 @@ def parse_ip_addr_inet6_line(cint, iface, line)
...
-          iface[r[:dev]] and # the iface exists
-          iface[r[:dev]][:addresses].has_key? r[:src] and # the src ip is set on the node
-          iface[r[:dev]][:addresses][r[:src]][:scope].downcase != "link" and # this isn't a link level addresse
-          ( r[:destination] == "default" or
-            ( default_route[:via] and # the default route has a gateway
-              IPAddress(r[:destination]).include? IPAddress(default_route[:via]) # the route matches the gateway
-              )
-            )
+        # the route must have a source field
+        return false unless defined?(r[:src])
```
mcquin: `r[:src]` will return `nil` if `:src` is not a key in the hash. It appears that `defined?(nil)` is truthy. Maybe you want `unless r[:src].nil?` or `.empty?`.

adamleff: @mcquin: That's a great find, because it actually uncovered an unrelated issue. The "returns" I was using was returning the whole method, not the individual `#select` eval block. When I implemented your fix, the tests uncovered unexpected results.

So thanks for that! I changed the logic and we should be good-to-go.

mcquin:
Code looks good to me. Thanks for the refactor @adamleff

adamleff:
@lamont-granquist if you get a chance today, can you take a peek at this and make sure it aligns with what you were thinking in terms of a refactor?  thank you :)

lamont-granquist:
yes, substantially more readable :+1:

lamont-granquist:
@chef/client-core any moar eyeballs on this?

thommay:
Looks like a great refactor, thanks! :+1:

PR #683, opened 2015-12-13
Handle old kernels where we can't determine the init system

tas50:
New kernels will always have /proc/1/comm, but older kernels don't.
Those systems are all init based (debian 6, ubuntu 10.04, centos 5.11).
A good number of these systems are EoL, but some like CentOS 5.11 are
going to be around for several more years.  We should probably support
detecting init on them properly.

### DIFF
--- a/lib/ohai/plugins/init_package.rb
+++ b/lib/ohai/plugins/init_package.rb
@@ -1,6 +1,6 @@
 #
 # Author:: Caleb Tennis (<caleb.tennis@gmail.com>)
-# Copyright:: Copyright (c) 2012 Opscode, Inc.
+# Copyright:: Copyright (c) 2012-2015 Chef Software, Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


lib/ohai/plugins/init_package.rb
```diff
@@ -20,12 +20,6 @@
...
 
   collect_data(:linux) do
-    package_name = nil
-
-    if File.exists?("/proc/1/comm")
-      package_name = File.open("/proc/1/comm").gets.chomp
-    end
-
-    init_package package_name
+    init_package File.exists?("/proc/1/comm") ? File.open("/proc/1/comm").gets.chomp : 'init'
```
thommay: any reason you can't/shouldn't just set `package_name` to `init` up top and leave the rest of the code untouched? I think it's probably a fair chunk more readable.

thommay:
one nit, but essentially :+1:

lib/ohai/plugins/init_package.rb
tas50: I'm not a huge fan of default values that you potentially change later on.  Without a comment it's often hard to understand why the initial value is set to something like init.  When you explicitly say if this file doesn't have content then use init it becomes a bit more clean to me that you chose init due to another value.

tas50:
ping @chef/client-core for review

lamont-granquist:
:+1:

lamont-granquist:
there should be similar code in the platform helpers in core chef as well...

PR #684, opened 2015-12-14
Remove old cucumber tests

tas50:
These haven't been updated in 7 years and don't actually run


mcquin:
:boom: 

:+1:

thommay:
burn it with :fire:

Issue #685, opened 2015-12-16
network plugin is broken on v6 only hosts


jaymzh:
* It can't determine `macaddress` largely because the code doesn't try:

```
if family[:name] == "inet"
  ipaddress route[:src]
  macaddress iface[route[:dev]][:addresses].select{|k,v| v["family"]=="lladdr"}.first.first unless iface[route[:dev]][:flags].include? "NOARP"
  else
    ip6address route[:src]
  end
end
```

Talk about epic clowntown. But even when you ammend that to pick up macaddress there, it won't work and this is because above that it does some route selection magic:

```
          route = routes.select do |r|
            # selecting routes
            r[:src] and # it has a src field
              iface[r[:dev]] and # the iface exists
              iface[r[:dev]][:addresses].has_key? r[:src] and # the src ip is set on the node
              iface[r[:dev]][:addresses][r[:src]][:scope].downcase != "link" and # this isn't a link level addresse
              ( r[:destination] == "default" or
                ( default_route[:via] and # the default route has a gateway
                  IPAddress(r[:destination]).include? IPAddress(default_route[:via]) # the route matches the gateway
                  )
                )
          end.sort_by do |r|
```

Which doesn't work because there is no `src` on your average v6 routes:

```
default via fe80::290:27ff:fe9d:6015 dev eth0  proto ra  metric 1024  expires 1762sec hoplimit 64 pref medium
```

So that code doesn't even come close to working. Therefore there's no route to look at so that mac-address selection code can't possibly work.

jaymzh:
This horribleness makes it sorta work for me... but I"m very unclear on 1. How network.rb and linux/network.rb are working together (and somehow both touch the same attr which is crazy to me) or 2. How this routewalking is supposed to work... but maybe this can get someone started:


```
[phild@dev2373 (master) plugins]$ git diff
diff --git a/lib/ohai/plugins/linux/network.rb b/lib/ohai/plugins/linux/network.rb
index 689001d..7e436b7 100644
--- a/lib/ohai/plugins/linux/network.rb
+++ b/lib/ohai/plugins/linux/network.rb
@@ -406,12 +406,19 @@ Ohai.plugin(:Network) do
             ]
           end.first
 
-          unless route.nil? or route.empty?
+          if route and !route.empty?
             if family[:name] == "inet"
               ipaddress route[:src]
[...]

PR #689, opened 2015-12-18
WIP: Fix detection of mac address on IPv6 systems

tas50:
We failed to detect the mac address when lo has ipv4/ipv6, but all other interfaces only have ipv6. This is just a simplified version of Phil's code from #685 with specs added.

### DIFF
--- a/lib/ohai/plugins/linux/network.rb
+++ b/lib/ohai/plugins/linux/network.rb
@@ -138,6 +138,7 @@ Ohai.plugin(:Network) do
     end.compact.flatten
   end
 
+  # determine layer 1 details for the interface using ethtool
   def ethernet_layer_one(iface)
     return iface unless ethtool_binary = find_ethtool_binary
     keys = %w[ Speed Duplex Port Transceiver Auto-negotiation MDI-X ]
@@ -161,6 +162,7 @@ Ohai.plugin(:Network) do
     iface
   end
 
+  # determine link stats, vlans, queue length, and state for an interface using ip
   def link_statistics(iface, net_counters)
     so = shell_out("ip -d -s link")
     tmp_int = nil
@@ -251,7 +253,6 @@ Ohai.plugin(:Network) do
     end
   end
 
-
   def parse_ip_addr_link_line(cint, iface, line)
     if line =~ /link\/(\w+) ([\da-f\:]+) /
       iface[cint][:encapsulation] = linux_encaps_lookup($1)
@@ -293,7 +294,7 @@ Ohai.plugin(:Network) do
         iface[cint][:addresses][tmp_addr][:scope] = ($1.eql?("host") ? "Node" : $1.capitalize)
       end
 
-      # If we found we were an an alias interface, restore cint to its original value
+      # If we found we were an alias interface, restore cint to its original value
       cint = original_int unless original_int.nil?
     end
     cint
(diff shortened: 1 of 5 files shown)


lib/ohai/plugins/linux/network.rb
```diff
@@ -406,17 +407,19 @@ def parse_ip_addr_inet6_line(cint, iface, line)
...
+          if route && !route.empty?
+            macaddress iface[route[:dev]][:addresses].select{|k,v| v["family"]=="lladdr"}.first.first unless iface[route[:dev]][:flags].include? "NOARP"
             if family[:name] == "inet"
               ipaddress route[:src]
-              macaddress iface[route[:dev]][:addresses].select{|k,v| v["family"]=="lladdr"}.first.first unless iface[route[:dev]][:flags].include? "NOARP"
             else
               ip6address route[:src]
             end
+          else
+            macaddress iface[default_route[:dev]][:addresses].select{|k,v| v["family"]=="lladdr"}.first.first unless iface[default_route[:dev]][:flags].include? "NOARP"
```
jaymzh: in the event routes were found, this won't fill in mac addr, right?

lib/ohai/plugins/network.rb
```diff
@@ -159,7 +166,9 @@ def network_contains_address(address_to_match, ipaddress, iface)
           Ohai::Log.debug("unable to detect ip6address")
         else
           ip6address r["ip"]
-          if r["mac"] and macaddress.nil? and ipaddress.nil?
+          # don't overwrite macaddress set by "#{os}::network" plugin
+          # and also
+          if r["mac"] and macaddress.nil? and ( ipaddress.nil? || ipaddress == "127.0.0.1" )
```
jaymzh: we should flag this - we have a strong preference for only taking from v6 if v4 isn't around, but that's a bad assumption. People will soon be doing v4 on tunnel interfaces in v6 places... this is fine for now as it preserves the previous behavior, but it probably wont' be good forever.

tas50: I'm in agreement that this isn't the long term desired behavior. It's purely for backwards compatibility so we don't overwrite the v4 mac with a v6 on every run.  At the moment we do have the expectation through the specs that v4 takes precedence, but not setting the v6 mac because ipaddress was set from lo seems a bit silly.  

A flag would be useful here, but I wonder if we're running into a fundamental limitation of the single macaddress attribute that contains both the mac of the v6 interface and the v4 interface.

jaymzh: yeah - I suggesting you change this - just ... it's a thing we should revisit.

tas50: I created #690 to track the long term issue of what stack to prefer

lib/ohai/plugins/linux/network.rb
jaymzh: don't forget this.

jaymzh:
Ping?

tas50:
Sorry @jaymzh I've been out with the flu.  Still recoverying

lib/ohai/plugins/linux/network.rb
```diff
@@ -406,17 +407,19 @@ def parse_ip_addr_inet6_line(cint, iface, line)
             ]
           end.first
 
-          unless route.nil? or route.empty?
+
+          if route && !route.empty?
+            macaddress iface[route[:dev]][:addresses].select{|k,v| v["family"]=="lladdr"}.first.first unless iface[route[:dev]][:flags].include? "NOARP"
```
jaymzh: I think this is fine, but ftr, if you have v4 and v6 routes that match, v6 will win since it's done second.

tas50:
Closing this out since @btm picked up the slack from my deathly illness and finished this.

PR #695, opened 2015-12-23
Fix detection of mac address on IPv6 only systems

### LINKED ISSUE #685: network plugin is broken on v6 only hosts

btm:
Fixes #685.
Builds on and replaces #689.

@jaymzh @tas50

### DIFF
--- a/lib/ohai/plugins/linux/network.rb
+++ b/lib/ohai/plugins/linux/network.rb
@@ -313,6 +313,65 @@ Ohai.plugin(:Network) do
     interfaces[interface][:addresses].select{|k,v| v["family"]=="lladdr"}.first.first unless interfaces[interface][:flags].include? "NOARP"
   end
 
+  # returns the default route with the lowest metric (unspecified metric is 0)
+  def choose_default_route(routes)
+    default_route = routes.select do |r|
+      r[:destination] == "default"
+    end.sort do |x,y|
+      (x[:metric].nil? ? 0 : x[:metric].to_i) <=> (y[:metric].nil? ? 0 : y[:metric].to_i)
+    end.first
+  end
+
+  # ipv4/ipv6 routes are different enough that having a single algorithm to select the favored route for both creates unnecessary complexity
+  # this method attempts to deduce the route that is most important to the user, which is later used to deduce the favored values for {ip,mac,ip6}address
+  def favored_default_route(routes, iface, default_route, family)
+    routes.select do |r|
+      if family[:name] == "inet"
+        # selecting routes for ipv4
+        # using the source field when it's specified :
+        # 1) in the default route
+        # 2) in the route entry used to reach the default gateway
+        r[:src] and # it has a src field
+          iface[r[:dev]] and # the iface exists
+          iface[r[:dev]][:addresses].has_key? r[:src] and # the src ip is set on the node
+          iface[r[:dev]][:addresses][r[:src]][:scope].downcase != "link" and # this isn't a link level addresse
+          ( r[:destination] == "default" or
+            ( default_route[:via] and # the default route has a gateway
+              IPAddress(r[:destination]).include? IPAddress(default_route[:via]) # the route matches the gateway
+              )
+            )
+      elsif family[:name] == "inet6"
+        # selecting routes for ipv6
+        iface[r[:dev]] and # the iface exists
+          iface[r[:dev]][:state] == "up" and # the iface is up
+          ( r[:destination] == "default" or
+            ( default_route[:via] and # the default route has a gateway
+              IPAddress(r[:destination]).include? IPAddress(default_route[:via]) # the route matches the gateway
+              )
+            )
+      end
+    end.sort_by do |r|
+      # sorting the selected routes:
+      # - getting default routes first
+      # - then sort by metric
+      # - then by prefixlen
+      [
+       r[:destination] == "default" ? 0 : 1,
+       r[:metric].nil? ? 0 : r[:metric].to_i,
+       # for some reason IPAddress doesn't accept "::/0", it doesn't like prefix==0
+       # just a quick workaround: use 0 if IPAddress fails
+       begin
+         IPAddress( r[:destination] == "default" ? family[:default_route] : r[:destination] ).prefix
+       rescue
+         0
+       end
+      ]
+    end.first
+  end
+
+  # Both the network plugin and this plugin (linux/network) are run on linux. This plugin runs first.
+  # If the 'ip' binary is available, this plugin may set {ip,mac,ip6}address. The network plugin should not overwrite these.
+  # The older code section below that relies on the deprecated net-tools, e.g. netstat and ifconfig, provides less functionality.
   collect_data(:linux) do
     require 'ipaddr'
 
@@ -359,18 +418,15 @@ Ohai.plugin(:Network) do
 
         routes = parse_routes(family, iface)
 
-        # using a temporary var to hold the default route
-        # in case there are more than 1 default route, sort it by its metric
-        # and return the first one
-        # (metric value when unspecified is 0)
-        default_route = routes.select do |r|
-          r[:destination] == "default"
-        end.sort do |x,y|
-          (x[:metric].nil? ? 0 : x[:metric].to_i) <=> (y[:metric].nil? ? 0 : y[:metric].to_i)
-        end.first
+        default_route = choose_default_route(routes)
 
         if default_route.nil? or default_route.empty?
-          Ohai::Log.debug("Unable to determine default_#{family[:name]}_interface as no default routes found")
+          attribute_name == if family[:name] == "inet"
+                              "default_interface"
+                            else
+                              "default_#{family[:name]}_interface"
+                            end
+          Ohai::Log.debug("Unable to determine '#{attribute_name}' as no default routes were found for that interface family")
         else
           network["#{default_prefix}_interface"] = default_route[:dev]
           Ohai::Log.debug("#{default_prefix}_interface set to #{default_route[:dev]}")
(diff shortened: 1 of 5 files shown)


lib/ohai/plugins/linux/network.rb
```diff
@@ -406,17 +412,25 @@ def parse_ip_addr_inet6_line(cint, iface, line)
...
             else
               ip6address route[:src]
+              if macaddress
+                Ohai::Log.debug("Not setting macaddress from ipv6 interface #{route[:dev]} because macaddress is already set")
+              else
+                macaddress get_mac_for_interface(iface, route[:dev])
+              end
             end
+          else
+            Ohai::Log.debug("Unable to determine ideal default route, not setting ipaddress/ip6address/macaddress. Should have considered #{default_route[:dev]}?")
```
jaymzh: why not pull it from default_route?

btm: I'd rather not change the behavior of ipaddress/ip6address/macaddress without a specific use case that we're solving for. Everyone wants the address attributes to be exactly the IP/mac address that they want in their use case, which is really hard to do without breaking someone else's use case.

We weren't previously setting the address attributes in this case, and we don't appear to need to for the ipv6 only case here. On the surface, it seems fine, but a whole pile of tests are broken by the change and would need to be updated. If we make a presumption here about a use case that nobody has asked for, we might find ourselves making a breaking change for another use case down the road. So I don't see the value in making that change until it is needed.

jaymzh: I may be mis-understanding the code... but you end up doing the exact same thing in `plugins/network.rb` anyway, so why not do it here? Also, it's one thing to _change_ the behavior for an existing case, but I can't imagine it's *ever* OK to just ... NOT pick a macaddress. That seems just broken in any and all use-cases.

btm: Can you identify any real scenario where we would get to this code path? If not, we're solving a problem that doesn't exist. I don't think we can get here, but I don't feel certain enough about that to fail to make someone report that to me.

It is absolutely okay to not pick a macaddress in some network configurations.

You need to be able to leave macaddress set to nil if the interface you're picking ipaddress from has no MAC. For a contrived case, if you run on a linux system with no routes and no external interfaces, you'll get nothing from `linux/network` because we have no routes, and fall back to the `network` plugin where you will get `ipaddress => 127.0.0.1` and `macaddress => nil` because the localhost interface has no need for layer 2 addressing. You wouldn't hit the code path we're discussing though because you don't have any routes.

If you go way back to [https://tickets.opscode.com/browse/OHAI-362](OHAI-362) we were breaking people who were using tun interfaces because we assumed macaddress would never be nil. Part of the fix of OHAI-362 is visible in the conditional check for the NOARP flag that still exists:

`macaddress iface[route[:dev]][:addresses].select{|k,v| v["family"]=="lladdr"}.first.first unless iface[route[:dev]][:flags].include? "NOARP"`

A point-to-point link isn't going to have a mac address because there's no addressing required. There's only one other host on the link, so anything we put on the wire is destined for that host and it expects everything it reads on the wire came from us.

btm:
@chef/client-core put down that cookie and review please.

lib/ohai/plugins/linux/network.rb
jaymzh: What? **I** get to that code!! That's the whole *point* of the bug.

btm: Are you sure? Have you run this branch and you're not getting a macaddress?

jaymzh: Allow me to prove the point I laid out in the original bug:

```
# pwd
/opt/chef/embedded/apps/ohai/lib/ohai/plugins/linux
# diff -u network.rb.orig network.rb
--- network.rb.orig	2015-12-23 15:00:11.551300264 -0800
+++ network.rb	2015-12-23 14:59:39.864068269 -0800
@@ -332,6 +332,8 @@
             else
               ip6address route[:src]
             end
+          else
+            puts "PHILD: I am here"
           end
         end
       end
# ohai | grep PHILD
PHILD: I am here
# 
```

I assure I while I said my patch wasn't good enough for production, I didn't just throw unreachable code in....

jaymzh:
Just to be clear here...  the PR does in fact give me a mac address, but it's `plugin/network.rb` that's doing it, not `plugin/linux/network.rb` - which is still finding 0 routes. It's not clear to me if it should or not but certainly that code is *not* unreachable.

I'm very unclear in general why both of those files try to determine `macaddress` `ipaddress` and `ip6address` in the first place, so this _could_ be the right fix - but I'd like to clarify what's actually happening here since the comments seem to imply things different from what's actually happening.

btm:
I understand the bug you're experiencing in v6 route detection, you shouldn't be falling through. I'll replicate and then fix that, standby.

lib/ohai/plugins/linux/network.rb
jaymzh: > Are you sure? Have you run this branch and you're not getting a macaddress?

I get a mac address with this branch for sure... but I also hit that code. When I use your branch:

```
[2015-12-23T15:25:39-08:00] DEBUG: Unable to determine ideal default route, not setting ipaddress/ip6address/macaddress. Should have considered eth0?
```

jaymzh:
Standing by. Ish --- I'm on a plane, so connection isn't terribly reliable, especially ssh over ssh over vpn over ATG wifi. :)

btm:
@jaymzh I've got to go back and work on tests now, but please take a look at this if you're around today.

jaymzh:
So I don't have time to test today, but looking over the code it's far far cleaner... though the `linux/network.rb` plugin still doesn't ever set a mac addr for my case, which is still the same comment I had before.

jaymzh:
@btm following up here now that the holiday is past.

btm:
@jaymzh WB. I got distracted by the `ip6address favored_route[:src]` business I realized because I had been focusing on removing the src requirement in the favored_route selection.  Can you share the debugging output from this branch for `ohai -l debug macaddress`? The ipv6 favored route detection doesn't use the src attribute anymore, so I would expect some progress here.

jaymzh:
Ah - you're right. I saw the `src` stuff, but it's v6-only.

OK, initial testing looks like this is working as expected. I'll do some more testing and hopefully by the time you have the tests fixed by able to give full thumbs-up

jaymzh:
OK here's my findings:

1. `plugins/network.rb` now determines my `macaddress` correctly.
2. If I completely remove `plugins/network.rb` and only rely on the linux-specific plugin **fails** to determine my `ip6address` which is presumably because of still using `src`:
```
              ip6address favored_route[:src]
```

Which I'm pretty sure is not the desired behavior?

btm:
Not desired, but expected due to the use of src there. Do you need to get ip6address from linux/network instead of network?

I started writing an RFC6724 implementation for when src isn't available, but realized I didn't think I needed that for macaddress because we take the device from the default route. We would need it to pick an ip6address from a route without a src.

jaymzh:
Much the same as the previous round where we made `plugins/network.rb` get macaddress but not `plugins/linux/network.rb` it seems like a bug in the linux-specific plugin.

jaymzh:
So I'm entirely sure you need a full RFC6724 implementation. The non-linux-specific plugin is already managing to pick a sane default address.

On top of that, if I only have one (global scope) address, that should be pretty clear. You're already picking a "default interface", right?

btm:
The network plugin returns the first address it finds on `default_inet6_interface` that is on the same network as `default_inet6_gateway`. If that gets you the value you expect, I think we should continue letting the network plugin set the value of ip6address (when the linux/network plugin doesn't set it because of a better address found on the default route via src) for you.

If we wanted to do better than that, I'd write a RFC6724 implementation in the linux/network plugin to make sure that the address being chosen for ip6address is more likely to be the one used as the source address when accessing the default gateway.

Does allowing the network plugin to set ip6address for you not work?

jaymzh:
This totally works for me. I'm good with this (at least for now).

That said I don't understand your logic. That RFC has nothing to do with Linux - if it should be anywhere, it should be in the non-OS-specific plugin... it's also weird to me that there's TWO plugins which can potentially fill in the SAME data, that seems a bit broken to me...

jaymzh:
Ping?

btm:
I don't disagree with you on the ideal design.

The `linux/network` plugin is an outlier, it shouldn't be setting ipaddress, ip6address, or macaddress. That should only be done by the `network` plugin so that those choices are consistent across platforms. Ideally we would remove that behavior from `linux/network`, and add RFC6724 address selection to `network`.

That said, it's been like that for almost 4 years. There's significant risk in changing the algorithm for {ip,ip6,mac}address. So that refactor would start with rewriting the tests for `linux/network`... nearly from scratch. 

My scope right now is fixing the original bug: choosing the "correct" macaddress on an ipv6 only host. I think we've done that (plus fixing the tests). I don't want to creep scope to a rewrite of the plugins unless we need to, which would necessitate triggering a prioritization conversation.

I'll work on cleaning up the tests so we can get this merged.

jaymzh:
yup, I'm on board. Once tests pass, I look it all over again and thumbs up... but this looks.... way cleaner.

jaymzh:
Ping?

btm:
these tests are extremely difficult to grok. working on it. not confident
that "fixing" the tests isn't just acknowledging breaking behavior.

jaymzh:
is looks sane to me... though it definitely needs a lookover from someone else. Perhaps the Ohai queen? @mcquin

jaymzh:
Oh, and to be clear, I'm :+1: 
CC @chef/client-core

thommay:
this looks correct :+1:

jaymzh:
there's not enough +1's for a non-employee to merge... but you guys wanna do this? or do we want more reviewers?

lamont-granquist:
there's no distinction in the RFC between employee and non-employee, so with 2 +1's its always mergeable.

jaymzh:
@lamont-granquist did we change that? I thought it was at least half?

btm:
@jaymzh yeah, two :+1: from maintainers. We backed off the "majority" thing because it was untenable once we ended up with a dozen maintainers.

https://github.com/chef/chef-rfc/blob/master/rfc030-maintenance-policy.md#how-a-patch-gets-merged

Some day we'll be explicit about what repositories that applies to other than chef/chef.

jaymzh:
ah... I do remember that PR now... totally forgot! thanks for the reminder.

PR #699, opened 2015-12-28
Create kernel[:processor] from uname -p output

MarkGibbons:
An important distiction on solaris servers is x86 vs. sparc processor
architecture.  Using uname -m to report the machine type does make that
distinction explicitly.  The uname -p value is explicit about the
difference.

### DIFF
--- a/lib/ohai/plugins/kernel.rb
+++ b/lib/ohai/plugins/kernel.rb
@@ -29,7 +29,8 @@ Ohai.plugin(:Kernel) do
   def init_kernel
     kernel Mash.new
     [["uname -s", :name], ["uname -r", :release],
-     ["uname -v", :version], ["uname -m", :machine]].each do |cmd, property|
+     ["uname -v", :version], ["uname -m", :machine],
+     ["uname -p", :processor]].each do |cmd, property|
       so = shell_out(cmd)
       kernel[property] = so.stdout.split($/)[0]
     end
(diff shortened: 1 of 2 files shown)


mcquin:
LGTM

\cc @chef/client-core

cachamber:
I'm not a fan of using "processor" for this attribute. It's an instruction set architecture identifier - therefore I would expect it to be named something like "arch".

MarkGibbons:
@cachamber For Solaris it identifies the instruction architecture. For other operating systems it identifies the processor type.  The man page for Solaris uname says it is "the current host's ISA or processor type".  So I took the intersection of the descriptions and came up with "processor" as ISA would be mildly misleading for other operating systems. I don't really see a better name and the code applies to the default ohai data not just to Solaris.

lamont-granquist:
it looks like `uname -p` may produce "Intel(R) Core(TM)2 CPU T7200 @ 2.00GHz" on vanilla linux kernels, and 'uname -m` is more appropriate for 'arch'.  so 'processor' for this probably is fine.

e.g. https://www.mulesoft.org/jira/browse/MULE-1288

cachamber:
@MarkGibbons  - I don't disagree with using `uname -m` to determine the processor architecture. In fact, that's what it's there for.  My issue is that using "processor" as a bare word is too vague.  

As a Chef developer, I would "assume" that processor = processor model.  I wouldn't naturally think to use the "processor" value in a block where I'm attempting to determine the platform architecture.  I would, however, use "processor" to determine the model/family of a processor for optimizations - such as  - `Intel Xeon` versus `AMD Opteron`, or `sun4u` versus `sun4v`.

@lamont-granquist - Your example is exactly my point.  `uname -m` isn't "processor" it's architecture.

When it comes down to it, what you are attempting to determine is the processor architecture, ie the ISA (instruction set architecture).  I would expect to use a value such as "arch", "processor_arch" - something along those lines.

MarkGibbons:
@cachamber uname -m (sun4u, sun4v) is already reported as kernel[:machine].   I want uname -p output in order to separate x86-64 from sparc on Solaris servers.  The use of "machine" and "processor" is consistent with how uname documents the results of the -m and -p options on solaris and on other OS flavors. So, the unnaturalness of the attribute name doesn't seem like a good argument to me.  

If i was using the arch command then I might create kernel[:arch], but I'm not.

Thanks

cachamber:
LGTM - I'll acquiesce. The value is descriptive enough, and will work.  I'm still annoyed that it won't have the same value across linux implementations, but it's not something that can't be worked around.

lamont-granquist:
:+1:

lamont-granquist:
@mcquin :+1: ?

thommay:
Logic is sound to me. :+1:

mcquin:
yeah i'm +1

PR #700, opened 2016-01-07
Allow route table override

### LINKED ISSUE #696: Main routing table too large for default Ohai plugin
As part of the `linux/network.rb` plugin `check_routing_table` causes us some issue listing the main routing table on machines where it is far too large. It would be helpful to figure out an acceptable way to limit this to the default routing table as a configurable option.

The change I'm talking about would be as seen in https://gist.github.com/whiteley/cbc4e9ef217310410878 but it's undesirable to change the default behavior. The standard case could effectively be `ip route show table main` but where the named table parameter is configurable. If you could recommend an acceptable solution such as using an attribute to `Ohai::Config`, a hints file, etc. I would be happy to submit the change with tests.

This would prevent us from having to maintain a fork of omnibus-chef and locally building the chef-client packages.

Thanks!

whiteley:
* read default route table name from hints file
* default to `main` table to maintain current behavior

fixes #696

### DIFF
--- a/lib/ohai/plugins/linux/network.rb
+++ b/lib/ohai/plugins/linux/network.rb
@@ -76,8 +76,8 @@ Ohai.plugin(:Network) do
   #    the routing table source field.
   # 3) and since we're at it, let's populate some :routes attributes
   # (going to do that for both inet and inet6 addresses)
-  def check_routing_table(family, iface)
-    so = shell_out("ip -o -f #{family[:name]} route show")
+  def check_routing_table(family, iface, default_route_table)
+    so = shell_out("ip -o -f #{family[:name]} route show table #{default_route_table}")
     so.stdout.lines do |line|
       line.strip!
       Ohai::Log.debug("Parsing #{line}")
@@ -384,6 +384,14 @@ Ohai.plugin(:Network) do
     counters Mash.new unless counters
     counters[:network] = Mash.new unless counters[:network]
 
+    # ohai.plugin[:network][:default_route_table] = 'default'
+    if configuration(:default_route_table).nil? || configuration(:default_route_table).empty?
+      default_route_table = 'main'
+    else
+      default_route_table = configuration(:default_route_table)
+    end
+    Ohai::Log.debug("default route table is '#{default_route_table}'")
+
     # Match the lead line for an interface from iproute2
     # 3: eth0.11@eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP
     # The '@eth0:' portion doesn't exist on primary interfaces and thus is optional in the regex
@@ -415,7 +423,7 @@ Ohai.plugin(:Network) do
 
         iface = extract_neighbors(family, iface, neigh_attr)
 
-        iface = check_routing_table(family, iface)
+        iface = check_routing_table(family, iface, default_route_table)
 
         routes = parse_routes(family, iface)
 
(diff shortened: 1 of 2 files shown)


lib/ohai/plugins/linux/network.rb
```diff
@@ -318,6 +318,10 @@ def parse_ip_addr_inet6_line(cint, iface, line)
     counters Mash.new unless counters
     counters[:network] = Mash.new unless counters[:network]
 
+    network_hint = hint?('network') || {}
+    default_route_table = network_hint.fetch('default_route_table', 'main')
+    Ohai::Log.debug("default route table is '#{default_route_table}'")
```
mcquin: There is a way for users to specify configurable values for plugins in their configuration files, which eliminates the need for hints files in cases like this. Here's an example of a plugin which uses configuration: https://github.com/chef/ohai/blob/081589066eefa00a60192f8cac8801e3b2b31bfe/lib/ohai/plugins/solaris2/filesystem.rb#L66-L74

whiteley: I created both #696 and Zendesk ticket 7712 ahead of time to avoid this type of revision but I've changed it as you describe.

lib/ohai/plugins/linux/network.rb
```diff
@@ -384,6 +384,14 @@ def favored_default_route(routes, iface, default_route, family)
     counters Mash.new unless counters
     counters[:network] = Mash.new unless counters[:network]
 
+    # ohai.plugin[:network][:default_route_table] = 'default'
+    if configuration(:default_route_table).nil? || configuration(:default_route_table).empty?
+      default_route_table = 'main'
```
mcquin: Does populating attribute values from `route show` produce the same results as `route show table main`? If not, the default behavior of this plugin should probably be to shell out with `route show` by default.

mcquin:
Looking for review from @chef/client-core

lib/ohai/plugins/linux/network.rb
whiteley: Implementation:
```
[rime][~]% diff -q <(ip route show) <(ip route show table main); echo $?
0
```

Documentation:
```
       ip route show
              list routes
              the command displays the contents of the routing tables or the route(s) selected by some criteria.

...

              table TABLEID
                     show the routes from this table(s). The default setting is to show table main.  TABLEID may either be the ID of a real table or one of the special values:

                             all - list all of the tables.

                             cache - dump the routing cache.
```

mcquin: Great! In that case, it seems reasonable to make the "new" default be `table main` :)

mcquin:
:+1:

jaymzh:
We ran into this too, and have a similar hack in our local systems... however, it's only temporary. We (Facebook) revamped the ipv6 routing system in the kernel and that code landed upstream in 4.1.

The problem is that the older naive implementation used to stick an entry in the routing table for every single host you spoke to as a way to cache the MTU, even if it was the standard MTU. On top of that the FIB code was super inefficient. In 4.1 we stop caching entries in the route table that don't have different MTUs, and we fixed the FIB code. Moving to 4.1 or later kernel may just fix this for you and still provide you full access to your routing table.

jaymzh:
(I can go find and point you to the patches specifically, if that helps)

lamont-granquist:
:+1:

thommay:
:+1:

whiteley:
@jaymzh I appreciate the info on that, it will definitely help us out in the near future.

PR #705, opened 2016-01-13
Update e-mails and copyrights

tas50:
Opscode -> Chef Sofware Inc. and bring all copyrights to current. End
the needed to keep editing these on every commit.  Lets just get it over
with.


thommay:
:+1:

thommay:
Just gonna merge this so it doesn't rot.

PR #711, opened 2016-02-01
Improve EC2 detection when hint isn't present

tas50:
This allows us to detect Linux hosts based on the DMI data as it includes amazon specific strings.  The present method of looking for an arp address fails when hosts are in a VPC, which they all are by default now.

### DIFF
--- a/lib/ohai/plugins/ec2.rb
+++ b/lib/ohai/plugins/ec2.rb
@@ -43,8 +43,10 @@ Ohai.plugin(:EC2) do
   end
 
   # look for amazon string in dmi bios data
+  # this only works on hvm instances as paravirt instances have no dmi data
   def has_ec2_dmi?
     begin
+      # detect a version of '4.2.amazon'
       if dmi[:bios][:all_records][0][:Version] =~ /amazon/
         Ohai::Log.debug("has_ec2_dmi? == true")
         true
--- a/lib/ohai/plugins/ec2.rb
+++ b/lib/ohai/plugins/ec2.rb
@@ -60,7 +60,7 @@ Ohai.plugin(:EC2) do
       end
       ec2[:userdata] = self.fetch_userdata
       #ASCII-8BIT is equivalent to BINARY in this case
-      if ec2[:userdata].encoding.to_s == "ASCII-8BIT"
+      if ec2[:userdata] && ec2[:userdata].encoding.to_s == "ASCII-8BIT"
         Ohai::Log.debug("Binary UserData Found. Storing in base64")
         ec2[:userdata] = Base64.encode64(ec2[:userdata])
       end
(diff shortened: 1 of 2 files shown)


mfischer-zd:
Do HVM and paravirtualized EC2 instances work identically in this regard?

tas50:
Let me spin up a paravirt box to make sure.

thommay:
As a tweak, can you add the line that you're finding in DMI as a comment? Otherwise :+1: assuming you test both HVM + paravirt inside and outside a VPC.

tas50:
So it looks like this won't work on the paravirts since they contain no BIOS so no DMI data.  I'm open to improving detection there, but it seems to be less important as new generation instance types only work on HVM AMIs.

tas50:

danielsdeleo:
FINALLY. :+1:

lamont-granquist:
:+1:

tas50:
I'll merge this whenever appveyer finishes.  I'm working on another PR to detect the paravirts, but it involves parsing dmesg, which is pretty dirty.

lamont-granquist:
I'm kinda :-1: on parsing dmesg since its a ring buffer...  that information will wind up getting lost and the detection will fail, and if the user doesn't have manual hints setup then suddenly ohai will stop reporting the correct ec2 information, which will lead to bug reports...

we could do something to read dmesg to setup a hint file on the box during bootstrapping...

tas50:
If we're bootstrapping they'll get the ec2 hint anyways.  Let me see if there's amazon fingerprints anywhere else.  Dmesg has a nice string that identifies the Xen hypervisor has being an Amazon version.  That might be in other places.

btm:
detecting ec2 from dmesg and then creating the hint file sounds like a super clever solution for those who aren't using `knife ec2 server create`. Of course, I wonder if there's a case where you'd want to opt out of that.

hints exists entirely because there's rarely obvious fingerprints about what cloud platform you're on. when there is, you buy those developers :beers:.

mfischer-zd:
I don't see any reference to dmesg in this PR.  Is Ohai not getting DMI information using dmidecode(8)?

lamont-granquist:
See above:  "I'm working on another PR to detect the paravirts, but it involves parsing dmesg, which is pretty dirty."

tas50:
/usr/bin/ec2metadata seems to actually be the best thing on both hvm and paravirts.  So here's the thought for the next PR

Hint -> DMI -> ec2metadata_bin -> MAC

Hint when someone bootstraps via knife-ec2
DMI if they're on a HVM box
ec2metadata if they're on an official amazon AMI on either HVM or Paravirt
MAC if they're in a classic networking setup

btm:
https://github.com/chef/ohai/pull/723

PR #714, opened 2016-02-02
Detect Azure using the Azure agent and DHCP options

tas50:
Azure has no metadata endpoint and we pass in everything we stick in the mash via the hint, but there's still value in detecting azure when we didn't knife bootstrap. Users still want to do different things on their cloud hosts and this gives them that ability.

1) Detect hosts as being on azure when the waagent is present (windows and linux)
2) Detect hosts as being on azure when we find unknown-245 dhcp option which is used by Azure (linux only)
3) Fix the specs for the lack of hint file or presence of rackspace hint file so they actually run.  This confused me for a while until I realized they were expecting nil and not running the plugin so they got nil. Ouch.

And a bonus commit: switch the order or MS virtualization detection since 99.9% of the time it's hyper-v. No point in checking for VirtualPC first only to skip it 99.9% of the time.

### DIFF
--- a/lib/ohai/mixin/dmi_decode.rb
+++ b/lib/ohai/mixin/dmi_decode.rb
@@ -22,10 +22,10 @@ module ::Ohai::Mixin::DmiDecode
       case line
       when /Manufacturer: Microsoft/
         if dmi_data =~ /Product Name: Virtual Machine/
-          if dmi_data =~ /Version: (VS2005R2|6.0)/
-            return 'virtualpc'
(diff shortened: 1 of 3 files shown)


tas50:

thommay:
LGTM :+1:

smurawski:
:+1: until we get a metadata endpoint in azure.  There may be conflicts for folks migrating between on-prem and azure, but that's a pretty minimal case.

PR #716, opened 2016-02-02
Cleanup to the lang plugins and additional debug information

tas50:
Avoid creating the mashes before we've successfully shelled out.  Most of the time we'll fail and then we wasted our time initializing the mash.

Be smarter about when we save the data we found.  Some of these plugins collect 1/2 dozen attributes, but only use them if a 'version' attribute was found. Seems to be a lot of cargo culting since this is almost all the same code. We should save if we get back anything.

Finally add debugging information. Debug what we successfully shelled out with, which is helpful when you don't want to dig into the code to figure out what we're doing to get information. Also rescue the failure exceptions on the shell out and print simple messages.  This avoids pages of output.

### DIFF
--- a/lib/ohai/plugins/c.rb
+++ b/lib/ohai/plugins/c.rb
@@ -6,9 +6,9 @@
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
-# 
+#
 #     http://www.apache.org/licenses/LICENSE-2.0
-# 
+#
 # Unless required by applicable law or agreed to in writing, software
 # distributed under the License is distributed on an "AS IS" BASIS,
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -30,6 +30,7 @@ Ohai.plugin(:C) do
     begin
       so = shell_out("gcc -v")
       if so.exitstatus == 0
+        Ohai::Log.debug("Successfully ran gcc -v")
         description = so.stderr.split($/).last
         output = description.split
         if output.length >= 3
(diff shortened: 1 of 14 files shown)


tas50:

lib/ohai/plugins/c.rb
```diff
@@ -39,6 +40,7 @@
         end
       end
     rescue Errno::ENOENT
```
lamont-granquist: TIL:  shell_out will throw an exception if the command does not exist...

tas50: I'm going to clean up the other places we have this issue.  Debug output for Ohai is hard to read right now

lamont-granquist:
:+1:

lib/ohai/plugins/c.rb
mcquin: Not always... :(

```
[5] pry(main)> so = Mixlib::ShellOut.new('cl /?').run_command
=> <Mixlib::ShellOut#70288827374520: command: 'cl /?' process_status: #<Process::Status: pid 65429 exit 127> stdout: '' stderr: 'sh: cl: command not found' child_pid: 65429 environment: {} timeout: 600 user:  group:  working_dir:  >
[6] pry(main)> so.exitstatus
=> 127
```

I'm really confused about this behavior in shell_out (raising errors vs. non-zero exit codes when a command isn't found).

tas50: hmm that is odd

mcquin: Example from a command a little later on in this PR:

```
[2] pry(main)> so = Mixlib::ShellOut.new('elixir -v').run_command
Errno::ENOENT: No such file or directory - elixir
from /opt/chefdk/embedded/lib/ruby/gems/2.1.0/gems/mixlib-shellout-2.2.5/lib/mixlib/shellout/unix.rb:338:in `exec'
```

The only difference is that I have C installed through XCode but Elixir has never been installed on my system?

mcquin:
Would it be useful to have a wrapper method for `shell_out` for this pattern? It's going to happen A LOT.

tas50:
#728 is the issue to track creating a helper that would rescue the exceptions and handle success / failure debug logging.

mcquin:
OK, great! :)

tas50:
I'm going to close this out and go the helper route

PR #720, opened 2016-02-02
Make pnp plugin opt-in via config

danielsdeleo:
Makes the windows `Drivers` plugin opt-in via a config setting, similar to the packages plugin (see: https://github.com/chef/ohai/pull/717 ). This is technically a breaking change, but we decided in this case that the plugin was delivering too much negative value so it's worth it. The plugin can be re-enabled by setting `ohai.plugins[:drivers][:enabled] = true` in the Chef config file.

### DIFF
--- a/lib/ohai/plugins/windows/drivers.rb
+++ b/lib/ohai/plugins/windows/drivers.rb
@@ -20,12 +20,13 @@ Ohai.plugin(:Drivers) do
   depends "kernel"
 
   collect_data(:windows) do
-    require 'wmi-lite/wmi'
-
-    kext = Mash.new
-    pnp_drivers = Mash.new
-
     if configuration(:enabled)
+
+      require 'wmi-lite/wmi'
+
+      kext = Mash.new
+      pnp_drivers = Mash.new
+
       wmi = WmiLite::Wmi.new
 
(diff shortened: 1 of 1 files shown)


chefsalim:
:+1:

tas50:
Pretty sure most windows users wanted this 3 years ago

jaym:
should the creation of the mashes be moved into the `if enabled`? If not, I can change the packages plugin to behave the same way

jaym:
The creation of the `wmi` object also does not need to happen unless enabled

mwrock:
:+1:

danielsdeleo:
> should the creation of the mashes be moved into the if enabled? If not, I can change the packages plugin to behave the same way

@jaym I thought that was how you did it, but maybe I was misremembering or you changed it or something? In any case I guess it's kind of a toss-up. In one case `node["attribute-from-disabled-plugin"].each` will blow up with `NoMethodError on nil`, in the other, the same code will iterate over zero items.

jaym:
That's the first way I did it, then I decided to change it so that it was like the plugin never ran.

danielsdeleo:
I'll change to make it consistent with what you did.

jaym:
:+1:

PR #721, opened 2016-02-02
Update changelog, bump version

chefsalim:
Prep for release of Ohai 8.9.1

@chef/client-core @tas50 @danielsdeleo @jaym

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Ohai Changelog
 
+## Release 8.10.0
+* [pr#720](https://github.com/chef/ohai/pull/720) Make pnp plugin opt-in via config
+* [pr#717](https://github.com/chef/ohai/pull/717) Don't enable packages plugin by default
+* [pr#711](https://github.com/chef/ohai/pull/711) Improve EC2 detection when hint isn't present
+
 ## Release 8.9.0
 * [**phreakocious**](https://github.com/phreakocious):
   - Collect layer 1 Ethernet information per NIC on Linux hosts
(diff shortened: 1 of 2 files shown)


danielsdeleo:
@chefsalim mine's not merged yet but should be soon

jaym:
Let's hold off until @danielsdeleo merges

chefsalim:
Yeah I'll rebase, this is just prep :)

lamont-granquist:
8.10.0 instead of 8.9.1?  its not a big release, but they're not strictly bugfixes...  dunno...

chefsalim:
Seems like they're kind of fixes, as we are cleaning up after the last release..?

tas50:
The ec2 thing is questionably a fix.  We did make an attempt to detect ec2 before and now we do a better job of it. It was *sorta* broken before.

danielsdeleo:
The pnp thing technically should be a major bump but we're making a semver exception because of large negative value there. If we're including it, I'd prefer to do a minor bump (or we could even do a major bump for ohai but not for chef-client).

lamont-granquist:
major bumps to ohai are pretty annoying to do because of pessimistic version constraints everywhere that'll need to get bumped...

danielsdeleo:
Yeah, I'm fine with minor if we'd rather.

btm:
:+1: on bumping minor instead of patch. it's free.

jaym:
I vote for 8.10.0 as well

jaym:
:+1:

danielsdeleo:
:+1: Make sure we're very clear about the PnP thing when we ship

PR #722, opened 2016-02-03
Detect paravirt amazon instances without hint files

tas50:
Official AMIs ship with the ec2metadata binary which we can look for. With this change we should be detecting the majority of scenarios without a hint file.

has_ec2_mac has been renamed has_xen_mac since we're actually looking for a xen mac.  Due to that the attempt to connect to the metadata only needs to happen when we only find the mac address since that can be either AWS or Xen.  If we find dmi data, a hint, or the metadata binary we can safely assume the metadata endpoint will respond.

 I also consolidated the "no hints or signs of AWS" scenario into a single spec since it seems to make more sense that way.

### DIFF
--- a/lib/ohai/plugins/ec2.rb
+++ b/lib/ohai/plugins/ec2.rb
@@ -28,17 +28,28 @@ Ohai.plugin(:EC2) do
   depends "network/interfaces"
   depends "dmi"
 
+  # look for ec2metadata which is included on paravirt / hvm AMIs
+  def has_ec2metadata_bin?
+    if File.exist?('/usr/bin/ec2metadata')
+      Ohai::Log.debug("ec2 plugin: has_ec2metadata_bin? == true")
+      true
+    else
+      Ohai::Log.debug("ec2 plugin: has_ec2metadata_bin? == false")
+      false
+    end
+  end
+
   # look for arp address that non-VPC hosts will have
-  def has_ec2_mac?
+  def has_xen_mac?
     network[:interfaces].values.each do |iface|
       unless iface[:arp].nil?
         if iface[:arp].value?("fe:ff:ff:ff:ff:ff")
-          Ohai::Log.debug("has_ec2_mac? == true")
+          Ohai::Log.debug("ec2 plugin: has_xen_mac? == true")
           return true
         end
       end
     end
-    Ohai::Log.debug("has_ec2_mac? == false")
+    Ohai::Log.debug("ec2 plugin: has_xen_mac? == false")
     false
   end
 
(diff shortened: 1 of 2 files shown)


tas50:

thommay:
merge in #723 and then I'm :+1:

tas50:
Merged in and tweaked the changes that @btm made to warn on the MAC detection logic

btm:
:+1: after appveyor is green

appveyor failure looks like a legit test failure:
```
 1) Ohai::System plugin ec2 without dmi or ec2metadata binary, with xen mac, and metadata address connected warns that the arp table method is deprecated
     Failure/Error: expect(Ohai::Log).to receive(:warn).with(/will be removed/)
 
       (Ohai::Log (class)).warn(/will be removed/)
           expected: 1 time with arguments: (/will be removed/)
           received: 0 times
     # ./spec/unit/plugins/ec2_spec.rb:267:in `block (3 levels) in <top (required)>'
 
Finished in 14.59 seconds (files took 1.67 seconds to load)
1467 examples, 1 failure, 1 pending
 
```

looks like travis passed but shows failed due to some kind of caching issue....

tas50:
@btm any idea on the expects for the warn not triggering.  It's clearly showing up in the output when you run the specs, but not triggering.  I shuffled things around a bit, but no luck.

btm:
@tas50 can you make sure this branch is still pushed?

btm:
@tas50 n/m, I see your repositories are named silly on github.

tas50:
@btm any magical ideas about that travis failure?

spec/unit/plugins/ec2_spec.rb
```diff
@@ -251,42 +253,39 @@
...
 
-  describe "without ec2 mac and metadata address connected" do
-    it_should_behave_like "!ec2"
+    it_should_behave_like "ec2"
 
-    before(:each) do
-      @plugin[:network][:interfaces][:eth0][:arp] = {"169.254.1.0"=>"00:50:56:c0:00:08"}
+    it "warns that the arp table method is deprecated" do
+      expect(Ohai::Log).to receive(:warn).with(/will be removed/)
+      @plugin.has_ec2_mac?
```
mcquin: Do you mean `has_xen_mac?` or `looks_like_ec2?`

lib/ohai/plugins/ec2.rb
```diff
@@ -28,17 +28,34 @@
...
       unless iface[:arp].nil?
         if iface[:arp].value?("fe:ff:ff:ff:ff:ff")
-          Ohai::Log.debug("has_ec2_mac? == true")
+          # using MAC addresses from ARP is unreliable because they could time-out from the table
+          # fe:ff:ff:ff:ff:ff is actually a sign of Xen, not specifically EC2
+          deprecation_message <<-EOM
+ec2 plugin: Detected EC2 by the presence of fe:ff:ff:ff:ff:ff in the ARP table. This method is unreliable and will be removed in a future version of ohai. Bootstrap using knife-ec2 or create "/etc/chef/ohai/hints/ec2.json" instead.
+EOM
+          Ohai::Log.warn(deprecation_message)
+          Ohai::Log.debug("ec2 plugin: has_ec2_mac? == true")
```
mcquin: Do you mean `has_xen_mac? == true`?

spec/unit/plugins/ec2_spec.rb
tas50: That was it.  Thanks @mcquin

lib/ohai/plugins/ec2.rb
```diff
@@ -48,25 +65,26 @@ def has_ec2_dmi?
...
 
-
   def looks_like_ec2?
-    # Try non-blocking connect so we don't "block" if
+    return true if hint?('ec2')
+
+    # if has ec2 mac try non-blocking connect so we don't "block" if
     # the Xen environment is *not* EC2
-    hint?('ec2') || ( has_ec2_dmi? || has_ec2_mac?) && can_metadata_connect?(Ohai::Mixin::Ec2Metadata::EC2_METADATA_ADDR,80)
+    return true if (has_ec2metadata_bin? || has_ec2_dmi?) || (has_xen_mac? && can_metadata_connect?(Ohai::Mixin::Ec2Metadata::EC2_METADATA_ADDR,80))
```
mcquin: The original logic and the new logic here look different to me (but I'm having trouble parsing it). It looks like the original logic would check `can_metadata_connect?` unless a hint was provided. The new logic only checks `can_metadata_connect?` if `has_xen_mac?` is true and both `has_ec2metadata_bin?` and `has_ec2_dmi?` are false. The variation I'm seeing here is that the new logic doesn't check `can_metadata_connect?` when `has_ec2_dmi?` is true.

tas50: The original (pre-8.10) behavior was hint or xen mac + try to connect.  I incorrectly changed it in 8.10 so that it would try to connect if it found the DMI data.  The connection attempt was only in place because the MAC detection triggers on any Xen based system and is very unreliable.  If we find the binary or the DMI data we know 100% that we're on Amazon so there's no need to try to open the socket before we later try to open the socket again.

PR #723, opened 2016-02-03
deprecate looking for EC2 by mac

btm:
we should have enough other methods now, hints have been around for a while.
this detects xen, not ec2.

also cleans up some confusing whitespacing in the test.

we don't really need to do this, but my :neckbeard: is itching.


tas50:
I renamed the method to detect "ec2 mac" in https://github.com/chef/ohai/pull/722 since it's not an ec2 mac at all.  We probably should get this marked for deprecation since going forward it's not going to catch EC2 at all, but it will catch Xen.  We're just going to end up timing out on the ec2 metadata connection attempt over and over again on Xen guests.

thommay:
@tas50 can you merge these changes into #722 and we'll do both at once, please.

tas50:
Closing in favor of #722

PR #724, opened 2016-02-03
Improved debug logging for cloud plugins

tas50:
Add debug for hints / mac detection to a few plugins
Fixed the eucalyptus plugin to not spam debug with mac detection false warnings for each interface


thommay:
:+1:

PR #730, opened 2016-02-11
Deprecate run_command and popen4 in the command mixin

tas50:
These aren't used anywhere and seem to be pre-mixlib shellout. They should go away, but for now someone might be using them in a plugin so we'll just throw a warn and then remove them in Ohai 9.

### DIFF
--- a/lib/ohai/mixin/command.rb
+++ b/lib/ohai/mixin/command.rb
@@ -37,6 +37,7 @@ module Ohai
       module_function :shell_out
 
       def run_command(args = {})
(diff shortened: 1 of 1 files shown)


tas50:
@btm you probably have some insight on the history of these methods

btm:
yeah run_command -> popen4 was our original shell method before mixlib-shellout. It turns out shelling is super hard and there are lots of race conditions to handle. :+1:

tas50:

thommay:
LGTM! :+1:

PR #731, opened 2016-02-11
Log sigar gem load failures

tas50:
These two plugins run on every system and spam debug logs. There are other plugins that use sigar, but they're confined to HPUX where you need sigar for ohai to operate.


tas50:

jaym:
:+1:

thommay:
:+1:

PR #732, opened 2016-02-11
WIP: Improve shellout logging and add timeout

tas50:
Log successful and failed commands. This gives you the necessary context to dig into issues

### DIFF
--- a/lib/ohai/mixin/command.rb
+++ b/lib/ohai/mixin/command.rb
@@ -29,7 +29,7 @@ module Ohai
   module Mixin
     module Command
       def shell_out(cmd)
-        m = Mixlib::ShellOut.new(cmd)
+        m = Mixlib::ShellOut.new(cmd, :timeout => 30)
         begin
           m.run_command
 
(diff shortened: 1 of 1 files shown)


lib/ohai/mixin/command.rb
```diff
@@ -28,10 +28,29 @@
...
   module Mixin
     module Command
-      def shell_out(cmd)
-        m = Mixlib::ShellOut.new(cmd)
-        m.run_command
-        m
+      # accept a command and any of the mixlib-shellout options
+      def shell_out(cmd, **options)
+        # unless specified by the caller timeout after 30 seconds
+        options[:timeout] = 30 unless options[:timeout]
```
lamont-granquist: `options[:timeout] ||= 30`

lamont-granquist:
tiny niggle, LGTM

lib/ohai/mixin/command.rb
tas50: That would make a lot more sense

lib/ohai/mixin/command.rb
```diff
@@ -28,10 +28,29 @@
...
+        m = Mixlib::ShellOut.new(cmd, options)
+        begin
+          m.run_command
+
+          # we should really fail here on non-0, but historically we haven't so we can't now
```
mcquin: I believe there is one plugin that collects on non-zero. :)

mcquin:
Generally, I like to caution against using one or two letter variable names in Ruby. You can get yourself into weird shadowing bugs.

lib/ohai/mixin/command.rb
tas50: Good point.  I'll fix that one.

tas50:
Good point. I'll fix that one.

mcquin:
Closing in favor of https://github.com/chef/ohai/pull/788

PR #734, opened 2016-02-12
chefstyle catchup for ohai

lamont-granquist:
there's still a few cops that are unaddressed by this change, so it
won't be clean after merging this.


lamont-granquist:
since this is all chefstyle autofixing and we've already discussed all these rules on chef/chef already imma just merge this...

tas50:
If it's possible to wrap this all up in just 1 more PR that'd be really great.  The bulk changes really mess with the outstanding PRs and branches. It's a good thing that we're doing it, but lets not drag it out.

lamont-granquist:
Yep, this was the big bulk one.

Issue #382, opened 2014-08-05
GCE Plugin infinite loop


brauliobo:
How to disable it with knife solo?
https://github.com/opscode/chef/issues/1756

```
/usr/lib/ruby/gems/1.9.1/gems/ohai-7.2.0/lib/ohai/mixin/gce_metadata.rb:54:in `http_client'
/usr/lib/ruby/gems/1.9.1/gems/ohai-7.2.0/lib/ohai/mixin/gce_metadata.rb:59:in `fetch_metadata'
/usr/lib/ruby/gems/1.9.1/gems/ohai-7.2.0/lib/ohai/mixin/gce_metadata.rb:69:in `block in fetch_metadata'
/usr/lib/ruby/gems/1.9.1/gems/ohai-7.2.0/lib/ohai/mixin/gce_metadata.rb:68:in `each'
/usr/lib/ruby/gems/1.9.1/gems/ohai-7.2.0/lib/ohai/mixin/gce_metadata.rb:68:in `fetch_metadata'
/usr/lib/ruby/gems/1.9.1/gems/ohai-7.2.0/lib/ohai/mixin/gce_metadata.rb:69:in `block in fetch_metadata'
/usr/lib/ruby/gems/1.9.1/gems/ohai-7.2.0/lib/ohai/mixin/gce_metadata.rb:68:in `each'
/usr/lib/ruby/gems/1.9.1/gems/ohai-7.2.0/lib/ohai/mixin/gce_metadata.rb:68:in `fetch_metadata'
/usr/lib/ruby/gems/1.9.1/gems/ohai-7.2.0/lib/ohai/mixin/gce_metadata.rb:69:in `block in fetch_metadata'
/usr/lib/ruby/gems/1.9.1/gems/ohai-7.2.0/lib/ohai/mixin/gce_metadata.rb:68:in `each'
/usr/lib/ruby/gems/1.9.1/gems/ohai-7.2.0/lib/ohai/mixin/gce_metadata.rb:68:in `fetch_metadata'
/usr/lib/ruby/gems/1.9.1/gems/ohai-7.2.0/lib/ohai/mixin/gce_metadata.rb:69:in `block in fetch_metadata'
/usr/lib/ruby/gems/1.9.1/gems/ohai-7.2.0/lib/ohai/mixin/gce_metadata.rb:68:in `each'
[...]

PR #736, opened 2016-02-16
Update GCE metadata API version and fail better

tas50:
Use the non-beta metadata URL. This requires setting the header on all requests so there's a bit of changes to how the http connection is initialized
Fail better if the google internal DNS name can't be resolved


thommay:
Looks legit but you'll need to update your tests.

tas50:
Yep

thommay:
Fixes #382

tas50:

thommay:
:+1: @chef/client-core

lamont-granquist:
:+1:

PR #739, opened 2016-02-18
Lcg/chefstyle2

lamont-granquist:


lamont-granquist:
@chef/client-core this makes ohai clean for chefstyle 0.3.0 that I want to release today (current chefstyle master, which was last updated a few days ago).

can i just say the networking plugin and its specs are terribad?

spec/unit/plugins/windows/cpu_spec.rb
```diff
@@ -66,7 +66,6 @@
     @processors = [{ "description" => "Intel64 Family 6 Model 70 Stepping 1",
                      "deviceid" => "CPU0",
                      "family" => 2,
-                     "l2cachesize" => 0,
```
tas50: ?

lamont-granquist:
And the fixes in 1a882ccad96fdfc5de234c9c72c35bef1a6baba1 is all stuff that has crept in the past few days since the previous chefstyle PR to ohai.  If we merge this and then I release chefstyle that'll start enforcing stuff like double-quotes on this repo.  Right now we're not getting that, so master is accumulating more entropy over time...

spec/unit/plugins/windows/cpu_spec.rb
lamont-granquist: look at the end of that hash

lamont-granquist: Lint/DuplicatedKey is the cop that gets raised =)

lib/ohai/plugins/aix/virtualization.rb
```diff
@@ -82,8 +82,8 @@
               value = {
                         "mountpoint" => splat[0],
                         "device"     => splat[1],
-                        "vfs"     => splat[2],
-                        "options"     => splat[3].split(","),
+                        "vfs" => splat[2],
+                        "options" => splat[3].split(","),
```
danielsdeleo: do we want to tweak this to line up nicely with the values above?

thommay:
:+1:

danielsdeleo:
:+1: would be nice to replace the multi-line conditionals with predicate helper methods but not a big deal.

lib/ohai/plugins/aix/virtualization.rb
lamont-granquist: i just kinda want to get it in for now.

i suspect there's another cop to turn on that will autofix all of this for us...

PR #743, opened 2016-02-21
Add support for opensuse leap

spuranam:
This fix install failures of latest version chef server, chef reporting, chef manage and likely other products.


spuranam:
fix the build

PR #745, opened 2016-02-21
Add support for opensuse leap

spuranam:
This fix address install failures of latest version chef server, chef reporting, chef manage and likely other products.

### DIFF
--- a/lib/ohai/plugins/linux/platform.rb
+++ b/lib/ohai/plugins/linux/platform.rb
@@ -125,7 +125,7 @@ Ohai.plugin(:Platform) do
       suse_version = suse_release[/VERSION = ([\d\.]{2,})/, 1] if suse_version == ""
       platform_version suse_version
       if suse_release =~ /^openSUSE/
-        platform "opensuse"
+        platform "opensuse" && platform_version.to_i < 42
       else
         platform "suse"
       end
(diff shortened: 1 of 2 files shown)


tas50:
Is Leap enough like Enterprise that identifying leap as enterprise would satisfy everyone?

spuranam:
[OpenSuse Leap tacks SLES](https://news.opensuse.org/2015/11/04/opensuse-leap-42-1-becomes-first-hybrid-distribution/), just like Centos tracks RHEL hence it should be safe to make that assumption.

tas50:
:+1:

lamont-granquist:
so i'm :-1: on returning `suse` for node[:platform] because it is NOT SLES.  it is a distribution built on top of SLES.  the platform probably needs to be 'opensuse' or 'leap' or or 'suseleap' or something else that makes sense there that isn't 'suse' (which means SLES in our world).  this isn't a CentOS-RHEL because CentOS is pretty much /identically/ RHEL (there was that period when they shipped yum and redhat didn't but that stopped back in RHEL4 days).  and anyway we still have a platform of 'centos' for it.

tas50:
@lamont-granquist has shown me the light here.  We probably want to make a platform of suseleap that maps to the suse platform_family since this is similar but different than suse.

lamont-granquist:
definitely :+1: on `suse` for platform_family.

tas50:
Closing this in favor of #784

PR #747, opened 2016-02-26
Use escape_glob_dir instead of escape_glob

jaym:
For ruby >= 2.2, Dir.glob behaves differently than ruby 2.0
and breaks in more places when backslashes are used when globbing.
This means that Windows is sad and will not load any plugins with
Ruby >= 2.2 in a lot of cases. This is the safest thing to do
because it will use forward slashes.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,6 +7,7 @@ platform:
 environment:
   matrix:
     - ruby_version: "200"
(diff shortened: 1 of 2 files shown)


tas50:
:+1:

thommay:
:+1:

PR #748, opened 2016-02-29
Fix false detection of Linux hosts with a DHCP lease as being azure

tas50:
The presence of a DHCP lease file was enough to match as Azure.  Fix that and add a test with an Openstack DHCP lease so this doesn't happen again.


tas50:
We'll need to get this one merged before we ship the next Ohai otherwise everyone is going to look like they're on a cloud host.

thommay:
:+1:

jkeiser:
:+1:

jkeiser:
Excluding from changelog since it's not a change from the last release.

Issue #749, opened 2016-03-01
KVM modules is a bogus way to detect a KVM host


tas50:
Here's an Openstack guest:

root@ohaitest:~# cat /proc/modules
dm_crypt 23177 0 - Live 0xffffffffa0012000
kvm_intel 143187 0 - Live 0xffffffffa028c000
kvm 455843 1 kvm_intel, Live 0xffffffffa00d3000
serio_raw 13462 0 - Live 0xffffffffa0035000
psmouse 106692 0 - Live 0xffffffffa0019000
floppy 69418 0 - Live 0xffffffffa0000000

Yet:

  "virtualization": {
    "systems": {
      "kvm": "host",
      "openstack": "guest"
    },
    "system": "openstack",
    "role": "guest"
  },

tas50:
Closing this now that we have a PR to fix it

PR #750, opened 2016-03-01
Improve KVM host and guest detection

tas50:
Before we detected KVM hosts via kvm modules, which are actually loaded for both guests and hosts.  Instead use /sys/devices/virtual/misc/kvm to determine if the system is KVM in anyway.  From there look for a hypervisor cpu feature flag to determine if we're a guest.  If that isn't there we're on a host.


tas50:
This resolves #749

thommay:
you really are turning into the cloud whisperer. :+1:

tas50:

lamont-granquist:
:+1:

PR #751, opened 2016-03-01
Detect Openstack hosts

tas50:
Detect openstack hosts by the presence of the nova binary.

This includes the changes in #750 so ignore everything, but the last commit


thommay:
:+1:

lamont-granquist:
:+1:

PR #755, opened 2016-03-03
Remove XML output in VirtualizationInfo and need for hpricot gem

tas50:
This is an early plugin that hasn't received any work since it's creation.  I seriously doubt anyone is using it, but this provides some minor cleanup

1) remove the hpricot dependency and all the commented out hpricot code. This plugin would fail unless it was installed, but it's not actually used
2) Stop storing XML data on the node. This is a horrible horrible thing to do. We're storing the exact same data in JSON. Why 2 formats?
3) Change the debug message to make it clean what plugin fails if the gems aren't there
4) Add a comment telling anyone reading this what gem to use.  The libvirt gem is deprecated and fails to load.

It's a bummer we need this gem to use this because the data is actually pretty useful once it's loaded, but I doubt anyone installs the gem.

### DIFF
--- a/lib/ohai/plugins/virtualization.rb
+++ b/lib/ohai/plugins/virtualization.rb
@@ -16,6 +16,9 @@
 # limitations under the License.
 #
 
+# Note: despite the name this is really a libvirt plugin.
+#       perhaps we'd be better off renaming it? -tsmith
+
 Ohai.plugin(:VirtualizationInfo) do
   %w{ uri capabilities nodeinfo domains networks storage }.each do |info|
     provides "virtualization/#{info}"
@@ -24,8 +27,7 @@ Ohai.plugin(:VirtualizationInfo) do
   collect_data do
     unless virtualization.nil? || !(virtualization[:role].eql?("host"))
       begin
-        require "libvirt"
-        require "hpricot"
+        require "libvirt" # this is the ruby-libvirt gem not the libvirt gem
 
         emu = (virtualization[:system].eql?("kvm") ? "qemu" : virtualization[:system])
         virtualization[:libvirt_version] = Libvirt.version(emu)[0].to_s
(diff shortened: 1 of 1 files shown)


tas50:
"version": "1.2.2",
      "uri": "xen:///system",
      "nodeinfo": {
        "cores": 1,
        "cpus": 1,
        "memory": 1048056,
        "mhz": 2793,
        "model": "x86_64",
        "nodes": 1,
        "sockets": 1,
        "threads": 1
      },
      "domains": {

      },
      "networks": {
        "default": {
          "bridge_name": "virbr0",
          "uuid": "b034d347-1cca-4d96-ba1b-972f19ca947d"
        }
      },
      "storage": {
        "default": {
          "autostart": true,
          "uuid": "0efedb47-cf7e-205a-ab57-bf8e45f406c2",
          "allocation": 2365468672,
          "available": 17308647424,
          "capacity": 19674116096,
          "state": 2,
          "volumes": {
            "test.img": {
              "key": "/var/lib/libvirt/images/test.img",
              "name": "test.img",
              "path": "/var/lib/libvirt/images/test.img",
              "allocation": 0,
              "capacity": 4294967296,
              "type": 0
            },
            "test2.img": {
              "key": "/var/lib/libvirt/images/test2.img",
              "name": "test2.img",
              "path": "/var/lib/libvirt/images/test2.img",
              "allocation": 0,
              "capacity": 4294967296,
              "type": 0
            }
          }

tas50:

thommay:
presumably the gem has native code dependencies? Or does it talk to the libvirt daemon?

tas50:
@thommay It's just a nice way to interfact with libvirt daemon. It's not worth bundling and I think the way we're storing the output is wrong. Unfortunately we screwed up the way we store virtualization information when we added systems and missed a great opportunity to properly nest data under the virtualization mash.  This data would be really great if we could store it somewhere like this.

- Virtualization:
  - kvm
    - function: host
 - libvirt
    - function: host
    - version
    - guests
      - guest 1
      - guest 2

lamont-granquist:
:+1:

smurawski:
:+1:

PR #756, opened 2016-03-04
Improve FreeBSD guest virtualization detection

tas50:
Add debug output for everything we detect
Detect bhyve hosts when /dev/vmm exists (vmm module)
Detect jail hosts with jails in the dying state. They still count
Detect guests with sysctl -n kern.vm_guest output. This detects vmware, xen, hyper-v, and bhyve guests even when dmidecode isn't present (it's not installed by default so big win here)


lamont-granquist:
:+1:

smurawski:
:+1:

PR #764, opened 2016-03-07
Handle mis-identifying Rackspace as EC2 better

tas50:
Be extra sure we're on EC2 in the first place
Timeout querying the metadata quicker. Without this Ohai runs on Rackspace will take 2 minutes

### DIFF
--- a/lib/ohai/mixin/ec2_metadata.rb
+++ b/lib/ohai/mixin/ec2_metadata.rb
@@ -99,7 +99,7 @@ module Ohai
       end
 
       def http_client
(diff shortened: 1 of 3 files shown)


thommay:
:+1: (travis bustage is unrelated)

tas50:
I've tested this on both a rackspace and a EC2 node. Seems to do the trick.

EC2 runtime: real	0m1.003s
Rackspace runtime: real	0m1.334s

The rackspace time was previously 2 minutes 8 secs

mcquin:
:+1:

PR #776, opened 2016-03-16
Don't pull chef from master

danielsdeleo:
When ohai is `bundle installed` without the development group, bundler
doesn't install the development gems but it still includes them in the
gemfile solution. When chef-client is on an unreleased version, this
results in the unreleased version of chef-client ending up in the
Gemfile.lock. Because chef-client has an exact equality pin on
chef-config, which ohai also uses, this results in ohai's Gemfile.lock
pinning ohai to the unreleased version of chef-config. When the ohai
executable is created by appbundler, this results in a broken executable
that tries to load a version of chef-config that doesn't exist.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,8 +3,6 @@ source "https://rubygems.org"
 gemspec
 
 group :development do
-  gem "chef", github: "chef/chef", branch: "master"
-
   gem "sigar", :platform => "ruby"
 
   gem "chefstyle", "= 0.3.0"


danielsdeleo:
Fixes https://github.com/chef/chef-dk/issues/718

mwrock:
:+1:

thommay:
kind of more fuel for https://github.com/chef/chef-rfc/pull/193

tas50:
:+1:

PR #779, opened 2016-03-18
Lock plist to 3.x

danielsdeleo:
Both 3.1 and 3.2 seem to work. Currently Chef is pinned to 3.1.x, so I
set the pin there to avoid conflict with Chef.


adamedx:
:+1:

tas50:
Any reason not to pin to 3.2 and bump chef?

danielsdeleo:
I just didn't want any intermediate builds to fail when eventually bundler's solver will pull in the latest and do the right thing

tas50:
Works for me :+1:

PR #783, opened 2016-03-30
Avoid false detection of systems with the ec2 binary as ec2

tas50:
Ensure the ec2metadata bin actually returns data. Otherwise we're probably not on ec2

### DIFF
--- a/lib/ohai/plugins/ec2.rb
+++ b/lib/ohai/plugins/ec2.rb
@@ -17,6 +17,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# How we detect EC2 from easiest to hardest & least reliable
+# 1. Ohai ec2 hint exists. This always works
+# 2. DMI data mentions amazon. This catches HVM instances in a VPC
+# 3. Has a xen MAC + can connect to metadata. This catches paravirt instances not in a VPC
+# 4. Has ec2metadata binary + can connect to metadata. This catches paravirt instances in a VPC
+
 require "ohai/mixin/ec2_metadata"
 require "base64"
 
@@ -28,18 +34,26 @@ Ohai.plugin(:EC2) do
   depends "network/interfaces"
   depends "dmi"
 
-  # look for ec2metadata which is included on paravirt / hvm AMIs
+  # look for ec2metadata and see if we get data back
+  # this gets us detection of paravirt instances that are within a VPC
   def has_ec2metadata_bin?
     if File.exist?("/usr/bin/ec2metadata")
-      Ohai::Log.debug("ec2 plugin: has_ec2metadata_bin? == true")
-      true
+      # make sure actual data is returned when we run it. Otherwise we might be on GCE or Rackspace
+      if shell_out("/usr/bin/ec2metadata").stdout =~ /id: ami-/
+        Ohai::Log.debug("ec2 plugin: has_ec2metadata_bin? == true")
+        return true
+      else
+        Ohai::Log.debug("ec2 plugin: has_ec2metadata_bin? found ec2metadata but no metadata returned. Not on EC2")
+        return false
+      end
     else
       Ohai::Log.debug("ec2 plugin: has_ec2metadata_bin? == false")
-      false
+      return false
     end
   end
 
-  # look for arp address that non-VPC hosts will have
+  # look for xen arp address
+  # this gets us detection of paravirt instances that are NOT within a VPC
   def has_xen_mac?
     network[:interfaces].values.each do |iface|
       unless iface[:arp].nil?
@@ -60,7 +74,7 @@ EOM
   end
 
   # look for amazon string in dmi bios data
-  # this only works on hvm instances as paravirt instances have no dmi data
+  # this gets us detection of HVM instances that are within a VPC
   def has_ec2_dmi?
     begin
       # detect a version of '4.2.amazon'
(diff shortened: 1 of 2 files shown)


tas50:
Confirmed this fixes false identification on GCE

astoltz:
Fantastic! Thanks!

mcquin:
:+1:

tas50:

coderanger:
This should also help with our misdetection of OpenStack instances outside of Rackspace. :+1: (but I'm not a maintainer)

johnbellone:
I can test this tomorrow, but I am pretty sure that ec2metadata returns results on OpenStack.

johnbellone:
We have a separate patch for GCE. When you have a honeypot name server the metadata address is caught and identities node as such.

tas50:
It's looking for results that specify include 'ami-', which I hope Openstack doesn't include.

tas50:
That being said I'd love to get this tested on Openstack.

johnbellone:
```sh
jbellone@ip-1-71-19-11:~$ uname -a
Linux ip-1-71-19-11 3.13.0-83-generic #127-Ubuntu SMP Fri Mar 11 00:25:37 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
jbellone@ip-1-71-19-11:~$ /usr/bin/ec2metadata |grep ami
ami-id: ami-00000052
ami-launch-index: 0
ami-manifest-path: FIXME
ancestor-ami-ids: unavailable
block-device-mapping: ami
```

tas50:
:(

johnbellone:
```sh
jbellone@ip-1-71-19-11:~$ uname -a
Linux ip-1-71-19-11 3.13.0-83-generic #127-Ubuntu SMP Fri Mar 11 00:25:37 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
jbellone@ip-1-71-19-11:~$ /usr/bin/ec2metadata |grep ami
ami-id: ami-00000052
ami-launch-index: 0
ami-manifest-path: FIXME
ancestor-ami-ids: unavailable
block-device-mapping: ami
```

johnbellone:
```sh
[jbellone@ip-1-71-19-4 ~]$ cat /etc/redhat-release 
Red Hat Enterprise Linux Server release 7.2 (Maipo)
[jbellone@ip-1-71-19-4 ~]$ /usr/bin/ec2metadata |grep ami
ami-id: ami-00000011
ami-launch-index: 0
ami-manifest-path: FIXME
ancestor-ami-ids: unavailable
block-device-mapping: ami
```

johnbellone:
Its same for RHEL instances as well. :(

tas50:
@johnbellone Any chance you could sanitize the complete output from Openstack.  I'd like to use that for the tests once I come up with a better method.

coderanger:
So two thoughts from sitting on the train. First, the whole point of having this elaborate detection for EC2 was to avoid making spurious network requests to the metadata server, but this patch does exactly that. It seems like we might as well do the HTTP get from Ohai directly, saves a few CPU cycles and easier to parse. Second, we can probably filter out both GCE and OpenStack by looking for VPC-related values, because non-VPC machines should be catchable on other detection routines.

johnbellone:
@tas50 https://gist.github.com/johnbellone/0241fb3d96d418ccb2cfd9860733cf21

johnbellone:
FWIW I am digging through some of the cloud config data... no idea what this is
```sh
root@ip-1-71-19-3:/var/lib/cloud/instance# cat obj.pkl |grep -i openstack
aS'OpenStack'
```

tas50:
Back to the drawing board to find a reliable way to detect paravirt systems in a VPC.

lamont-granquist:
One thing you can do is do a DNS lookup on "instance-data.ec2.internal" which will give you back a response fast.  You can't rule out being in EC2 but you can rule it in.

Another idea is that if ohai is running as root and it detects that it is running in EC2 via some method which is reliable that it should drop a hint file so it can pick it up next time.

Those two methods woudl work best together since on first-run it can DNS lookup "instance-data.ec2.internal" when it hits the default amazon nameservers, then it drops a hint file, then after the nameservers have been changed to 8.8.8.8 and the ""instance-data.ec2.internal" no longer works the hint file is still there.

Which suggests another tactic to use which is to do expensive-but-correct things and write out state data to avoid it.  That way you could implement things like hitting 'checkip.dyndns.org' and then doing a reverse DNS lookup on the public/NAT'd IP address to see if it resolves to amazonaws.com or not.

In general, abandon the idea of making ohai always run fast 100% of the time, and accept it taking longer and doing network connections the first time it runs.  Make it manage its own hint state.

lamont-granquist:
I'm also thinking that a certain amount of policy may be applicable here.

- if you drop a hint file (that knife ec2 server create and such creates by default) then we can positively ID you as being in EC2
- if you are running an HVM we can use DMI to positively ID you has being in EC2
- if you fail both of those and are not running paravirt Xen (bare metal or any other hypervisor) then we know you're not EC2

The remaining use cases are slim, and should get slimmer as people SHOULD switch to HVM.

This could just be closed as a problem that we stop trying to solve, and remove all the metadata-endpoint-detection-code and aggressively point people at hinting.

I'm also thinking that most of the tactics I just outlined might better be stuck into a cookbook.  They're all expensive and could be opt-in.  Users who irrationally hate on hinting could be pointed at the 'ohaicloudwtfbbq' cookbook which could do all the work of finding your public/NAT IP, doing the reverse lookup and fixing your hints file for you.

tas50:
So my goal is to avoid the user having to ever drop in a hint, which does put us in a position where we need accurate detection.  We just started adding hints in via provisioning yesterday, and not everyone uses knife-ec2 since it has a lot of rough edges, so 100% reliance on hints isn't realistic.  Hints should be the absolute last resort and we should use them more like we do hints in Azure where we load it with data since there is not metadata endpoint.  I would like to move us to a world where Ohai creates hints to save work on the next chef run. Once we know for certain we're on EC2, relying on hints is perfect. Making that happen is also pretty simple.

All that being said I'm going to close this PR out and open a new PR to remove the ec2metadata command line detection.  It's 100% wrong as is, and results in a lot of false detection.  I have a few other 100% ways to detect EC2 nodes and I'll handle those in other PRs.

PR #785, opened 2016-03-31
Detect EC2 Windows instances

tas50:
This builds on the ec2metdata changes so there will be 3 commits until that's merged


tas50:
Closing this since it's based on the ec2metadata changes

PR #787, opened 2016-03-31
Remove ec2metadata CLI as a EC2 detection method

tas50:
It's not a reliable detection method

### DIFF
--- a/lib/ohai/plugins/ec2.rb
+++ b/lib/ohai/plugins/ec2.rb
@@ -17,6 +17,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# How we detect EC2 from easiest to hardest & least reliable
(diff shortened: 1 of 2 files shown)


tas50:

thommay:
:+1:

btm:
rackspace only "looked like ec2" because it had `/usr/bin/ec2metadata`?

I presume this change is unreliable because other systems had `/usr/bin/ec2metadata` for some kind of compatibility. If that's the case, :+1:

tas50:
ec2metadata is on everything actually.  It was a bad marker for EC2

twmb:
How soon will a patch including this be released? Any logic revolving around which cloud provider you are on will short circuit if ec2 is listed first.

tas50:
Very soon is the most precise answer I can give you right now

PR #789, opened 2016-04-01
Update chefstyle to 0.3.1 and fix new offenses.

mcquin:
@chef/client-core


lamont-granquist:
:+1:

tas50:
:+1: even with all the funky trailing commas

PR #793, opened 2016-04-07
Improve Linux EC2 detection, fix false detection, and add Windows detection

tas50:
Remove the Xen MAC address detection method, which we previously deprecated
Detect Windows hosts based on Kernel owner information
Detect Linux hosts based on the Xen UUID starting with ec2

### DIFF
--- a/lib/ohai/plugins/ec2.rb
+++ b/lib/ohai/plugins/ec2.rb
@@ -43,6 +43,9 @@ Ohai.plugin(:EC2) do
       if dmi[:bios][:all_records][0][:Version] =~ /amazon/
         Ohai::Log.debug("ec2 plugin: has_ec2_dmi? == true")
         true
+      else
+        Ohai::Log.debug("ec2 plugin: has_ec2_dmi? == false")
+        false
       end
     rescue NoMethodError
       Ohai::Log.debug("ec2 plugin: has_ec2_dmi? == false")
@@ -72,6 +75,9 @@ Ohai.plugin(:EC2) do
       if kernel[:os_info][:organization] =~ /Amazon/
         Ohai::Log.debug("ec2 plugin: has_amazon_org? == true")
         true
+      else
+        Ohai::Log.debug("ec2 plugin: has_amazon_org? == false")
+        false
       end
     rescue NoMethodError
       Ohai::Log.debug("ec2 plugin: has_amazon_org? == false")
--- a/lib/ohai/plugins/ec2.rb
+++ b/lib/ohai/plugins/ec2.rb
@@ -20,8 +20,9 @@
 
 # How we detect EC2 from easiest to hardest & least reliable
 # 1. Ohai ec2 hint exists. This always works
+# 2. Xen hypervisor UUID starts with 'ec2'. This catches Linux HVM & paravirt instances
 # 2. DMI data mentions amazon. This catches HVM instances in a VPC
-# 3. Kernel data mentioned Amazon. This catches Windows instances
+# 3. Kernel data mentioned Amazon. This catches Windows HVM & paravirt instances
 
 require "ohai/mixin/ec2_metadata"
 require "base64"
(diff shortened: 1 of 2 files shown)


lib/ohai/plugins/ec2.rb
```diff
@@ -19,8 +20,9 @@
 
 # How we detect EC2 from easiest to hardest & least reliable
 # 1. Ohai ec2 hint exists. This always works
+# 2. Xen hypervisor UUID starts with 'ec2'. This catches Linux HVM & paravirt instances
 # 2. DMI data mentions amazon. This catches HVM instances in a VPC
```
mcquin: Should this line be removed?

spec/unit/plugins/ec2_spec.rb
```diff
@@ -246,33 +250,45 @@
...
-      expect(@plugin[:ec2]["metrics_vhostmd"]).to be_nil
+      expect(plugin[:ec2]).not_to be_nil
+      expect(plugin[:ec2]["metrics"]).to be_nil
+      expect(plugin[:ec2]["metrics_vhostmd"]).to be_nil
     end
   end # shared examples for ec2
 
-  describe "without dmi data, kernel organization, with xen mac, and metadata address connected" do
+  describe "with ec2 dmi data" do
+    it_should_behave_like "ec2"
```
mcquin: You could replace this with `it_behaves_like` ;)

lib/ohai/plugins/ec2.rb
```diff
@@ -69,11 +50,40 @@ def has_ec2_dmi?
...
+  # looks for the Amazon.com Organization in Windows Kernel data
+  # this gets us detection of Windows systems
+  def has_amazon_org?
+    begin
+      # detect an Organization of 'Amazon.com'
+      if kernel[:os_info][:organization] =~ /Amazon/
+        Ohai::Log.debug("ec2 plugin: has_amazon_org? == true")
+        true
+      end
+    rescue NoMethodError
```
mcquin: I'm slightly uncomfortable catching the nonexistance of `kernel[:os_info][:organization]` with an exception. If it's set to not Amazon, then you'll also miss the logging.

tas50: I was pretty aggressive there since the keys aren't going to exist on a lot of platforms.  kernel[:os_info][:organization] only exists on Windows and I haven't checked, but I assume we don't have kernel on every platform.  Do we have a better way to check for the existence of the complete path to that key and it's content?

mcquin: There's a DSL method `attribute?` that may be helpful https://github.com/chef/ohai/blob/8f48805e75e5e6be27da414d00fb8c7f8bbc9378/lib/ohai/dsl/plugin.rb#L124-L128

tas50: so switch that to this?:

if kernel.attribute?('os_info') && kernel[:os_info].attribute?('organization') && kernel[:os_info][:organization] =~ /Amazon/

tas50: After some offline discussion we're going to stick with this for now and add a helper to allow better checking nested Mashes with logging #794

lib/ohai/plugins/ec2.rb
```diff
@@ -62,18 +43,53 @@ def has_ec2_dmi?
...
+  # looks for a xen UUID that starts with ec2
+  # this gets us detection of Linux HVM and Paravirt hosts
+  def has_ec2_xen_uuid?
+    if ::File.exist?("/sys/hypervisor/uuid")
+      if ::File.read("/sys/hypervisor/uuid") =~ /^ec2/
+        Ohai::Log.debug("ec2 plugin: has_ec2_xen_uuid? == true")
+        return true
+      end
+    else
+      Ohai::Log.debug("ec2 plugin: has_ec2_xen_uuid? == false")
```
jkeiser: This won't show up if the file exists and does not contain ec2. If you just remove the else line and put this at the end of the method unconditionally, it will probably do what you want since the if statement terminates with a return.

tas50:
I'm normally not a fan of empty rescues, but I think it makes sense here to avoid duplicating the debug logging in each method.

lib/ohai/plugins/ec2.rb
```diff
@@ -62,18 +43,53 @@ def has_ec2_dmi?
...
+      if kernel[:os_info][:organization] =~ /Amazon/
+        Ohai::Log.debug("ec2 plugin: has_amazon_org? == true")
+        true
+      else
+        Ohai::Log.debug("ec2 plugin: has_amazon_org? == false")
+        false
+      end
+    rescue NoMethodError
+      Ohai::Log.debug("ec2 plugin: has_amazon_org? == false")
+      false
```
jkeiser: You could do the same if else reduction here and save yourself a repeat of this debug line, but this one isn't as important since it works.

jkeiser:
:+1:

tas50:
@mcquin Any chance you can look over the updates to this?

tyler-ball:
:+1:

mcquin:
Looks great @tas50 :+1:

Issue #794, opened 2016-04-07
Create a helper for accessing Mash data


tas50:
It's difficult to check deep Mashes right now without throwing errors that need to be rescued.  Instead of writing repetitive defense code around this it would be nice to create a helper for accessing the methods.  If the attribute data isn't there it would log and return nil, otherwise it'd return the data.  This would prevent the need for things like this:

```ruby
    begin
      if kernel[:os_info][:organization] =~ /Amazon/
        Ohai::Log.debug("ec2 plugin: has_amazon_org? == true")
        true
      end
    rescue NoMethodError
      Ohai::Log.debug("ec2 plugin: has_amazon_org? == false")
      false
    end
  end
```

mcquin:
Notes: We have checkers and fetchers for top-level attributes, but no helpers for subattributes.

PR #796, opened 2016-04-13
Safely get or check the existence of attributes

mcquin:
Resolves https://github.com/chef/ohai/issues/794 by extending the `attribute?` and `get_attribute` DSL methods to apply to sub-attributes. Attributes can be specified as a `/`-delimited string, a list of string keys, a list of symbol keys, or a combination thereof.

Open to suggestions like:

1) Only supporting a list of strings and symbols (get rid of that `/`-delimiter stuff, that's too fancy)
2) Adding similar support to `set_attribute` in this PR (was gonna do it separate since the implementation is different).
3) ___________________________________________________________________________.

### DIFF
--- a/lib/ohai/dsl/plugin.rb
+++ b/lib/ohai/dsl/plugin.rb
@@ -125,7 +125,9 @@ module Ohai
         @data.has_key?(name)
       end
 
-      alias :attribute? :has_key?
+      def attribute?(name, *keys)
+        !safe_get_attribute(name, *keys).nil?
+      end
 
       def set(name, *value)
         set_attribute(name, *value)
@@ -155,8 +157,8 @@ module Ohai
         @data[name]
       end
 
-      def get_attribute(name)
-        @data[name]
+      def get_attribute(name, *keys)
+        safe_get_attribute(name, *keys)
       end
 
       def hint?(name)
@@ -182,6 +184,14 @@ module Ohai
       end
 
       private
+
+      def safe_get_attribute(*keys)
+        keys.inject(@data) { |attrs, key| attrs[key] }
+      rescue NoMethodError, TypeError
+        # NoMethodError occurs when trying to access a key on nil
+        # TypeError occurs when trying to access a key on a value that is not a Hash
+        nil
+      end
 
       def Array18(*args)
         return nil if args.empty?
--- a/lib/ohai/dsl/plugin.rb
+++ b/lib/ohai/dsl/plugin.rb
@@ -186,10 +186,14 @@ module Ohai
       private
 
       def safe_get_attribute(*keys)
-        keys.inject(@data) { |attrs, key| attrs[key] }
-      rescue NoMethodError, TypeError
+        keys.inject(@data) do |attrs, key|
+          unless attrs.nil? || attrs.is_a?(Array) || attrs.is_a?(Hash)
+            raise TypeError.new("Expected Hash but got #{attrs.class}.")
+          end
+          attrs[key]
+        end
+      rescue NoMethodError
         # NoMethodError occurs when trying to access a key on nil
-        # TypeError occurs when trying to access a key on a value that is not a Hash
         nil
       end
 
--- a/spec/unit/dsl/plugin_spec.rb
+++ b/spec/unit/dsl/plugin_spec.rb
@@ -21,11 +21,11 @@ require File.expand_path("../../../spec_helper.rb", __FILE__)
 
 shared_examples "Ohai::DSL::Plugin" do
   context "#initialize" do
-    it "should set has_run? to false" do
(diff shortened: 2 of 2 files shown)


thommay:
:tada: :clap: :+1:

mcquin:
@thommay minor change to require at least one parameter (o/w `attribute?` returns true and `get_attribute` returns all attributes and that would just be not good).

tas50:
+1 :shipit:

mcquin:
Thoughts from @chef/client-core ? :)

thommay:
@mcquin good catch!

jaymzh:
Hmm. Should we instead try to make this match with the way we did `node.rm`? (which is `node.rm('foo', 'bar')`? i.e, take a list? What happens if I have an attribute with a slash in it (which is common in, say, the filesystem plugin).

I like the idea... but I don't dig the implementation.

mcquin:
@jaymzh whoa attributes with `/` characters in them is a real good reason to not support that! I'll rip that out. It currently does support lists of strings and symbols. :)

jaymzh:
Yeah, sorry, it's very early, I only read the description, and hadn't gotten to the code. Which I've now read. Minus the `/` bit, I'm good. Looks great. :+1: from me once that's changed.

mcquin:
Updated to remove support for `/` @jaymzh

lib/ohai/dsl/plugin.rb
```diff
@@ -183,6 +185,15 @@ def method_missing(name, *args)
 
       private
 
+      def safe_get_attribute(*keys)
+        attrs = @data
+        keys.each do |key|
+          return nil unless attrs.respond_to?(:has_key?) && attrs.has_key?(key)
```
lamont-granquist: array indexes?

mcquin: Arrays don't respond to `has_key?`.

lamont-granquist: copied and adapted from some other code of mine (still a bit Before Caffeine for me so I may be overlooking something):

```ruby
begin
  keys.inject(@data) { |memo, key| memo[key] }
rescue NoMethodError
  nil
end
```

mcquin: Ahh that is so nice. It breaks in the case where some intermediate key is not a hash, though. Hm...

mcquin: Haha `rescue NoMethodError, TypeError` works

thommay:
:+1:

coderanger:
There is an RFC declaring `.` is the One True Separator, if we want to support that (though an array should always be supported too for places where that character is in a key).

mcquin:
@coderanger I thought about that. But I didn't want to mix `provides "foo/bar"` and `attribute?("foo.bar")`.

coderanger:
Sure, but the former should be fixed, not the latter :)

mcquin:
🚧

lamont-granquist:
I think its time to argue about RFC 18 tomorrow AM: 

https://github.com/chef/chef-rfc/pull/195

coderanger:
To be clear, I'm still :+1: on this as written. Just if we _did_ want to support the single-string syntax it should be with `.`. This seems like an advanced-mode kind of API though, so I don't think we need to put the max level of UX slickness on it.

jaymzh:
:+1:

tas50:
@chef/client-core can we get another set of eyes on this since there's been some changes since the initial 👍 s

lib/ohai/dsl/plugin.rb
```diff
@@ -183,6 +185,18 @@ def method_missing(name, *args)
 
       private
 
+      def safe_get_attribute(*keys)
+        keys.inject(@data) do |attrs, key|
+          unless attrs.nil? || attrs.is_a?(Hash)
```
lamont-granquist: `|| attrs.is_a?(Array)`?

mcquin: I don't think we want to allow Array access here. The idea is to prevent accessing values as subattribute keys. We expect the structure to be a Mash until we reach a value.

lamont-granquist: Yeah but Mashes turn into Arrays once you traverse them, and `attrs[key]` is fine to run on an Array.

I think if you set it up and try to do something like

`safe_get_attributes("etc", "passwd", "group", "wheel", "members", 0)` 

It'll throw a TypeError instead of giving you back `"root"` (on my Mac anyway).

mcquin: > Mashes turn into Arrays once you traverse them

Whoa, ugh. Ok.

thommay:
👍

PR #797, opened 2016-04-13
Add debug logging to hints and improve cloud specs

tas50:
Adds some basic logging to .hint? and breaks up a large method before making it even larger with debug logging.  Also cleanup how we test  use of hints.  We're already testing the Hint class in its own spec.  There's no need to mock out hint files on the filesystem and test it again in each cloud provider. Instead we should just mock the call to the hint? method itself and save some time. While I was in there I cleaned up the specs, improved some wording, and removed a bunch of specs that didn't actually work.

### DIFF
--- a/lib/ohai/hints.rb
+++ b/lib/ohai/hints.rb
@@ -26,15 +26,13 @@ module Ohai
     end
 
     def self.parse_hint_file(filename)
-      begin
-        json_parser = FFI_Yajl::Parser.new
-        hash = json_parser.parse(File.read(filename))
-        hash || {} # hint
-        # should exist because the file did, even if it didn't
-        # contain anything
-      rescue FFI_Yajl::ParseError => e
-        Ohai::Log.error("Could not parse hint file at #{filename}: #{e.message}")
-      end
+      json_parser = FFI_Yajl::Parser.new
+      hash = json_parser.parse(File.read(filename))
+      hash || {} # hint
+      # should exist because the file did, even if it didn't
+      # contain anything
+    rescue FFI_Yajl::ParseError => e
+      Ohai::Log.error("Could not parse hint file at #{filename}: #{e.message}")
     end
 
     def self.hint?(name)
--- a/lib/ohai/hints.rb
+++ b/lib/ohai/hints.rb
@@ -45,7 +45,7 @@ module Ohai
         @hints[name] = parse_hint_file(filename)
       end
 
-      Ohai::Log.debug("Did not find hint #{name}.json in the hint path(s): #{Ohai.config[:hints_path].join(',')} ") unless @hints.key?(name)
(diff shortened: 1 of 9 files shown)


lib/ohai/hints.rb
```diff
@@ -22,28 +22,30 @@
...
-          rescue FFI_Yajl::ParseError => e
-            Ohai::Log.error("Could not parse hint file at #{filename}: #{e.message}")
-          end
-        end
+        next unless File.exist?(filename)
+        Ohai::Log.debug("Found hint #{name}.json at #{filename}")
+        @hints[name] = parse_hint_file(filename)
       end
 
+      Ohai::Log.debug("Did not find hint #{name}.json in the hint path(s): #{Ohai.config[:hints_path].join(',')} ") unless @hints.key?(name)
```
mcquin: Maybe add a space to the separator, for readability? `', '`

spec/unit/plugins/digital_ocean_spec.rb
```diff
@@ -160,44 +154,13 @@
 
   describe "without digital_ocean hint file" do
     before do
-      allow(File).to receive(:exist?).with(hint_path_nix).and_return(false)
-      allow(File).to receive(:exist?).with(hint_path_win).and_return(false)
-    end
-
-    describe "with the /etc/digitalocean file" do
```
mcquin: I'm not sure that these describe blocks should be removed. Is this code path tested elsewhere?

spec/unit/plugins/digital_ocean_spec.rb
```diff
@@ -160,44 +154,13 @@
...
+      allow(plugin).to receive(:hint?).with("digital_ocean").and_return(false)
     end
 
     describe "with the /etc/digitalocean file" do
       before do
         allow(File).to receive(:exist?).with(digitalocean_path).and_return(true)
-        @plugin.run
+        plugin.run
       end
       it_should_behave_like "digital_ocean_networking"
```
mcquin: `it_behaves_like` :P

spec/unit/plugins/digital_ocean_spec.rb
tas50: The diff makes this look a bit odd, but we tested the same behavior with an EC2 file and with a DO file. We still have the same 2 checks we're just not checking with EC2 stuff anymore.

mcquin: Ah I see it now

mcquin:
LGTM 👍

tas50:
@chef/client-core Can I get a quick review on this one?

thommay:
I think the cleanup of hints looks good, and I trust that you're still testing the proper things :+1:

PR #798, opened 2016-04-13
Don't execute .so libs for Windows

chefsalim:
This fixes an issue where evaluating the C plug-in causes Windows to issue spurious DNS calls because the file path can get evaluated as a UNC in some cases.  We should not be trying to shell out and execute .so files in a non-Unix environment.

@chef/client-core

### DIFF
--- a/lib/ohai/plugins/c.rb
+++ b/lib/ohai/plugins/c.rb
@@ -53,7 +53,7 @@ Ohai.plugin(:C) do
           c[:glibc][:version] = $1
           c[:glibc][:description] = description
         end
-      end unless c[:glibc]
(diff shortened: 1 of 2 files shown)


danielsdeleo:
:+1:

lamont-granquist:
why on earth are we even doing this on unix?  but 👍  yeah, definitely...

what a bizzare API...  i guess it solves the problem of finding out your glibc version without needing to know about your package management tools in a way which is at least portable across linuxes...

adamedx:
:+1:

chefsalim:
For historical reference - http://unix.stackexchange.com/questions/61646/why-are-so-files-executable

PR #799, opened 2016-04-13
Fix the scala plugin

tas50:
There were several problems with the scala plugin:

1) scala -version is returned on stderr not stdout so we were never setting this and thus the plugin never ran

2) If sbt or scala weren't installed it threw exceptions

3) we were storing shell outs return in the output value for no real reason

4) if scala failed for some reason, but sbt succeeded we weren't setting the sbt values since were relied on scala[:version] being set

5) the specs incorrectly mocked scala output to stdout

6) the specs that checked to make sure values weren't set didn't actually run the plugin so they would always no find scala's key

While I was here I moved the version key for sbt into it's own key.  This would allow us to add other sbt values in the future if we found them. If we make version to top level value we lose that ability forever.

### DIFF
--- a/lib/ohai/plugins/scala.rb
+++ b/lib/ohai/plugins/scala.rb
@@ -20,24 +20,31 @@ Ohai.plugin(:Scala) do
   depends "languages"
 
   collect_data(:default) do
-    # Check for scala
-    output = nil
-
     scala = Mash.new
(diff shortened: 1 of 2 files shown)


thommay:
:+1:

mcquin:
Great catches! :+1:

PR #800, opened 2016-04-14
Fix mono builddate capture and add debug logging

tas50:
We weren't capturing the builddate right based on the latest output.  I don't have an example of the old output to know if something changed or if this ever worked right.  We didn't have specs that included the data or a check for the attribute.  Those are added now.

### DIFF
--- a/lib/ohai/plugins/mono.rb
+++ b/lib/ohai/plugins/mono.rb
@@ -37,14 +37,14 @@ Ohai.plugin(:Mono) do
       if so.exitstatus == 0
         mono = Mash.new
         output = so.stdout.split
-        mono[:version] = output[4] if output.length >= 4
-        if output.length >= 11
+        mono[:version] = output[4] unless output[4].nil?
+        if output.length >= 12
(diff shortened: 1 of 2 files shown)


lib/ohai/plugins/mono.rb
```diff
@@ -18,24 +18,33 @@
...
+        output = so.stdout.split
+        mono[:version] = output[4] if output.length >= 4
+        if output.length >= 11
+          mono[:builddate] = "%s %s %s %s %s %s" % [output[7], output[8], output[9], output[10], output[11], output[12].delete!(")")]
+        end
+        languages[:mono] = mono unless mono.empty?
       end
-      languages[:mono] = mono if mono[:version]
+    rescue Ohai::Exceptions::Exec
+      Ohai::Log.debug('Mono plugin: Could not shell_out "mono -v". Skipping plugin')
```
mcquin: `mono -V`?

tas50: yep...

mcquin:
🚀

tas50:

danielsdeleo:
:+1:

PR #808, opened 2016-04-20
Extra debug logging and error handling in plugin loading

tas50:
Pretty simple stuff here. Add the plugin paths and the discovered plugins to debug logging. Add the config too so you can see where we're getting the paths from.  Then if you pass in a bogus dir make sure we warn instead of silently continuing like we do now.

### DIFF
--- a/spec/unit/loader_spec.rb
+++ b/spec/unit/loader_spec.rb
@@ -219,6 +219,14 @@ EOF
           expect { loader.load_plugin(path_to("bad_name.rb")) }.not_to raise_error
         end
       end
+
+      describe "when plugin directory does not exist" do
+        it "logs an invalid plugin path warning" do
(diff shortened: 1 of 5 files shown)


tas50:
@mcquin + @chef/client-core this is ready

mcquin:
:+1:

iennae:
👍

spec/unit/loader_spec.rb
```diff
@@ -219,6 +219,14 @@
           expect { loader.load_plugin(path_to("bad_name.rb")) }.not_to raise_error
         end
       end
+
+      describe "when plugin directory does not exist" do
+        it "logs and invalid plugin path warning" do
```
thommay: "an"

thommay:
teensiest of nits but 👍

PR #813, opened 2016-04-23
Update packages plugin to support PLD Linux as an RPM distro

glensc:
include pld linux as platform with rpm support


tas50:
Confirmed via https://www.pld-linux.org/

thommay:
:+1: LGTM

PR #816, opened 2016-05-04
Adding version for linux module when avalaible

jmauro:
Hi,

I wonder if getting the module version for linux in the 'kernel.modules' could be possible?

I have created the Pull Request to get those.

Let me know if it is good feature or not.

Regards,
JM

### DIFF
--- a/lib/ohai/plugins/kernel.rb
+++ b/lib/ohai/plugins/kernel.rb
@@ -117,6 +117,11 @@ Ohai.plugin(:Kernel) do
     so.stdout.lines do |line|
       if line =~ /([a-zA-Z0-9\_]+)\s+(\d+)\s+(\d+)/
         modules[$1] = { :size => $2, :refcount => $3 }
+        # Making sure to get the module version that has been loaded
+        if File.exist?("/sys/module/#{$1}/version")
+          version = File.read("/sys/module/#{$1}/version").chomp.strip
+          modules[$1]["version"] = version unless version.empty?
+        end
       end
     end
 
--- a/spec/unit/plugins/linux/kernel_spec.rb
+++ b/spec/unit/plugins/linux/kernel_spec.rb
@@ -31,15 +31,41 @@ serio_raw              13031  0
 virtio_balloon         13168  0
 floppy                 55441  0
 ENV_LSMOD
+    @version_module = {
+      dm_crypt: "",
+      psmouse: "",
+      acpiphp: "",
+      microcode: "1.2.3",
+      serio_raw: "",
+      virtio_balloon: "",
+      floppy: "",
+    }
+
+    @expected_result = {
+      "dm_crypt"       => { "size" => "22321", "refcount" => "0" },
+      "psmouse"        => { "size" => "81038", "refcount" => "0" },
+      "acpiphp"        => { "size" => "23314", "refcount" => "0" },
+      "microcode"      => { "size" => "18286", "refcount" => "0", "version" => "1.2.3" },
+      "serio_raw"      => { "size" => "13031", "refcount" => "0" },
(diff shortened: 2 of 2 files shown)


lib/ohai/plugins/kernel.rb
```diff
@@ -116,7 +116,12 @@ def os_lookup(sys_type)
     so = shell_out("env lsmod")
     so.stdout.lines do |line|
       if line =~ /([a-zA-Z0-9\_]+)\s+(\d+)\s+(\d+)/
+        _status, version, _stderr = run_command(
```
mcquin: `run_command` will be deprecated in Ohai 9.0, please use `shell_out`.

Using `cat` to get system information is a bit unusual in Ohai. Does `"modinfo #{mod} -F version"` work?

spec/unit/plugins/linux/kernel_spec.rb
```diff
@@ -31,15 +31,40 @@
...
+    @version_module.each do |mod, vers|
+      allow(@plugin).to receive(:run_command).with({ command: "cat /sys/module/#{mod}/version", :no_status_check => true }).and_return([1, vers, ""])
+    end
     expect(@plugin).to receive(:shell_out).with("env lsmod").at_least(1).times
     @plugin.run
   end
 
   it_should_check_from_deep_mash("linux::kernel", "kernel", "os", "uname -o", [0, "Linux", ""])
 
+  it "Testing linux::kernel::modules" do
```
mcquin: The string here should be a description of the test. For example, `it "collects linux::kernel::modules" do`.

mcquin:
Hi @jmauro , I think collecting this information makes sense. I have a little feedback, and otherwise the code looks great! Thanks for adding tests. :)

lib/ohai/plugins/kernel.rb
jmauro: OK, I did not know about `run_command`. 

Regarding the `"modinfo #{mod} -F version"` I am not confortable with this command because it only reflects the module present on the file system not the version loaded. I will figured out a way to do it.

spec/unit/plugins/linux/kernel_spec.rb
jmauro: will do.

jmauro:
Everything is fixed. Let me know if I need to do something else.

mcquin:
ping for review @chef/client-core

thommay:
Not super keen on `shell_out`ing to cat, rather than just reading the file natively.

jmauro:
Fair enough

lib/ohai/plugins/kernel.rb
```diff
@@ -116,7 +116,9 @@ def os_lookup(sys_type)
     so = shell_out("env lsmod")
     so.stdout.lines do |line|
       if line =~ /([a-zA-Z0-9\_]+)\s+(\d+)\s+(\d+)/
```
btm: I personally like putting a comment with the line I'm matching against above a regex like this, so without leaving the code you can figure out what $1 is supposed to be.

btm:
I agree that reading the file using Ruby is preferred as shellout has higher performance costs. Of course you're going to have to deal with handing cases if the file is missing and all that, so it's a little more work.

lib/ohai/plugins/kernel.rb
jmauro: Hey, no my writing :smile:

jmauro:
Ok guys, I have done the changes you wanted let me know if you need something else.

Regards,
JM

thommay:
:+1: LGTM

tas50:
👍

PR #817, opened 2016-05-06
WIP: Properly detect Openstack + fetch metadata

tas50:
This is a continuation of the work by @acaiafa to make this plugin actually fetch metadata. At the moment it is 100% broken.

Changes:
- Remove HP Cloud logic
- If the hint is present add the mash even if metadata can't be fetched
- Properly fetch metadata (before we had non-functional copy pasta from ec2)
- Add additional debug output
- Fetch metadata if we have openstack metadata

Updated tests coming soon


tas50:
Closing this out in favor of the much simpler #818

PR #818, opened 2016-05-09
Properly poll Openstack detection + other Openstack improvements

tas50:
The existing Openstack metadata logic failed as it was hardcoded to use metadata from 2013 that doesn't seem to exist in standard installs. Instead lets use the logic in the EC2 metadata mixin to find the latest version we support.
Wrap metadata fetching around  socket connection check from the EC2 metadata plugin. This prevents hung Ohai runs on providers that don't support the metadata endpoint
Grab the Openstack data correctly in the Cloud plugins
Detect a system as being on Openstack based on DMI data
Remove HP cloud logic since that's dead
Add debug logging so we can follow what's happening in the logs better

Previously nil data we now have:

```json
 "cloud": {
    "public_ips": [
      "172.16.20.52"
    ],
    "private_ips": [
      "192.168.1.203"
    ],
    "public_ipv4": "172.16.20.52",
    "public_hostname": "ohai.novalocal",
    "local_ipv4": "192.168.1.203",
    "local_hostname": "ohai.novalocal",
    "provider": "openstack"
  },
  "cloud_v2": {
    "public_ipv4_addrs": [
      "172.16.20.52"
    ],
    "local_ipv4_addrs": [
      "192.168.1.203"
    ],
    "provider": "openstack",
    "public_hostname": "ohai.novalocal",
    "local_hostname": "ohai.novalocal",
    "public_ipv4": "172.16.20.52",
    "local_ipv4": "192.168.1.203"
  },
 "openstack": {
    "provider": "openstack",
    "metadata": {
      "ami_id": "ami-00000002",
      "ami_launch_index": "0",
      "ami_manifest_path": "FIXME",
      "block_device_mapping_ami": "vda",
      "block_device_mapping_root": "/dev/vda",
      "hostname": "ohai.novalocal",
      "instance_action": "none",
      "instance_id": "i-00000211",
      "instance_type": "m1.medium",
      "kernel_id": "None",
      "local_hostname": "ohai.novalocal",
      "local_ipv4": "192.168.1.203",
      "placement_availability_zone": "nova",
      "public_hostname": "ohai.novalocal",
      "public_ipv4": "172.16.20.52",
      "public_keys_0_openssh_key": "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQCxAVkrfknXoHFQtyXtYZ6vONLHn7HTKQq4mkwvCOPMKb3gr6gsLQkCdzmO3CQwOcFl7uDWGdfjH/ocuSUSn6xMvfCdsdGzDiLQXcDJ4jGOuuyOioniN1N00IhAD7Tqsf4AvbQJ7gDqfPKDCw6LdlZ0d0d9pWlJf7eKf/Hq3x8RCqWek/+UCoU4l4T1eAM0N/bA5tj3EfvRuDniO8U5ClsuUbtJ1uxJmrWMhFox3jFp6oNrleHWFTiRsNKfbtfTrTJRYlqcOG2jGAb49JY+p6Ykku4vq0ezYZFYXGJki6a0yLu1kMg5aHJo9UX0d9JW6y11KfXbGKN4okeRcH/tSNov jasghar@remasghar01.local",
      "ramdisk_id": "None",
      "reservation_id": "r-nvo4xzvf",
      "security_groups": [
      ]
    }
```

### DIFF
--- a/lib/ohai/plugins/cloud.rb
+++ b/lib/ohai/plugins/cloud.rb
@@ -189,7 +189,7 @@ Ohai.plugin(:Cloud) do
     cloud[:public_hostname] = openstack["public_hostname"]
     cloud[:local_ipv4] = openstack["local_ipv4"]
     cloud[:local_hostname] = openstack["local_hostname"]
-    cloud[:provider] = "openstack"
+    cloud[:provider] = openstack["provider"]
   end
 
   # ----------------------------------------
--- a/lib/ohai/plugins/cloud.rb
+++ b/lib/ohai/plugins/cloud.rb
@@ -183,12 +183,12 @@ Ohai.plugin(:Cloud) do
 
   # Fill cloud hash with openstack values
   def get_openstack_values
-    cloud[:public_ips] << openstack["metadata"]["public_ipv4"]
-    cloud[:private_ips] << openstack["metadata"]["local_ipv4"]
-    cloud[:public_ipv4] = openstack["metadata"]["public_ipv4"]
-    cloud[:public_hostname] = openstack["metadata"]["public_hostname"]
-    cloud[:local_ipv4] = openstack["metadata"]["local_ipv4"]
-    cloud[:local_hostname] = openstack["metadata"]["local_hostname"]
+    cloud[:public_ips] << openstack["public_ipv4"]
+    cloud[:private_ips] << openstack["local_ipv4"]
+    cloud[:public_ipv4] = openstack["public_ipv4"]
+    cloud[:public_hostname] = openstack["public_hostname"]
+    cloud[:local_ipv4] = openstack["local_ipv4"]
+    cloud[:local_hostname] = openstack["local_hostname"]
     cloud[:provider] = "openstack"
   end
 
--- a/lib/ohai/plugins/cloud.rb
+++ b/lib/ohai/plugins/cloud.rb
@@ -183,13 +183,13 @@ Ohai.plugin(:Cloud) do
 
   # Fill cloud hash with openstack values
   def get_openstack_values
-    cloud[:public_ips] << openstack["public_ipv4"]
-    cloud[:private_ips] << openstack["local_ipv4"]
-    cloud[:public_ipv4] = openstack["public_ipv4"]
-    cloud[:public_hostname] = openstack["public_hostname"]
-    cloud[:local_ipv4] = openstack["local_ipv4"]
-    cloud[:local_hostname] = openstack["local_hostname"]
-    cloud[:provider] = openstack["provider"]
+    cloud[:public_ips] << openstack["metadata"]["public_ipv4"]
+    cloud[:private_ips] << openstack["metadata"]["local_ipv4"]
+    cloud[:public_ipv4] = openstack["metadata"]["public_ipv4"]
+    cloud[:public_hostname] = openstack["metadata"]["public_hostname"]
+    cloud[:local_ipv4] = openstack["metadata"]["local_ipv4"]
+    cloud[:local_hostname] = openstack["metadata"]["local_hostname"]
+    cloud[:provider] = "openstack"
   end
 
   # ----------------------------------------
--- a/lib/ohai/plugins/openstack.rb
+++ b/lib/ohai/plugins/openstack.rb
@@ -1,5 +1,6 @@
 #
 # Author:: Matt Ray (<matt@chef.io>)
+# Author:: Tim Smith (<tsmith@chef.io>)
 # Copyright:: Copyright (c) 2012-2016 Chef Software, Inc.
 # License:: Apache License, Version 2.0
 #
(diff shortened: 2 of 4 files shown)


jjasghar:
:metal: this is awesome sauce.

tas50:
This is ready for review @chef/client-core @mcquin @btm

lib/ohai/plugins/openstack.rb
```diff
@@ -18,53 +19,67 @@
...
+    else
+      Ohai::Log.debug("Plugin Openstack: openstack hint not present")
+      return false
+    end
+  end
 
-  def collect_openstack_metadata(addr = Ohai::Mixin::Ec2Metadata::EC2_METADATA_ADDR, api_version = "2013-04-04")
-    path = "/openstack/#{api_version}/meta_data.json"
-    uri = "http://#{addr}#{path}"
+  # dreamhost systems hae the dhc-user on them
```
thommay: `have`

lib/ohai/plugins/openstack.rb
```diff
@@ -18,53 +19,67 @@
...
-        Ohai::Log.debug("Encountered 404 response retreiving OpenStack specific metadata path: #{path} ; continuing.")
-        nil
-      else
-        raise "Encountered error retrieving OpenStack specific metadata (#{path} returned #{response.code} response)"
+      return "dreamhost" if etc["passwd"]["dhc-user"]
+    rescue NoMethodError
+      # handle etc not existing on non-linux systems
+    end
+    return "openstack"
+  end
```
thommay:
```ruby
def openstack_provider
  "dreamhost" if  etc["passwd"]["dhc-user"]
rescue NoMethodError
  "openstack"
end
```

might be cleaner?

thommay:
some tweaks, but 👍

lib/ohai/plugins/openstack.rb
tas50: I had it that way and it ends up returning nil when it can't find the key, which isn't what you want.

Scenarios:
etc plugin doesn't exist at all and the rescue is necessary -> openstack
etc plugin exists, but dhc-user doesn't exist -> openstack
etc plugin exists and dhc-user does exist -> dreamhost

lib/ohai/plugins/openstack.rb
tas50: fixed

lib/ohai/plugins/cloud.rb
```diff
@@ -189,7 +189,7 @@ def get_openstack_values
     cloud[:public_hostname] = openstack["public_hostname"]
     cloud[:local_ipv4] = openstack["local_ipv4"]
     cloud[:local_hostname] = openstack["local_hostname"]
-    cloud[:provider] = openstack["provider"]
+    cloud[:provider] = "openstack"
```
mcquin: It looks like this can sometimes be "dreamhost"?

lib/ohai/plugins/openstack.rb
```diff
@@ -18,53 +19,62 @@
...
+      openstack[:provider] = openstack_provider
 
-        if hint?("hp")
-          openstack["provider"] = "hp"
-        else
-          openstack["provider"] = "openstack"
-          Ohai::Log.debug("connecting to the OpenStack specific metadata service")
-          openstack["metadata"] = collect_openstack_metadata
+      # fetch the metadata if we can do a simple socket connect first
+      if can_metadata_connect?("169.254.169.254", 80)
```
mcquin: Do you want to use `Ohai::Mixin::Ec2Metadata::EC2_METADATA_ADDR`?

lib/ohai/plugins/cloud.rb
tas50: So this one gets a bit weird and here's why I went with openstack. When you're in the context of the cloud plugin says provider you'd think openstack (right? maybe...). When you're in the context of the openstack plugin you know you're openstack so when someone says provider you're thinking about what openstack provider and that could be dreamhost. I guess it could really go either way though.

mcquin: It could be breaking for some users. Though I understand your argument, I think it's safer to keep it as it was.

tas50: The previous behavior was HP or Openstack, but I'll switch this one back.  People on Dreamhost can figure it out probably.

lib/ohai/plugins/openstack.rb
tas50: While it is the same thing it seems a bit odd to me to use that variable

mcquin: You're using `fetch_metadata` which relies on that variable. Best to use it in the unlikely case something changes.

mcquin: The variable is referenced in the tests, as well.

mcquin:
👍 🚀 🎉

Issue #819, opened 2016-05-11
packages provider inconsistent results


glensc:
## Description

i find debian packages listing inconsistent with one that rpm provides.

 - rpm system has "release" portion
 - deb system "release" is included in "version" portion
 - some deb package names contain archtecture `zlib1g:amd64`
 - debian versions also include "epoch"

ideally i'd see:
 - rpm systems include extra `arch` tag
 - deb systems include `arch` tag, extract it from `name` if needed
 - deb systems include `release` tag
 - epoch moved to extra tag

# examples from systems

rpm system:
```
# knife node show fqdn -a packages
pld:
  packages:
    glibc-misc:
      release: 21
      version: 2.3.6
    glibc64:
      release: 21
      version: 2.3.6
```
debian-7 system:
```
# knife node show fqdn -a packages
debian:
  packages:
    acpi:
      version: 1.6-1
    zlib1g:amd64:
      version: 1:1.2.7.dfsg-13
```
## Ohai Version

Ohai: 7.4.0, but packages is taken from ohai that was merged in v8.9.0
18ae02f5e4dc0ff096271f39fd1a970a0adc52c7

## Platform Version

Debian 7

glensc:
#820 was merged adding:
- rpm: add epoch, arch
- deb: add arch
- deb: fix arch being part of the name attribute

however, there's still todo:
- deb: explode version to components (epoch, version, release), see questions in #820
- deb: add installdate attribute

PR #820, opened 2016-05-11
improve packages attribute. #819

glensc:
improve packages attribute. #819

- rpm: add epoch, arch
- deb: add arch

i tried to explode the version component on debian systems, but wasn't sure what are the component parts:

- 3.4.2-1
- 1:1.0.9-1
- 4.1+Debian13+nmu1
- 5.1.1alpha+20120614-2+b3

rpm has `%{INSTALLTIME}`, add it as `installdate` attribute? but what format? unix timestamp? (it's the easiest one)

(tests are expected to fail, until get some feedback)


glensc:
| ${Version} | epoch | version | release |
| ------------- | ------------- | ------------- | ------------- |
| 3.4.2-1 | nil | 3.4.2 | 1 |
| 1:1.0.9-1 | 1 | 1.0.9 | 1 |
| 4.1+Debian13+nmu1 | nil | ? | ? |
| 5.1.1alpha+20120614-2+b3 | nil | ? | ? |

glensc:
found some documentation of Debian ${Version} components: http://linux.die.net/man/5/deb-version

seems not very simplie split and go logic here... suggestions?

glensc:
for consistency should deb/rpm use or not to use `shellwords`. i would drop the shellwords because the input is fixed string (not containing user input).

mcquin:
Can someone from @chef/client-ubuntu lend a hand to answer @glensc question at https://github.com/chef/ohai/pull/820#issuecomment-218443156

lamont-granquist:
i think shellwords is likely there to prevent shell expansion of special characters in the string in question so that you don't have to go insane thinking about everything you'd need to backwhack, not necessarily about security...

glensc:
my string is between `''`, so it's never expanded

lamont-granquist:
👍

lamont-granquist:
spec test need to be updated to make it pass, its failing everything right now

glensc:
@lamont-granquist i specifically wrote in summary:
> (tests are expected to fail, until get some feedback)

and haven't got the necessary feedback!

glensc:
so, i assume nobody (from ohai maintainers) cares what the result is and will be merged as is? i'm personally are satisfied with current result of this PR.

glensc:
also, how do you expect me to update the unit test?
 
replace `spec/data/plugins/dpkg-query.output` and `spec/data/plugins/rpmquery.output` from my system?

glensc:
rpm data replaced with output from

`docker run --rm -it centos bash`

and truncated it

lamont-granquist:
yeah that looks fine, i'd prefer to have slightly tighter fixtures so it was more obvious what was important and what wasn't...

glensc:
so, ready for merge?

lamont-granquist:
👍

lamont-granquist:
@thommay whack merge if you're satisfied...

thommay:
thanks, @glensc!

PR #821, opened 2016-05-11
Cleanup attribute access in cloud plugins

tas50:
Use get_attribute to simplify code in places where we previously had to rescue NoMethodError


tas50:
Ready for review @chef/client-core

mcquin:
👍

tas50:
Care to take a look @thommay

thommay:
technically you and claire were enough eyes on this (two maintainers), but 👍 regardless

PR #822, opened 2016-05-11
Extend set_attribute to set sub-attributes.

mcquin:
Use:
- set_attribute(:attr, value) assigns attribute 'attr' value.
- set_attribute(:attr, :subattr, value) assigns attribute 'attr/subattr' value. Extends to subattributes of arbitrary depth.

Raises a TypeError if some :subattr is not an Array, Hash, or Mash (e.g., a String value).

@tas50

### DIFF
--- a/lib/ohai/dsl/plugin.rb
+++ b/lib/ohai/dsl/plugin.rb
@@ -152,8 +152,19 @@ module Ohai
         end
       end
 
-      def set_attribute(name, *values)
-        @data[name] = Array18(*values)
+      def set_attribute(name, *attrs, value)
+        # Initialize the path in the @data Mash with new Mashes, if needed.
+        # Will raise a TypeError if we hit a subattribute that is not a
+        # Hash, Mash, or Array.
+        keys = [name] + attrs
+        attribute = keys[0..-2].inject(@data) do |attrs, key|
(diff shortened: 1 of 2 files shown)


mcquin:
Ready for review @chef/client-core

lib/ohai/dsl/plugin.rb
```diff
@@ -152,8 +152,19 @@ def from_with_regex(cmd, *regex_list)
         end
       end
 
-      def set_attribute(name, *values)
-        @data[name] = Array18(*values)
+      def set_attribute(name, value, *values)
```
lamont-granquist: can this just be `set_attribute(*keys, value)`?

lamont-granquist:
You might want to look at chef/chef#5029, I'm implementing functional getter/setter APIs, but the words I'm using is read/write/unlink, also with a slightly different implementation.

mcquin:
That's a neat implementation @lamont-granquist . Are we okay with merging this as-is? I don't see the need for an implementation re-write now. Perhaps a future a mixlib-attribute library could prove useful for unifying attribute access between Chef and Ohai. ;)

tas50:
@mcquin @lamont-granquist Any update on this?

PR #823, opened 2016-05-11
Add Digital Ocean metadata from their metadata API

tas50:
At the moment we rely on hint data from our digital ocean knife plugin.  Not everyone boostraps with knife though.  Instead of relying on the hint data we should detect digital ocean nodes via DMI or hint file and then fetch all the DO metadata from their API. This builds on the original PR from @GenPage but adds detection of DO droplets via DMI data and updates the cloud plugins.

### DIFF
--- a/lib/ohai/mixin/do_metadata.rb
+++ b/lib/ohai/mixin/do_metadata.rb
@@ -54,17 +54,18 @@ module Ohai
         Net::HTTP.start(DO_METADATA_ADDR).tap { |h| h.read_timeout = 6 }
       end
 
-      def fetch_metadata()
+      def fetch_metadata
         uri = "#{DO_METADATA_URL}"
         response = http_client.get(uri)
         case response.code
         when "200"
-          response.body
+          parser = FFI_Yajl::Parser.new
+          parser.parse(response.body)
         when "404"
-          Ohai::Log.debug("Encountered 404 response retreiving Digital Ocean metadata: #{uri} ; continuing.")
-          Hash.new
+          Ohai::Log.debug("DOMetadata mixin:Encountered 404 response retreiving Digital Ocean metadata: #{uri} ; continuing.")
+          {}
         else
-          raise "Encountered error retrieving Digital Ocean metadata (#{uri} returned #{response.code} response)"
+          raise "DOMetadata mixin:Encountered error retrieving Digital Ocean metadata (#{uri} returned #{response.code} response)"
         end
       end
 
--- a/lib/ohai/mixin/do_metadata.rb
+++ b/lib/ohai/mixin/do_metadata.rb
@@ -21,7 +21,7 @@ module Ohai
   module Mixin
     module DOMetadata
 
-      DO_METADATA_ADDR = "169.254.269.254" unless defined?(DO_METADATA_ADDR)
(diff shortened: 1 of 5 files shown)


tas50:
I've tested this on a new Ubuntu 16.04 droplet and here's the old code vs. the new code. It should also be noted that the old code failed to auto detect the node as DO so I had to use a hint, which is 👎 

```javascript
  "digital_ocean": {
    "networks": {
      "v4": [
        {
          "ip_address": "138.68.99.253",
          "type": "public",
          "netmask": "255.255.240.0"
        },
        {
          "ip_address": "10.19.0.5",
          "type": "private",
          "netmask": "255.255.0.0"
        }
      ],
      "v6": [
        {
          "ip_address": "2a03:b0c0:0003:00d0:0000:0000:322a:3001",
          "type": "public",
          "cidr": "128"
        },
        {
          "ip_address": "fe80:0000:0000:0000:d4b1:9eff:fe61:8cce",
          "type": "private",
          "cidr": "128"
        }
      ]
    }
  }
```

```javascript
"digital_ocean": {
    "droplet_id": 12345678,
    "hostname": "mytestnode",
    "public_keys": [
      "ssh-rsa SOMEKEY",
    ],
    "auth_key": "SOMEKEY",
    "region": "fra1",
    "interfaces": {
      "public": [
        {
          "ipv4": {
            "ip_address": "138.68.99.253",
            "netmask": "255.255.240.0",
            "gateway": "138.68.96.1"
          },
          "ipv6": {
            "ip_address": "2A03:B0C0:0003:00D0:0000:0000:322A:3001",
            "cidr": 64,
            "gateway": "2A03:B0C0:0003:00D0:0000:0000:0000:0001"
          },
          "anchor_ipv4": {
            "ip_address": "10.19.0.5",
            "netmask": "255.255.0.0",
            "gateway": "10.19.0.1"
          },
          "mac": "d6:b1:9e:61:8c:ce",
          "type": "public"
        }
      ]
    },
    "floating_ip": {
      "ipv4": {
        "active": false
      }
    },
    "dns": {
      "nameservers": [
        "2001:4860:4860::8844",
        "2001:4860:4860::8888",
        "8.8.8.8"
      ]
    },
    "tags": null
  },
```

bdwyertech:
Seems like `cloud` and `cloud_v2` aren't getting populated, even though `digital_ocean` is.  Chef 13.0.62, can't seem to get it to pick up the hint either, in `/etc/digitalocean` or `/etc/chef/ohai/hints/digital_ocean.json`

tas50:
@bdwyertech I'll take a look and see what was missed. What OS was this on?

tas50:
@bdwyertech Looks like it's parsing the old format in the cloud plugins. I think I might have busted this with a rebase. I'll dig in and get the cloud/cloudv2 parsing fixed before we ship. Thanks for noticing this.

tas50:
I've opened https://github.com/chef/ohai/issues/966 to track fixing the cloudv2 plugin

bdwyertech:
Sorry @tas50 didn't get the Github pings for some reason...  Thanks for nipping that, I can finally sunset all my DO-specific monkey patches. 🍻

PR #827, opened 2016-05-25
expose ring parameters in the network plugin

davide125:
Extend the network plugin to expose interface ring parameters using ethtool.

I'm a Facebook employee and @jaymzh will do the final merge.

### DIFF
--- a/lib/ohai/plugins/linux/network.rb
+++ b/lib/ohai/plugins/linux/network.rb
@@ -178,24 +178,24 @@ Ohai.plugin(:Network) do
       next unless iface[tmp_int][:encapsulation] == "Ethernet"
       so = shell_out("#{ethtool_binary} -g #{tmp_int}")
       Ohai::Log.debug("Parsing ethtool output: #{so.stdout}")
-      # Nasty, brittle regex to capture the ethtool output
-      nasty_regex =
-/Ring parameters for #{tmp_int}:
-Pre-set maximums:
-RX:\s+(\d+)
-.*
-TX:\s+(\d+)
-Current hardware settings:
-RX:\s+(\d+)
-.*
-TX:\s+(\d+)/m
-      if nasty_regex.match(so.stdout)
-        iface[tmp_int]["ring_params"] = {
-          "max_rx" => $1.to_i,
-          "max_tx" => $2.to_i,
-          "current_rx" => $3.to_i,
-          "current_tx" => $4.to_i,
-        }
+      type = nil
+      iface[tmp_int]["ring_params"] = {}
+      so.stdout.lines.each do |line|
+        next if line.start_with?("Ring parameters for")
+        next if line.strip.nil?
+        if line =~ /Pre-set maximums/
+          type = "max"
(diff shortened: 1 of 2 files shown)


chef-supermarket:

* @davide125

lamont-granquist:
@davide125 you should probably still do the minimum viable clickiness to shut up currybot, even though you're covered under facebook...

thommay:
👍

lamont-granquist:
i find it amusing that you care this much about the TX and RX ring buffer settings in your infrastructure (having worked at amazon for 5 years back in the dark ages where we abused the crap out of all our networking equipment as well...)

:+1:

jaymzh:
@lamont-granquist as mentioned in several meetings with various members of Chef, no one else at Facebook can sign the CCLA... cause reasons. Legal stuff sucks. That's why we have this stupid "phil has to merge it to make sure it legally falls under phil's CCLA" crap.

lamont-granquist:
sigh, lawyers...   okay...

lib/ohai/plugins/linux/network.rb
```diff
@@ -171,6 +171,36 @@ def ethernet_layer_one(iface)
...
   end
 
+  # determine ring parameters for the interface using ethtool
+  def ethernet_ring_parameters(iface)
+    return iface unless ethtool_binary = find_ethtool_binary
+    iface.each_key do |tmp_int|
+      next unless iface[tmp_int][:encapsulation] == "Ethernet"
+      so = shell_out("#{ethtool_binary} -g #{tmp_int}")
+      Ohai::Log.debug("Parsing ethtool output: #{so.stdout}")
+      # Nasty, brittle regex to capture the ethtool output
```
mcquin: Is there anything that would make this less nasty and brittle? Multiple queries for each match? Or just that we're reading from something that may change?

davide125: I'm open to suggestions. The main thing that makes this brittle to me is the reliance on perfect output matching, which means that any change in ethtool formatting of this would break it.

jaymzh: I can think of two things to make this less brittle:

1. Easy: Don't assume "RX" will always be first, so stick a .* before that in case they add preceeding lines

2. A bit less easy: Parse the output line by line so you don't assume any order:

```
type = nil
lines.each do |line|
  next if line.start_with?('Ring params for')
  next if line.strip.nil?
  if line =~ /Pre-set maximums/
    type = 'max'
    next
  end
  if line =~ /Current hardware settings/
    type = 'current'
    next
  end
  key, val = line.split(/:\s+/)
  iface[tmp_int]["ring_params"]["#{type}_{key.downcase}"] = val
end
```

This will also pull in other settings which will be awesome.

jaymzh:
This looks good to me, I'm :+1:  - though I'm a bit biased, cause I basically write this version. :)

Can one of you @chef/client-core ack this version as well before I merge?

mcquin:
💯

PR #828, opened 2016-05-27
Switch to kernel version to identify Gentoo versions

tas50:
We're looking at the gentoo release file to identify the version currently which is 2.2. It's been 2.2 for years since that's just the Gentoo Base System version which never changes. We should do what we do on Arch and use the kernel version string.  This is a copy / paste of that code.


thommay:
ugh, but 👍

PR #829, opened 2016-06-13
Add a simple plugin to get the local timezone.

johnbellone:
We have an internal plugin where we are doing the same thing, and we were a little surprised that this didn't exist yet. I didn't think there was a better namespace place to shove this under.

### DIFF
--- a/lib/ohai/plugins/timezone.rb
+++ b/lib/ohai/plugins/timezone.rb
@@ -1,0 +1,23 @@
+#
+# Author:: John Bellone (<jbellone@bloomberg.net>)
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


thommay:
@smurawaski can you take a look at the powershell failures in appveyor please?

mcquin:
@johnbellone Great addition! Since we don't have a lot of other plugins collecting time information, I don't feel like trying to nest this under a `Time` plugin makes sense (it was my initial thought, but we don't have a good enough understanding of what the data we want to collect will look like). If we do decide to scope this under another plugin, we can add it as a dependency.

:+1: ⌚

johnbellone:
@mcquin Sounds good to me.

tas50:
👍

lamont-granquist:
so this probably breaks half the timezone cookbooks out there in the rest of the world.

`node.automatic["timezone"]` will override `node.default["timezone"]` in any cookbook.  lots of those use a sub-hash to configure things so it'll obliterate the "schema"

https://github.com/L2G/timezone-ii/blob/master/attributes/default.rb#L7-L15

in my own cookbooks I use `node.default["timezone"]["value"]`  and now my runs crash with:

```
  NoMethodError
  -------------
  undefined method `value' for "PDT":String
```

lamont-granquist:
yeah i think unfortunately, i'm going to whack revert on this, i don't think we want this going out in the next chef release without more discussion

johnbellone:
No worries, makes sense to me. Do you want to leave it open?

tas50:
@johnbellone We moved everything under node['time'] and shipped 8.17.1. It's all good now

PR #830, opened 2016-06-15
Add additional info to networking interfaces/addresses

jaymzh:
I'd like to know if addresses are dynamic or not, for example.

### DIFF
--- a/lib/ohai/plugins/linux/network.rb
+++ b/lib/ohai/plugins/linux/network.rb
@@ -310,10 +310,18 @@ Ohai.plugin(:Network) do
   end
 
   def parse_ip_addr_inet6_line(cint, iface, line)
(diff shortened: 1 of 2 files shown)


jaymzh:
CC @chef/client-core 
Also... does app-veyer ever work?

smurawski:
@jaymzh Yeah, appveyor has been having a rough week or so with our code.There is a broken functional test because of how we parsed out the powershell compatible versions which broke when we went from powershell 4 to 5 on the build agents.  (https://github.com/chef/ohai/issues/831).  This new error seems to be releated to the rubygems version (at least that's the only difference I see with test runs that pass and fail.  I'm troubleshooting in a PR against test-kitchen).

smurawski:
Oh and :+1: on the PR

mcquin:
LGTM 🚀

tas50:
👍

PR #832, opened 2016-06-16
Update PowerShell Version Compat Detection / Unblock bundler on Appveyor

### LINKED ISSUE #831: PowerShell Compatible Version Detection
We've got functional tests breaking because we parse the output of `$psversiontable` for PowerShell compatible versions, when we need to look at `$psversiontable.pscompatibleversions` to get the full array.   The higher level view uses '...' to indicate more versions and that breaks the current regex.

/cc @thommay - this is the appveyor breakage after they moved their base build image to ps5.  With ps4, everything fit in the summary table.

smurawski:
- [x] Update overcommit to prevent install failures on Windows.
- [x] Update PowerShell Version Compat Detection Resolves #831

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,6 +6,8 @@ group :development do
   gem "sigar", :platform => "ruby"
 
   gem "chefstyle", "= 0.3.1"
-  # gem 'pry-byebug'
-  # gem 'pry-stack_explorer'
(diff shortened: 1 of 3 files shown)


smurawski:
Cool.  Now we are just back to breaking on #831

lamont-granquist:
👍

PR #835, opened 2016-06-30
Revert "Add a simple plugin to get the local timezone."

lamont-granquist:
Reverts chef/ohai#829


tas50:
Instead of reverting this I'm going to move it under 'time'

PR #836, opened 2016-06-30
Move timezone value under time

tas50:
This prevents us from breaking the various timezone cookbooks that
already use that namespace. Attributes are hard. Good thing lamont is
going to fix this all ;)


tas50:
Ping @lamont-granquist @mcquin

lamont-granquist:
👍

tas50:
Tests are going to fail until we get a new chef gem out the door so we're going to have to merge this one with failing tests.

tas50:
Ping @btm

btm:
:+1:

PR #838, opened 2016-06-30
Remove support for Ruby 2.0

tas50:
Also test on more specific Ruby versions since the rbenv aliases don't
point to current.


thommay:
seems reasonable +1

PR #839, opened 2016-07-01
Add hardware plugin for ohai on darwin

natewalck:
This adds information about OS X that is basically unusable in the current system_profiler ohai plugin. I made it in such a way that more information could always be parsed out and used. 

Apple's formatting for the system_profiler info is really terrible, so even with XML output, it required quite a bit of massaging.

I'm with Facebook and this must be merged by @jaymzh 

P.S. I plan on doing rspec tests, but I wanted to make sure I was on the right track before writing them.


chef-supermarket:

* @natewalck

natewalck:
Revisions planned.....

lib/ohai/plugins/darwin/hardware.rb
```diff
@@ -0,0 +1,93 @@
...
+
+  collect_data(:darwin) do
+    begin
+      require "plist"
+    rescue Gem::LoadError
+      # In case the plist gem isn't present, skip this plugin.
+      next
+    end
+
+    hardware Mash.new
```
tas50: We shouldn't create this if hardware already exists. We check first around ohai so we don't globber other plugins that use the same namespace.

lib/ohai/plugins/darwin/hardware.rb
```diff
@@ -0,0 +1,93 @@
...
+#
+
+Ohai.plugin(:Hardware) do
+  provides 'hardware'
+
+  collect_data(:darwin) do
+    begin
+      require "plist"
+    rescue Gem::LoadError
+      # In case the plist gem isn't present, skip this plugin.
```
tas50: Some sort of debug logging stating we're skipping due to the gem not being there. Checkout out the high level virtualization plugin where we do the same thing with libvirt.

lib/ohai/plugins/darwin/hardware.rb
natewalck: Would it be better to make it a different namespace? I plan on adding similar data for Windows and Linux as well (to be used mainly for client machines).

tas50: hardware is a good high level namespace. The subkeys seem to be pretty clean here. If someone later on wanted to add something like HP iLO on the Linux side there'd be room and it wouldn't be horribly cluttered.

natewalck: I've looked in the code base and poked the Ohai object. What is the best way to look for an existing namespace? I assume I can return at the beginning of the collect_data function before anything happens.

lib/ohai/plugins/darwin/hardware.rb
```diff
@@ -36,6 +36,7 @@ def system_profiler(datatype)
       require "plist"
     rescue Gem::LoadError
       # In case the plist gem isn't present, skip this plugin.
+      Ohai::Log.debug("Plugin Virtualization: Can't load gem: #{e}. Cannot continue.")
```
tas50: Still says Virtualization plugin

tas50:
Works btw. Also I'm going to need a new battery soon


 "hardware": {
    "SMC_version_system": "2.19f12",
    "boot_rom_version": "MBP112.0138.B17",
    "cpu_type": "Intel Core i7",
    "current_processor_speed": "2.8 GHz",
    "l2_cache_core": "256 KB",
    "l3_cache": "6 MB",
    "machine_model": "MacBookPro11,3",
    "machine_name": "MacBook Pro",
    "number_processors": 4,
    "packages": 1,
    "physical_memory": "16 GB",
    "platform_UUID": "08CD3042-4DB0-550F-A936-0BB809D15B7E",
    "serial_number": "C02N63D0G3QP",
    "operating_system": "Mac OS X",
    "operating_system_version": "10.11.5",
    "build_version": "15F34",
    "architecture": "x86_64",
    "storage": [
      {
        "name": "Macintosh HD",
        "bsd_name": "disk1",
        "capacity": 499046809600,
        "drive_type": "ssd",
        "smart_status": "Verified",
        "partitions": 1
      }
    ],
    "battery": {
      "current_capacity": 6872,
      "max_capacity": 7728,
      "fully_charged": false,
      "is_charging": true,
      "charge_cycle_count": 474,
      "health": "Good",
      "serial": "D864315Y5LRF9CPAD",
      "remaining": 88
    }
  },

lib/ohai/plugins/darwin/hardware.rb
natewalck: Ooops. Fixed :)

natewalck:
I'm not seeing any other Ohai plugins that check the namespace before using it (unless I am missing the few that do). What is the best way to do this?

tas50:
This sort of thing https://github.com/chef/ohai/blob/80c91e74ac403959d54c9f15b1aa7a3c4fc4bd8b/lib/ohai/plugins/freebsd/network.rb#L24

natewalck:
Aha! I didn't realize all the Ohai data wasn't scoped in any way. I kept trying  ohai['cpu'], node['cpu'], etc. I knew CPU existed, but wasn't sure where on the namespace. I'll do this check for hardware now that I know how its scoped.

Thanks!

natewalck:
Ready for feedback now. Is this sufficient? Should I get started on the specs? 

I'm not sure why the CI is failing....it looked like a bundle install error. I'll check it out again after this run.

tas50:
I'll take a look at what's there now.  CI is failing due to rack 2.0.1 getting released. We're waiting on a new chef gem to fix that.

lib/ohai/plugins/darwin/hardware.rb
```diff
@@ -0,0 +1,102 @@
...
+  provides 'hardware'
+
+  def system_profiler(datatype)
+    sp_cmd = "system_profiler #{datatype} -xml"
+    # Hardware queries
+    sp_std = shell_out(sp_cmd)
+    if sp_std.exitstatus == 0
+      sp_hash = Plist.parse_xml(sp_std.stdout)
+    else
+      Ohai::Log.debug("system_profiler failed to run")
```
mcquin: As a note, the shell_out method has some debugging which may duplicate this message: https://github.com/chef/ohai/blob/master/lib/ohai/mixin/command.rb#L39. It's your decision.

lib/ohai/plugins/darwin/hardware.rb
```diff
@@ -0,0 +1,102 @@
...
+    end
+    sp_hash
+  end
+
+  collect_data(:darwin) do
+    unless hardware
+      hardware Mash.new
+    else
+      Ohai::Log.debug("Plugin Darwin Hardware: namespace already exists")
+      next
```
mcquin: It would be fantastic if there was a test to verify the plugin does not collect new "hardware" information.

lib/ohai/plugins/darwin/hardware.rb
```diff
@@ -0,0 +1,102 @@
...
+    end
+
+    begin
+      require "plist"
+    rescue Gem::LoadError
+      # In case the plist gem isn't present, skip this plugin.
+      Ohai::Log.debug("Plugin Hardware: Can't load gem: #{e}. Cannot continue.")
```
mcquin: `e` is not defined.

natewalck:
Rebased and changed a few things based on feedback.

mcquin:
It looks like you have some RuboCop/ChefStyle offenses. You can fix most of these automatically by running `bundle exec rubocop -a lib/ohai/plugins/darwin/hardware.rb` (leave off the `-a` to only detect offenses).

natewalck:
If this looks good now, I can start on the rspec. Let me know!

mcquin:
Go forth and write tests (though there's really no need to wait ;) ).

natewalck:
reminder, this must be merged by @jaymzh . I'll get on the rspecs.

natewalck:
There is a fail on the build, but it is a chefstyle error that I am uncertain on how to fix. when I add the 2 space indent, it causes the output blocks to get messed up ;(

mcquin:
@natewalck You don't have to indent the entire string block, just the lines beginning with `Hardware`, `Storage`, and `Power`. It'll look something like:

```ruby
module HardwareSystemProfilerOutput
  Hardware = <<hardware_output
<?xml version="1.0" encoding="UTF-8"?>
# ...
hardware_output

# ...
end
```

If you don't like the way that looks, which is a fine choice, you may want to explore copying the command outputs to `spec/data/plugins`. For example, if you saved hardware output to `spec/data/plugins/system_profiler_hardware.output`, you could replace `HardwareSystemProfilerOutput::Hardware` with `File.read(File.join(SPEC_PLUGIN_PATH, "system_profiler_hardware.output"))`

spec/unit/plugins/darwin/hardware_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+    )
+
+    allow(plugin).to receive(:shell_out).with(
+      "system_profiler SPPowerDataType -xml"
+    ).and_return(
+      mock_shell_out(0, HardwareSystemProfilerOutput::Power, "")
+    )
+  end
+
+  it "should parse hardware data correctly" do
```
mcquin: (Minor style comment) In alignment with [BetterSpecs](http://betterspecs.org/#should) we prefer to not use "should" in test descriptions.

natewalck:
I guess I don't understand why this is failing. It runs great on my own machine(OS X), but since it is failing on *nix, it seems like the shell_out mocks are not working quite as I expect them to.

I threw some pry binds in and I confirmed plugin contains the expected (mocked) data. Any idea where this could be breaking down? I basically copied the allow commands from system_profiler_spec, which isn't failing.

mcquin:
Ohai only runs the default and operating system-specific plugins. The tests pass on your machine because Ohai detects OS X (`:darwin`), but fail on Travis because it detects another operating system (`:linux`). 

You can trick Ohai into running this plugin during testing by adding this line to your `before(:each)` block:

```ruby
allow(plugin).to receive(:collect_os).and_return(:darwin)
```

natewalck:
That did it! Let me know if there are any other changes. If not, I'll squash and tag Phild.

mcquin:
This looks great. Squash and summon Phil.

natewalck:
@jaymzh, I choose you! (please don't kill me...)

natewalck:
@mcquin Thanks a ton for working through this diff with me! It was my first PR for ohai, which was a little different to test than providers, etc. Should be more complete for the initial PR next time.

jaymzh:
:+1:

jaymzh:
Sorry for the delay. Was in TX. :)

PR #843, opened 2016-07-05
TEST: Attempt to upgrade RVM in travis

tas50:
Upgrading gets us the latest version aliases so we can switch back to
things like 2.2 vs. 2.2.5


tas50:
This experiment showed we're going to have to maintain the exact versions we want to test on

PR #844, opened 2016-07-05
Update Ruby version + branches we build

tas50:
Also remove the builds of stable branches we don't use. We're never releasing 7 stable again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ sudo: false
 branches:
   only:
     - master
-    - 7.4-stable
 
 before_install:
   - gem update --system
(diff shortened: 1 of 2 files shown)


mcquin:
What prompted the decision to not build stable branches?

tas50:
The stable branch model doesn't seem like a thing we're doing anymore.  At the very least we're never going to build 7.4-stable again.

mcquin:
Sounds fair. Can we give this a name that also reflects the change to remove build versions?

tas50:
@mcquin done

mcquin:
Thanks! 👍

PR #845, opened 2016-07-05
Build on the trusty travis image

tas50:
Newer build image with everything that comes along with 2 years of Linux
development


tas50:
Turns out we'd need sudo for this to work

PR #848, opened 2016-07-17
Properly detect PHP 7

tas50:
PHP 7 strings have changed and our old regex broke as the build date is no longer included. Update the regex to make the build data optional and add tests for the new strings


thommay:
:+1: LGTM

mcquin:
:+1:

PR #852, opened 2016-07-29
Avoid global mutation.

### LINKED ISSUE #851: Inadvertent global mixin
Looking at how things are using `FileMixin` I ran across https://github.com/chef/ohai/blob/80c91e74ac403959d54c9f15b1aa7a3c4fc4bd8b/lib/ohai/plugins/linux/sessions.rb#L21

I think this line is in error, it is mixing the helper in globally (i.e. in to `Object`). I think the intention was to do it inside the plugin body but when things were moved from the old DSL to the new one, it got mixed up. This should be removed ASAP because dumping things into the global namespace is terrifying, but doing this backwards compat might be tricky. My vote would be to add that mixin the `Plugin` base class so any existing plugins will continue to work.

coderanger:
Closes #851.

### DIFF
--- a/lib/ohai/dsl/plugin.rb
+++ b/lib/ohai/dsl/plugin.rb
@@ -82,9 +82,6 @@ module Ohai
       include Ohai::Mixin::OS
       include Ohai::Mixin::Command
       include Ohai::Mixin::SecondsToHuman
-
-      # Pending next major release of Ohai.
-      # See https://github.com/chef/ohai/issues/851 for details.
       include Ohai::Util::FileHelper
 
       attr_reader :data
--- a/lib/ohai/dsl/plugin.rb
+++ b/lib/ohai/dsl/plugin.rb
@@ -22,6 +22,7 @@ require "ohai/mixin/os"
 require "ohai/mixin/command"
 require "ohai/mixin/seconds_to_human"
 require "ohai/hints"
+require "ohai/util/file_helper"
 
 module Ohai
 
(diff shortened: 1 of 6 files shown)


lib/ohai/dsl/plugin.rb
```diff
@@ -82,6 +83,10 @@ class Plugin
       include Ohai::Mixin::Command
       include Ohai::Mixin::SecondsToHuman
 
+      # Pending next major release of Ohai.
```
tas50: Removal pending? The comment is a bit vague. Also when we leave comments in code that say next major version no action is ever taken. Pick the version where we plan to deprecate it.

coderanger: We should probably decide first if we want it to be removed at all or if FileHelper should just stay as part of the Plugin base API. That would seem fine too :)

tas50: I'd be fine if we kept this part of the Plugin base API. It's small and useful for writing plugins.

coderanger: Roger, updating PR :)

coderanger:
The DmiDecode change _might_ break something but it seems much less likely someone was accidentally using those methods :) Will see what the tests say.

tas50:
I highly doubt anyone is using the DMI helper and it's the right thing to do. Lets just fix that.

tas50:
+1

PR #856, opened 2016-07-30
Add a plugin for collecting available shells

tas50:
The exact shell paths vary by OS so it would be really nice to have a list of the shells so you could find the path to bash by walking the array.


mcquin:
👍

PR #859, opened 2016-08-05
fixing solaris2 ohai plugin regex

acaiafa:
The current regex in place does not properly catch the interfaces. Updating this regex allows for the network information to actually populate.


scotthain:
👍 verified. Thanks @johnbellone

acaiafa:
@scotthain any idea how long until this makes it out?

scotthain:
@acaiafa I don't actually know, I'll see if I can find out.

tas50:
Not the next chef release, but the release after that

tas50:
Looks good to me. 👍

PR #860, opened 2016-08-10
Add platform detection for Arista EOS

jerearista:
Arista EOS can be detected based on the /etc/Eos-release file which includes the EOS version string.

### DIFF
--- a/lib/ohai/plugins/linux/platform.rb
+++ b/lib/ohai/plugins/linux/platform.rb
@@ -153,6 +153,10 @@ Ohai.plugin(:Platform) do
     elsif File.exist?("/etc/alpine-release")
       platform "alpine"
       platform_version File.read("/etc/alpine-release").strip()
+    elsif File.exist?("/etc/Eos-release")
+      platform "arista_eos"
+      platform_version File.read("/etc/Eos-release").strip.split[-1]
+      platform_family "fedora"
     elsif os_release_file_is_cisco?
       raise "unknown Cisco /etc/os-release or /etc/cisco-release ID_LIKE field" if
         os_release_info["ID_LIKE"].nil? || ! os_release_info["ID_LIKE"].include?("wrlinux")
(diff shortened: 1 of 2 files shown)


jerearista:
@tas50 This is the PR we discussed via email earlier today.

tas50:
Thanks for getting this filed @jerearista. The tests are failing due to issues with rack 2.0 and the new rspec release. I have a PR to get that fixed. Once we get it merged in you should be able to rebase off master to get your PR to pass.

I do have some reservations about making the platform_family fedora in this PR.  I know that EOS is based on Fedora, but the Arista install may differ enough from Fedora that having the platform_family fedora will become a disservice to anyone attempting to make generalization about "fedora-ness" of the setup. Pinging @thommay and @lamont-granquist for their thoughts on this since we've discussed greater platform_family issues a lot lately.

lamont-granquist:
IDK what fedora looks like now after systemd, but if there's /etc/sysconfig that looks like the rest of the redhat systems, and it has "*-devel" RPMs, and things like that, then its 'fedora-derived' system.  Since its a switch O/S I'm sure they messed around with networking stuff in important ways -- maybe hostname, ipaddress, etc are no longer set the same was a fedora -- but it doesn't matter since that is inherently going to have to switch off `node[:platform] == "arista_eos"` anyway.

If `node[:platform_family] == 'fedora'` gets *anything* correct about the system config, then its probably the right thing to do -- the /differences/ are the wrong thing to look at since platform_family is inherently fuzzy-matching.

tas50:
👍 from me obviously if lamont is ok with the platform family.  Test failure was due to rack coming in with github changelog generator. That's been fixed in another PR so this will run fine in master now.

jerearista:
Thanks.  I'll keep an eye on the fedora setting.

lamont-granquist:
Yeah that was a 👍  -- although I'm kind of blindly guessing what EOS looks like...

PR #863, opened 2016-08-12
only configure logging if we must

thommay:


thommay:
This should probably move to mixlib-log, but it's the minimal fix to the disappearing logging in chef/mixlib-log#17

lamont-granquist:
👍

PR #866, opened 2016-08-15
Require at least mixlib-log 1.7.1

tas50:
Our use of the configured method now fails if we're not on 1.7.1.


mwrock:
:+1:

jkeiser:
:+1:

PR #867, opened 2016-08-15
add a plugin for machine information

davide125:
Add a new plugin to expose machine information. On systemd hosts, this uses hostnamectl if available, otherwise it parses directly /etc/machine-id and /etc/machine-info. It falls back to /var/lib/dbus/machine-id if necessary, which is the old location from when this was managed by dbus directly.

I'm a Facebook employee and @jaymzh will do the final merge.


chef-supermarket:

* @davide125

davide125:
cc @chef/client-core

jaymzh:
CC @chef/client-core

jaymzh:
:+1:

jaymzh:
Oh @chef/client-core ? @mcquin ?

thommay:
it feels weird that this is a separate hash of information that duplicates things we already know from other plugins

jaymzh:
is there another plugin we should add this to?

davide125:
@thommay: happy to move this somewhere else if you prefer, but I don't think this is duplicate information. There's nothing in ohai parsing /etc/machine-id and nothing else that provides this. About hostnamectl, it is true that some of the data coming out of it (e.g. the kernel version) is also reported in other places, but it seemed cleaner to just add everything to the hash versus picking and choosing the fields to use (as that'd be a pain to maintain in the long run).

davide125:
I see this akin to the LSB plugin, that provides extra information in `node['lsb']` that might duplicate some of the platform stuff, but is fundamentally separate.

btm:
What data are you trying to use from this output?

hostname and/or machinename are going to be the same as machine[:hostname]
kernel[:machine] seems like it'll be fundamentally the same as machine[:architecture]
kernel[:name] + kernel[:release] looks the same as machine[:kernel]
lsb[:description] looks like machine[:operating_system] 
virtualization[:role] could be similar to machine[:chassis] in purpose, but I can see unique uses.

We should avoid duplicating most of that information unless there are special cases where it is different, in which case we should document how it is different than other data. Continuing to follow the pattern of creating top level branches based on the tool the data comes from will lead to people using data from a part of the tree that isn't cross-platform. Hostname is an extreme example of data we want people to only get from a single location in the data set so we can focus on populating that for multiple platforms.

I suspect the data you're looking for would be "overloading" kernel if we put it there. I like the idea of a top level "machine" or "chassis" or "hardware" or "host" branch. I'm curious about 'Machine ID' vs 'Boot ID' vs 'BIOS UUID' (from dmi). I'd appreciate a brief comment on their usefulness if those are indeed data you're looking for.

davide125:
I wrote this specifically because I needed access to `machine_id` while writing a cookbook to manage `systemd-boot` (which uses the machine id to reference boot entries). This is a stable UUID that's specific to the system (it changes from box to box). dbus uses it as well to identify the machine and it's referenced in several configs. Boot ID is a UUID that identifies a given boot (i.e. it changes every time the box reboots), and is useful to correlate logs. Both are completely unrelated to the DMI UUID. Other unique fields this provides are `pretty_hostname`, `icon_name`, `chassis`, `deployment`, which are all admin-set via `hostnamectl` or `/etc/machine-info`; of these, I can see `chassis` and `deployment` being useful to have access to from Chef. Also note that `chassis` has nothing to do with virtualization, it's used to bucket what kind of machine you're running on (server, laptop, tablet, etc.).

lamont-granquist:
it feels to me like dumping all the output from hostnamectl into ohai should go into `node['hostnamectl']` and that utility should just own that entire namespace and nothing is there if we're not running systemd.  then `node['machine_id`]` could be set by multiple different sources (from hostnamectl or dbus or whatever).  i don't like having part of the schema determined entirely by hostnamectl, but then overriding bits and pieces of it to try to emulate that information if its not there.

davide125:
Sounds good @lamont-granquist, I'll rework this to use `node['hostnamectl']` and populate `node['machine_id']`.

davide125:
@lamont-granquist I coulnd't find a way to make the machineid plugin conditionally depend on hostnamectl and use its provided result; either I make it `depends`, and then it hard fails if the attribute is missing, or I don't, and then the attribute value is not available. So for now I just made machineid read from the files, under the assumption that if the box is running systemd /etc/machine-id will always be populated anyway.

lamont-granquist:
that looks fine.  i also wouldn't mind if machine_id also called hostnamectl to only extract what it wants, the perf hit from that is miniscule compared to everything else.  either way though 👍

thommay:
👍 feel free to merge, @jaymzh . And thank you for dealing with my rather drive-by NACKing, @davide125

spec/unit/plugins/linux/machineid_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    allow(File).to receive(:read).with("/etc/machine-id").and_return(machine_id)
+    plugin.run
+    expect(plugin[:machine_id]).to eq(machine_id)
+  end
+
+  it "should read /var/lib/dbus/machine-id if available" do
+    machine_id = "6f702523e2fc7499eb1dc68e5314dacf"
+
+    allow(File).to receive(:exists?).with("/etc/machine-id").and_return(false)
+    allow(File).to receive(:exists?).with("/var/lib/dbus/machine-id").and_return(true)
```
thommay: You need to mock the `read` here.

jaymzh:
:+1:

PR #869, opened 2016-08-16
Solaris_update_version: Create kernel[:update] for solaris 2

MarkGibbons:
Extract the Solaris version update information from /etc/release.

The solaris version update information is not reported by Ohai.  Chef is making decisions about chef-client support based on the version update of the solaris os.   Reporting the os information might make it simpler to find older servers that need to be upgraded.

### DIFF
--- a/spec/unit/plugins/solaris2/kernel_spec.rb
+++ b/spec/unit/plugins/solaris2/kernel_spec.rb
@@ -139,7 +139,30 @@ describe Ohai::System, "Solaris2.X kernel plugin" do
     allow(@plugin).to receive(:init_kernel).and_return({})
     allow(@plugin).to receive(:shell_out).with("uname -s").and_return(mock_shell_out(0, "SunOS\n", ""))
     allow(@plugin).to receive(:shell_out).with("modinfo").and_return(mock_shell_out(0, MODINFO, ""))
+    @release = StringIO.new("                   Oracle Solaris 10 1/13 s10s_u11wos_24a SPARC\n Assembled 17 January 2013")
+    allow(File).to receive(:open).with("/etc/release").and_yield(@release)
   end
+
+  it "should give the Solaris update version informtion" do
+    @release = StringIO.new("                      Solaris 10 10/08 s10s_u6wos_07b SPARC\n Use is subject to license terms.\n Assembled 27 October 2008")
+    allow(File).to receive(:open).with("/etc/release").and_yield(@release)
+    @plugin.run
+    expect(@plugin[:kernel][:update]).to eq('10 10/08 s10s_u6wos_07')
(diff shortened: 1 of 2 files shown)


spec/unit/plugins/solaris2/kernel_spec.rb
```diff
@@ -139,6 +139,29 @@
...
+  end
+
+  it "should give the Solaris update version informtion" do
+    @release = StringIO.new("                      Solaris 10 10/08 s10s_u6wos_07b SPARC\n Use is subject to license terms.\n Assembled 27 October 2008")
+    allow(File).to receive(:open).with("/etc/release").and_yield(@release)
+    @plugin.run
+    expect(@plugin[:kernel][:update]).to eq("10 10/08 s10s_u6wos_07")
+  end
+
+  it "should give the Oracle Solaris update version informtion" do
```
tas50: Typo in the name of the specs. The others have it too

tas50:
Fix the spec name typo, but 👍  from me

scotthain:
I'm not 100% sure this will cover the use cases... If I recall correctly, if you install a single patchset update on Solaris 10 it will append something like 'update X applied' to the end, and keep the original installation's version at the top, but if you apply *multiple* patchsets, it will reflect the 'latest' that was installed at the top of the file with no 'update applied' line.

Am I remembering this correctly @MarkGibbons @benr ?

tas50:
@scotthain Do we have a system we could test this on?

scotthain:
@tas50 we have many, but all of ours are patched up, so we don't really have an easy way to test the upgrade path here. A good test case would be to grab a Sol10u8 or so ISO, put it in virtualbox, do the patchset upgrade(s) and test at each stage.

That being said, that's a huge pain in the ass for everyone involved.
@MarkGibbons in your opinion, based on your fleet and your industry knowledge, do you think the 'single patch' use case is common at all?

I didn't mean to make a big deal out of a corner case, just wanted to make it visible. I don't think we  want to encourage behavior that leads to less patching. /shrug

MarkGibbons:
I don't think the patch level updates are an issue.  This snip of the contents of /etc/release is from a server that gets patched every 6 months.  As far as I know, on solaris 10  SUNWsolnm (/etc/release package) doesn't get updated by patching. The package only gets replaced when an update version to the OS is created and shipped.


                   _Oracle Solaris 10 1/13 s10s_u11wos_24a SPARC
  Copyright (c) 1983, 2013, Oracle and/or its affiliates. All rights reserved.
                            Assembled 17 January 2013_

tas50:
Sounds like we're good to go on this then 👍

thommay:
👍

PR #870, opened 2016-08-19
chefstyle 0.4.0

lamont-granquist:
fixing all the new enforcement with chefstyle 0.4.0

### DIFF
--- a/lib/ohai/mixin/command.rb
+++ b/lib/ohai/mixin/command.rb
@@ -177,7 +177,7 @@ module Ohai
           $VERBOSE = nil
           ps.last.fcntl(Fcntl::F_SETFD, Fcntl::FD_CLOEXEC)
 
-          cid = fork {
+          cid = fork do
             Process.setsid
 
             pw.last.close
@@ -226,7 +226,7 @@ module Ohai
             end
             ps.last.close unless ps.last.closed?
             exit!
-          }
+          end
         ensure
           $VERBOSE = verbose
         end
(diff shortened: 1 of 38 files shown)


lib/ohai/mixin/command.rb
```diff
@@ -177,7 +177,7 @@ def popen4(cmd, args = {}, &b)
           $VERBOSE = nil
           ps.last.fcntl(Fcntl::F_SETFD, Fcntl::FD_CLOEXEC)
 
-          cid = fork {
```
tas50: might as well fix it, but this is hopefully getting deleted soon

lib/ohai/plugins/windows/cpu.rb
```diff
@@ -54,9 +54,11 @@
       cpu[current_cpu]["vendor_id"] = processor["manufacturer"]
       cpu[current_cpu]["family"] = processor["family"].to_s
       cpu[current_cpu]["model"] = processor["revision"].to_s
-      cpu[current_cpu]["stepping"] = processor["stepping"].nil? \
-                  ? processor["description"].match(/Stepping\s+(\d+)/)[1] \
-                  : processor["stepping"]
+      cpu[current_cpu]["stepping"] = if processor["stepping"].nil?
```
tas50: Killing these multiline operators is fantastic. They're a real pain to read

lamont-granquist: you didn't see the one in chef/chef where someone used a multiline ternary and broke a string in the middle with \ in order to line wrap it and get it under 80 columns.  completely unreadable omgwtfbbq.

this example is just from someone who didn't understand whitespace parsing and tokenization in ruby...

lib/ohai/mixin/command.rb
lamont-granquist: rubocop fixed it for me with `-a`

lamont-granquist:
this is actually passing but chefstyle is bitching about Lint/UnnededDisable which is a rule that i'll disable for chefstyle 0.4.0 to make this annoyance stop.

PR #876, opened 2016-09-08
Ohai uptime plugin hangs in Windows.

Aliasgar16:
Ohai uptime plugin hangs in Windows, solution:

1. Changed WMI class to `Win32_OperatingSystem` for newer versions of Windows.
2. Kept the old WMI class `Win32_PerfFormattedData_PerfOS_System` for older versions of Windows.

WMI class is from where we retrieve the system's uptime for Windows platform.

### DIFF
--- a/spec/unit/plugins/windows/uptime_spec.rb
+++ b/spec/unit/plugins/windows/uptime_spec.rb
@@ -1,0 +1,88 @@
+#
+# Author:: Aliasgar Batterywala (<aliasgar.batterywala@msystechnologies.com>)
+# Copyright:: Copyright (c) 2016 Chef Software, Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


spec/unit/plugins/windows/uptime_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+describe Ohai::System, "Windows plugin uptime" do
+
+  let(:wmi) { double("wmi", { :first_of => "" }) }
+
+  before(:each) do
+    allow(WmiLite::Wmi).to receive(:new).and_return(wmi)
+  end
+
+  ## Windows newer versions category here includes server OS starting from Windows Server 2008 ##
+  shared_context "WMI class to use to fetch system uptime for newer versions of Windows platform" do
```
btm: can you make the names of the shared context's more brief? that's a lot of words with only one in the middle that is different. It should be easier to tell the difference just looking at them.

Aliasgar16:
Done.

btm:
:+1: 

@mwrock , @tas50 review?

mwrock:
:+1:

tas50:
@Aliasgar16 before I merge this can you just toss a sentence or two into the PR description explaining why we'd want to do this?

Aliasgar16:
@tas50 I have modified the description of the PR. Let me know your feedback.

tas50:
Thanks @Aliasgar16

PR #881, opened 2016-09-22
Doc changes for #877

coderanger:
Can someone double check that that's the right config syntax? Haven't played much with configurable plugins.

### DIFF
--- a/DOC_CHANGES.md
+++ b/DOC_CHANGES.md
@@ -17,10 +17,16 @@ means that a node with the same source inputs will always result in the same
 shard seed, but collisions are possible.
 
 The sources used to compute the seed value can be configured either in your
-Chef config file or your Ohai config file ([ed: does anyone use Ohai config files?]):
+Chef config file:
 
 ```ruby
 Ohai::Config[:plugins][:shard_seed][:sources] = [:fqdn]
+```
+
+or your Ohai config file if you run Ohai outside of Chef:
+
+```ruby
+ohai.plugin[:shard_seed][:sources] = [:fqdn]
 ```
 
 The default sources are `[:machinename, :serial, :uuid]` but this value should
--- a/DOC_CHANGES.md
+++ b/DOC_CHANGES.md
@@ -20,7 +20,7 @@ The sources used to compute the seed value can be configured either in your
 Chef config file:
 
 ```ruby
-Ohai::Config[:plugins][:shard_seed][:sources] = [:fqdn]
+Ohai::Config[:plugin][:shard_seed][:sources] = [:fqdn]
 ```
 
 or your Ohai config file if you run Ohai outside of Chef:
--- a/DOC_CHANGES.md
+++ b/DOC_CHANGES.md
@@ -7,3 +7,69 @@ Description of the required change.
 -->
 
 # Ohai Doc Changes:
+
+## Shard Seed Plugin
+
+The new `shard_seed` attribute provides a normally distributed integer ID value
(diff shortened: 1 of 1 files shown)


DOC_CHANGES.md
```diff
@@ -7,3 +7,69 @@ Description of the required change.
...
+## Shard Seed Plugin
+
+The new `shard_seed` attribute provides a normally distributed integer ID value
+to all nodes. This value will be between 0 and 4294967295 (0xFFFFFFFF), based
+on a consistent hash of one or more pieces of node data. A consistent hash
+means that a node with the same source inputs will always result in the same
+shard seed, but collisions are possible.
+
+The sources used to compute the seed value can be configured either in your
+Chef config file or your Ohai config file ([ed: does anyone use Ohai config files?]):
```
jaymzh: you have to whenever you are developing Ohai. We need to make this better though.

DOC_CHANGES.md
```diff
@@ -7,3 +7,69 @@ Description of the required change.
...
+
+```ruby
+Ohai::Config[:plugins][:shard][:sources] = [:fqdn]
```
jaymzh: s/shard/shard_seed/

DOC_CHANGES.md
```diff
@@ -7,3 +7,69 @@ Description of the required change.
 -->
 
 # Ohai Doc Changes:
+
+## Shard Seed Plugin
+
+The new `shard_seed` attribute provides a normally distributed integer ID value
```
MarkGibbons: Hopefully it's not an approximately normal distribution.  I think md5 would be close to a uniform distribution though.
http://michiel.buddingh.eu/distribution-of-hash-values

DOC_CHANGES.md
coderanger: I thought the whole point of splitting `chef-config` into its own gem was to allow stuff like Ohai to read Chef config files directly.

jaymzh: Oh that'd be neat. But now, run `ohai -c /etc/chef/client.rb` and watch it fail.

coderanger: Think it's worth mentioning ohai config files at all for normal users though? Probably a red herring for most people that are smart and don't touch ohai core :)

jaymzh: Believe or not I know several people who use Ohai outside of Chef, so yes.

DOC_CHANGES.md
jaymzh: Also isn't those Ohai.config now?

jaymzh: (or Ohai.plugin for fun)

jaymzh: apparently it's "ohai.plugin" according to the docs. /me/sighs.

DOC_CHANGES.md
```diff
@@ -7,3 +7,75 @@ Description of the required change.
...
+
+Two main use cases for the seed value are slow rollouts of new features and
+computing service IDs.
+
+To do slow rollouts, you can wrap new features or options in your recipe code
+with a conditional `if` statement like:
+
+```ruby
+# Enable feature for 20% of nodes.
+if (node['shard_seed'] % 100) < 20
```
jaymzh: fun side note: technically because we start at 0, you're this is 21% :)

tas50:
@coderanger and @jaymzh Is this ready for merge?

coderanger:
As long as the config file bit looks correct to people that know the Ohai config APIs, yes.

Issue #833, opened 2016-06-17
plugins/linux/virtualization: if lxc and docker are on the same host, it can't detect lxc


goacid:
## Description

Phys Host : ubuntu 16.04 LTS with docker engine 1.11 and lxc 2.0.1 installed.
LXC containers are ubuntu 14.04 with chef (from 11 to latest),which embeded ohai, always the same problem virtualization.rb doesn't detect the container has a guest.
With native ohai same problem.
On same environment, without docker installed in the host, it works fine.
## Ohai Version

From 7.0.4 to latest (maybe before)
## Platform Version

Host : Ubuntu 16.04
Lxc container : Ubuntu 14.04
## Ohai Output

An LXC guest return : 

```
"virtualization": {
    "systems": {

    }
```

On an LXC container running on a physical host without docker : 

```
  "virtualization": {
    "system": "lxc",
    "role": "guest"
  },
```

Content of cgroup inside the ubuntu 14.04 containers :

```
cat /proc/1/cgroup 
9:cpu,cpuacct:/
8:pids:/
7:devices:/
6:cpuset:/
5:blkio:/
4:net_cls,net_prio:/
3:freezer:/
2:memory:/
1:name=systemd:/
root@crawler01:~# cat /proc/self/cgroup 
9:cpu,cpuacct:/
8:pids:/
7:devices:/
6:cpuset:/
5:blkio:/
4:net_cls,net_prio:/
3:freezer:/
2:memory:/
1:name=systemd:/
```

The problem seems here : 
file virtualization.rb line 180

`if File.read("/proc/self/cgroup") =~ %r{^\d+:[^:]+:/(lxc|docker)/.+$}`

I'm currently searching a way to find any difference to be able to know if a container is lxc or docker from inside the container.

huckley:
cat /proc/1/environ say what?

goacid:
HOME=/TERM=linuxBOOT_IMAGE=/bzImage-3.14.32-XXXX

lehn-etracker:
See my patch here https://github.com/chef/ohai/compare/master...lehn-etracker:lxc-2.0?expand=1
I will send a pull request after signing Chef CLA!

goacid:
Thanks  !

jeunito:
This was an issue for me too. I have made another PR with tests passing at https://github.com/chef/ohai/pull/1055

tas50:
This should be fixed on master now with the updated PR from @jeunito

PR #883, opened 2016-09-27
Add LXC-2.0 Support. Detect lxc-host and lxc-containers via /proc/1/e…

lehn-etracker:
…nviron

Fixes https://github.com/chef/ohai/issues/833

### DIFF
--- a/lib/ohai/plugins/linux/virtualization.rb
+++ b/lib/ohai/plugins/linux/virtualization.rb
@@ -26,6 +26,10 @@ Ohai.plugin(:Virtualization) do
     which("lxc-version")
   end
 
+  def lxc_ls_exists?
+    which("lxc-ls")
+  end
+
   def docker_exists?
     which("docker")
   end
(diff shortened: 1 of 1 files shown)


lehn-etracker:
Okay. It seems I have to fix the spec tests.

thommay:
@lehn-etracker hey, did you have any luck updating those tests?

lehn-etracker:
@thommay not yet. I had not enough time. Maybe you are able to assist me? Just give me a pull request with tests and i will send another one to ohai.

lib/ohai/plugins/linux/virtualization.rb
```diff
@@ -188,6 +216,18 @@ def nova_exists?
         # But, this could overwrite virtualization[:systems][:lxc] = "guest"
         # If so, we may need to look further for a differentiator (OHAI-573)
         virtualization[:systems][:lxc] = "host"
+      elsif init_environ_content =~ %r{\x00?container=lxc\x00?}
```
stefanor: Shouldn't that be: `%r{(^|\x00)container=lxc(\x00|$)}`? Otherwise the nulls aren't really doing anything useful.

lehn-etracker: @stefanor Possibly yes, but i've never seen this parameter at the beginning or the end of /proc/1/environ

stefanor: Ordering wouldn't be guaranteed. And the `\x00?` is really not useful. Any character can appear before / after `container=lxc` and the regex will match.

So, either remove `\x00?` or match the boundaries properly.

ccope: I was just starting to implement this myself because I realized lxc detection stopped working when I upgraded to 16.04 hosts. Here was my solution:
```
init_environ = File.read('/proc/1/environ').split("\0".b).map { |x| x.split('=', 2) }.to_h
if init_environ['container']
  ...
```
This splits the data by the null bytes, and then splits each environment variable statement into keys and values in a hash. Might be overkill if LXC is the only thing setting useful info in the init environment though.

ccope:
Also, I use nested containers which are both guest and host. I usually care more about whether an environment is a guest than a host, so if we don't want to include both, I'd prefer guest to win.

damm:
Hmph; seems like this is abandoned then?

aalmenar:
Some systems due to using lxc 2.0 get identified in ohai as:

kvm host, which is totally wrong. If at least was detected as guest, then it should not be a greater issue.

:D

tas50:
Closing this in favor of #1055

PR #885, opened 2016-10-03
Rename to info_getter

tas50:
Ohai I'm just here renaming your app. Ohai seems so confusing. We should just rename it.

BTW: This is entirely a joke. I just want to see if the DCO bot works


mcquin:
I was worried.

PR #887, opened 2016-10-14
Adjusting redhat/fedora/suse/pld packages output

jrmuir:
For some packages such as the kernel there can be multiple installed versions.  

This will load packages into array format to capture duplicate version/release/installdate information into additional array/hash output.

Signed-off-by: John Muir jrmuir@hotmail.com


tas50:
@jrmuir I have a few questions regarding this.  In our current setup what happens with the duplicate packages? Do we just take the last one added?  Also can you add testing for this. It seems like something we want to make sure we don't regress later and we should explain exactly what we expect to happen via a set of tests.

lib/ohai/plugins/packages.rb
```diff
@@ -47,7 +47,12 @@
 
       pkgs.each do |pkg|
         name, epoch, version, release, installdate, arch = pkg.split
-        packages[name] = { "epoch" => epoch, "version" => version, "release" => release, "installdate" => installdate, "arch" => arch }
+        unless packages[name].nil?
```
tas50: It might be worth noting why this would be nil here in a comment

jrmuir: Ah, that was to help make the determination if I'm dealing with a duplicate or not - only a duplicate would trigger adding additional items for that software package.  I'll add comments there, and I'm definitely open to suggestion on the logic there if there's a better way to handle a duplicate situation.

With RPM, it seems that we only get the last listed duplicate.  I'm not sure if there's a good way to determine what that's going to be on any given system as it's completely dependent on rpm's output ordering and the exist logic just overwrites previous entries as it runs through the list.  

I'll look at what I can do to include valid tests as you're right and we don't want to regress.  Other than kernel related packages, I doubt many others can actually have multiple copies at any given time unless perhaps talking source rpms which aren't likely to exist on a chef-managed server though I think this would catch that..

jrmuir:
Existing output from a system with multiple kernels via knife show -l
  kernel:
    arch:        x86_64
    epoch:       0
    installdate: 1476801222
    release:     327.28.3.el7
    version:     3.10.0

New output:
  kernel:
    arch:
      x86_64
      x86_64
      x86_64
    epoch:
      0
      0
      0
    installdate:
      1460741256
      1457983178
      1476801222
    release:
      327.13.1.el7
      121.el7
      327.28.3.el7
    version:
      3.10.0
      3.10.0
      3.10.0

Doing a little extra [investigation](http://serverfault.com/questions/601427/sort-installed-packages-version), it looks like with the formatting options in place we should always get the newest last, though the others appear to not sort in order which is probably fine.  Assuming that's correct the existing packages plugin should always come up with the newest which was a concern I had when starting on this.

lamont-granquist:
Huge :-1: on having keys that change between arrays and scalars.

Also huge :-1: on this crazy way of having a hash table of arrays rather than an array of hashes.

We could change this as a major breaking change in Chef 13 so that every package got an array of hash tables.

Honestly I think the code that consumes this that checks for multiple installed kernel versions should simply not use ohai attributes at all, and if you need custom reporting on multiple kernel versions it should go into custom plugins and custom attributes and not here -- as long as the most recent kernel version is correct this code is correct and it should not be made more complicated and we should not accept this responsibility in this plugin.

thommay:
I totally understand the use case, and thank you for the PR. But  this represents an API change, and I'm not sure it's the correct approach.
It would seem better to collect arrays of the hash table, but as this stands it makes everyone's code more fragile, since you have to check if you're getting an array or a string and behave differently.
So we'll be closing this PR; I'd recommend writing a site specific plugin for this.
Thanks again, and I'm sorry we can't take this.

lamont-granquist:
The big show-stopper here is that the usual code that walks through array-or-scalar values doesn't work ( `[ might-be-array-or-scalar ].flatten` ) because this constructs a hash-of-arrays instead of an array-of-hashes.  I still don't like the array-of-hashes approach either.

jrmuir:
I wasn't sure what else might be impacted on this, but did see the missing information, and life took over for a bit.  I know this is now closed but what if there was just a boolean flag (or incremented count?) for the existence of duplicates without additional data?  This could potentially be an indicator to check for without causing headaches elsewhere.

PR #890, opened 2016-10-25
ip_scopes update

n-marton:
This modification prevents ip_scopes plugin to return 'tunl' or 'docker' type interfaces' ipaddress as privateaddress.

### DIFF
--- a/lib/ohai/plugins/ip_scopes.rb
+++ b/lib/ohai/plugins/ip_scopes.rb
@@ -32,7 +32,7 @@ Ohai.plugin(:IpScopes) do
           begin
             attrs["ip_scope"] = address.to_ip.scope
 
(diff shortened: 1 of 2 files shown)


tas50:
Seems pretty reasonable. +1

lib/ohai/plugins/ip_scopes.rb
```diff
@@ -51,7 +51,15 @@ def private_addr?(address)
     address.to_ip.scope =~ /PRIVATE/
   end
 
-  def tunnel_iface?(interface)
+  def ppp_iface?(interface)
     interface["type"] == "ppp"
   end
+
+  def tunel_iface?(interface)
```
btm: why the change from tunnel to tunel? is this a typo in multiple places?

n-marton:
fixed tunnel vs tunel stuff

PR #892, opened 2016-10-28
Improve OS X Ohai support

benr:


lib/ohai/plugins/darwin/passwd.rb
```diff
@@ -0,0 +1,9 @@
+Ohai.plugin(:Passwd) do
```
thommay: Needs a license/copyright header.

lib/ohai/plugins/darwin/passwd.rb
```diff
@@ -0,0 +1,9 @@
...
+  provides "loggedin_user"
+
+  collect_data(:darwin) do
+    unless loggedin_user
+      loggedin_user shell_out("stat -f \'\%Su\' /dev/console").stdout.chop
```
tas50: We have a few questions here from our review:

1) Can more than 1 user be logged in at a time?
2) Can we expose log in type? What if the user is logged in remotely?

tas50:
Closing this out due to lack of response. Feel free to open it back up again with testing

PR #893, opened 2016-10-31
Add sysconf plugin to expose system configuration variables

davide125:
Add a new plugin to expose system configuration variables. See `man sysconf` and `man getconf` for what this is about, the tl;dr is that it exposes a number of system-level variables that are useful to understand what kind of machine or system we're running on. For example, the `LONG_BIT` variable is the only reliable way to determine whether we're running a 32bit userspace on top of a 64bit kernel. This interface is defined by POSIX and should be available on all UNIX-like OS. I tested this on Linux, but verified the interface is documented as available on AIX, Solaris, Darwin, BSD, IRIX and HP-UX.

### DIFF
--- a/lib/ohai/plugins/sysconf.rb
+++ b/lib/ohai/plugins/sysconf.rb
@@ -1,0 +1,45 @@
+#
+# Author:: Davide Cavalca <dcavalca@fb.com>
+# Copyright:: Copyright (c) 2016 Facebook
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+Ohai.plugin(:Sysconf) do
+  provides "sysconf"
+
+  collect_data(:aix, :linux, :solaris2) do
+    getconf_path = which("getconf")
+    if getconf_path
+      getconf = shell_out("#{getconf_path} -a")
+
+      if getconf.exitstatus == 0
+        sysconf Mash.new unless sysconf
+
+        getconf.stdout.split("\n").each do |line|
+          key, val = /^(\S+)\s*(.*)?$/.match(line).captures
+          if val && !val.empty?
+            begin
+              sysconf[key] = Integer(val)
+            rescue
+              sysconf[key] = val
+            end
+          else
+            sysconf[key] = nil
+          end
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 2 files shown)


jaymzh:
CC @chef/client-core

jaymzh:
Ping. @thommay  and @lamont-granquist

lib/ohai/plugins/sysconf.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  collect_data do
+    getconf_path = which("getconf")
+    if getconf_path
+      getconf = shell_out("#{getconf_path} -a")
+
+      sysconf Mash.new unless sysconf
+
+      getconf.stdout.split("\n").each do |line|
+        key, val = line.split
```
lamont-granquist: ah, you probably need something more like:

```
key, val = /^(\S+)\s+(.*)$/.match(line).captures
```

lib/ohai/plugins/sysconf.rb
```diff
@@ -0,0 +1,43 @@
...
+
+Ohai.plugin(:Sysconf) do
+  provides "sysconf"
+
+  collect_data do
+    getconf_path = which("getconf")
+    if getconf_path
+      getconf = shell_out("#{getconf_path} -a")
+
+      sysconf Mash.new unless sysconf
```
btm: wrap from here down in `if getconf.exitstatus == 0` so we don't create `sysconf` or bother parsing it if we have a getconf that does't do what we expect, like darwin lacks a `-a` flag.

spec/unit/plugins/sysconf_spec.rb
```diff
@@ -0,0 +1,683 @@
...
+#
+
+require File.expand_path(File.dirname(__FILE__) + "/../../spec_helper.rb")
+
+describe Ohai::System, "sysconf plugin" do
+  let(:plugin) { get_plugin("sysconf") }
+
+  #before(:each) do
+  #  allow(plugin).to receive(:collect_os).and_return(:linux)
+  #end
```
btm: remove the commented code

thommay:
Running this command on my mac results in:
```
19:01 ~/work/ohai % getconf -a                                         [master]
getconf: illegal option -- a
```

thommay:
`-a` appears to be SYSVish - it works on Solaris and Linux but none of the BSD variants.

btm:
@thommay yeah, that's why I asked for the change to only create the Mash if `getconf -a` succeeds. I couldn't find an equivalent on OS X.

https://github.com/chef/ohai/pull/893/commits/985abb5629933672eb0b0ecf10682e5f873b1ec1#diff-9c1895802e3208d67fc331c2a77898ebR27

jaymzh:
@davide125

davide125:
Yeah you're right, on OSX and on other BSD the `-a` option is not available, and I can't find a way to easily enumerate all the available properties (short of hardcoding them in the plugin). Should I restrict this to just the SysV unices for now, or is ok if it fails on those platforms?

tas50:
It would be nice to avoid at least Windows / OSX in this plugin so we don't waste time on every run with those platforms

davide125:
Alright, latest version only runs for Linux, Solaris and AIX.

PR #894, opened 2016-11-10
Remove the doc changes doc

tas50:
We're not using these anymore


tas50:
Obvious fix rule. Docs don't need DCO

btm:
Yes, but you could just set up git to always sign off and then you wouldn't have to have big all that red show up in the github checks box nor have to cite obscure minutiae. 🌮

PR #895, opened 2016-11-10
Wipe the release notes

tas50:
These were very out of date


tas50:
Merging under obvious fix policy

Issue #896, opened 2016-11-10
languages[:ruby][:gem_bin_dir] is not correctly set


damm:
## Description

* Using ruby from Ubuntu 16.04
node['languages']['ruby']['gem_bin_dir'] is pointed to `/usr/bin` however binaries are installed into `/usr/local/bin`.  

## Ohai Version

root@warehouse:~# ohai --version
Ohai: 8.14.0

Tell us which version of Ohai you are using.

## Platform Version

```
root@warehouse:~# ohai platform
[
  "ubuntu"
]
root@warehouse:~# ohai platform_version
[
  "16.04"
]
```

## Ohai Output

https://gist.github.com/damm/54a944dde9ed3c4459b42ee0f104d2d2

### Personal debugging

Reviewing the code I noticed we use rbconfig here to determine the binary path; it appears rbconfig is lying here.  Rubygems knows the right answer

```
irb(main):008:0* Gem.bindir
=> "/usr/local/bin"
irb(main):009:0> RbConfig::CONFIG['bindir']
=> "/usr/bin"
irb(main):010:0>
```

PR #897, opened 2016-11-10
Update ruby plugin to determine the gem_bin_dir

damm:
Updated tests to match hopefully

Signed-off-by: Scott M. Likens <scott@likens.us>

### DIFF
--- a/lib/ohai/plugins/ruby.rb
+++ b/lib/ohai/plugins/ruby.rb
@@ -21,7 +21,7 @@ Ohai.plugin(:Ruby) do
   depends "languages"
 
   def run_ruby(command)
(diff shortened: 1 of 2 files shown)


damm:
Ref #896

damm:
I really can't figure out the tests for the moment; so I'm not going to leave this PR open.

Maybe my fix is also incorrect ... it's a weird problem 
```
     - /var/lib/gems/2.3.0
     - /root/.gem/ruby/2.3.0
     - /usr/lib/x86_64-linux-gnu/rubygems-integration/2.3.0
     - /usr/share/rubygems-integration/2.3.0
     - /usr/share/rubygems-integration/all
```

So many different paths and often they are valid locations as well.

PR #901, opened 2016-11-11
Add shard plugin support for all platforms

smorad:
Shard is set up in a way to provide sharding regardless of platform, however we don't define a `:default` collect_data, so shards only work on Darwin and Linux.

This allows other platforms to use shard with the basic seed types (so not hardware identifiers).

Signed-off-by: Steven Morad <stevemo@fb.com>


jaymzh:
This won't work - we need to fail on platforms we don't support rather than silently give them bad data...  this plugin needs to either support a platform equally or not at all.

PR #907, opened 2016-11-28
Fetch AWS Account ID from metadata

Fodoj:
In addition to meta-data and userdata EC2 also provides "dynamic" data (not sure why it's called dynamic though). One extra useful attribute that it has, and that seems to be not available in other sources, is AWS account id (try running curl http://169.254.169.254/latest/dynamic/instance-identity/document). It is possible to fetch account id from the instance that has IAM role attached (by parsing ARN of the role) but for instances without any IAM role this dynamic data is the only way to figure out to which account it belongs to.

### DIFF
--- a/lib/ohai/mixin/ec2_metadata.rb
+++ b/lib/ohai/mixin/ec2_metadata.rb
@@ -200,6 +200,18 @@ module Ohai
         response.code == "200" ? response.body : nil
       end
 
+      def fetch_dynamic_data
+        api_version = best_api_version
+        return {} if api_version.nil?
+        response = http_client.get("/#{api_version}/dynamic/instance-identity/document/")
+
+        if json?(response.body) && response.code == "200"
+          FFI_Yajl::Parser.parse(response.body)
+        else
+          {}
+        end
+      end
+
       private
 
       def expand_path(file_name)
--- a/lib/ohai/mixin/ec2_metadata.rb
+++ b/lib/ohai/mixin/ec2_metadata.rb
@@ -201,14 +201,16 @@ module Ohai
       end
 
       def fetch_dynamic_data
-        api_version = best_api_version
-        return {} if api_version.nil?
-        response = http_client.get("/#{api_version}/dynamic/instance-identity/document/")
+        @fetch_dynamic_data ||= begin
+          api_version = best_api_version
+          return {} if api_version.nil?
+          response = http_client.get("/#{api_version}/dynamic/instance-identity/document/")
 
-        if json?(response.body) && response.code == "200"
-          FFI_Yajl::Parser.parse(response.body)
-        else
-          {}
+          if json?(response.body) && response.code == "200"
+            FFI_Yajl::Parser.parse(response.body)
+          else
+            {}
+          end
         end
       end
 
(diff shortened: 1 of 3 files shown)


lib/ohai/mixin/ec2_metadata.rb
```diff
@@ -19,6 +19,7 @@
 
 require "net/http"
 require "socket"
+require "json"
```
lamont-granquist: should be ffi-yajl

lib/ohai/mixin/ec2_metadata.rb
```diff
@@ -200,6 +201,13 @@ def fetch_userdata
         response.code == "200" ? response.body : nil
       end
 
+      def fetch_dynamic_data
+        api_version = best_api_version
+        return nil if api_version.nil?
+        response = http_client.get("/#{api_version}/dynamic/instance-identity/document/")
+        response.code == "200" ? JSON.parse(response.body) : nil
```
lamont-granquist: should be `FFI_Yajl::Parser.parse`

lib/ohai/mixin/ec2_metadata.rb
```diff
@@ -200,6 +201,13 @@ def fetch_userdata
         response.code == "200" ? response.body : nil
       end
 
+      def fetch_dynamic_data
+        api_version = best_api_version
+        return nil if api_version.nil?
```
lamont-granquist: this is going to trainwreck later with a "NoMethodError on NilClass" and should really handle this better

lib/ohai/mixin/ec2_metadata.rb
thommay: Please use FFIYajl for JSON handling; see https://github.com/chef/ohai/blob/master/lib/ohai/mixin/gce_metadata.rb#L71 (also this validates that we've got valid JSON)

lib/ohai/mixin/ec2_metadata.rb
lamont-granquist: and the nil response here similarly should have better error handling if we're gonna worry about it

lib/ohai/mixin/ec2_metadata.rb
```diff
@@ -200,6 +201,13 @@ def fetch_userdata
         response.code == "200" ? response.body : nil
       end
 
+      def fetch_dynamic_data
```
lamont-granquist: also this could be memoized since it looks like you intend it to get reused `@fetch_dynamic_data ||= begin ... end`

Fodoj: I don't think it should be memoized. It's a dynamic data and some values from it can change any time.

Fodoj: Also, other attributes like user data is not memoized as well in this plugin

Fodoj:
@lamont-granquist I've replaced json gem with ffi-yajl. Otherwise implementation is same as fetch_user_data few lines above method I've added. I've also replaced default return value with empty hash, instead of nil, so it doesn't explode inside ec2 plugin. Looking forward to your feedback on this. 

PS

I also rebased and squashed commits into one.

Fodoj:
@lamont-granquist any update here?

Fodoj:
@thommay done! Thanks for your review

Issue #891, opened 2016-10-25
SmartOS VMWare guests who are also hosts should be detected as hosts even when not running nested guests.


numericillustration:
## Description

When running SmartOS as a VMware Fusion guest with nested virtualization enabled, Ohai improperly detects a SmartOS host as a Virtualization guest rather than a host when there are no virtualization guests running on it.  As soon as a single guest instance is running on the SmartOS host, ohai properly detects the host as a host.
## Ohai Version

Ohai: 8.11.1
## Platform Version

uname -a output says:
SunOS coal-cn1 5.11 joyent_20160929T030056Z i86pc i386 i86pc

this system is Joyent SmartOS, based on illumos
###### 

ohai output for a nested virt host that is not running any instances currently:

```
  "virtualization": {
    "system": "vmware",
    "role": "guest"
  },
```

As soon as I create a guest instance on this node ohai will set this section as follows:

```
  "virtualization": {
    "system": "zone",
    "role": "host",
    "guests": {
      "global": {
        "id": "0",
        "state": "running",
        "root": "/",
        "uuid": "",
        "brand": "joyent",
        "ip": "shared"
      },
      "92a772f6-047e-ccde-dd92-ec8f20875055": {
        "id": "4",
        "state": "running",
        "root": "/zones/92a772f6-047e-ccde-dd92-ec8f20875055",
        "uuid": "92a772f6-047e-ccde-dd92-ec8f20875055",
        "brand": "joyent",
        "ip": "excl"
      }
    }
  },
```

my understanding of the code paths that lead to this mis-detection are as follows:

Ohai runs the appropriate virt detection plugin for SmartOS which is solaris2/virtualization 
In the smbios detection section:
https://github.com/chef/ohai/blob/13485c17da688654890047d1952d9f71e735d162/lib/ohai/plugins/solaris2/virtualization.rb#L48-L58

this line in the dmi_decode function is properly returning vmware.
https://github.com/chef/ohai/blob/13485c17da688654890047d1952d9f71e735d162/lib/ohai/mixin/dmi_decode.rb#L33

So ohai sets the role to guest
https://github.com/chef/ohai/blob/13485c17da688654890047d1952d9f71e735d162/lib/ohai/plugins/solaris2/virtualization.rb#L54-L56

a little further on this block of code
https://github.com/chef/ohai/blob/13485c17da688654890047d1952d9f71e735d162/lib/ohai/plugins/solaris2/virtualization.rb#L75-L89
[...]

PR #908, opened 2016-11-29
fix for #891 to detect the global zone of a Solaris system as a Virt …

numericillustration:
…host even when not currently hosting any guests


thommay:
Please run `bundle exec chefstyle -a` to fix the style issues, and then this is good. Thanks!

PR #909, opened 2016-11-29
Rework / fix logic in the joyent plugin and improve specs

tas50:
This builds on the work that @bahamas10 did in #499. It fixes up the specs to provide proper testing of the logic of the plugin and based on those specs makes some additional fixes to the plugin.


tas50:
It should be noted that the original work is covered under the CLA

PR #910, opened 2016-11-29
Release 8.22.0

tas50:

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -8,3 +8,22 @@ Details about the thing that changed that needs to get included in the Release N
 -->
 
 # Ohai Release Notes:
+
+## Haskell Language plugin
+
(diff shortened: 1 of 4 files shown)


RELEASE_NOTES.md
```diff
@@ -8,3 +8,22 @@ Details about the thing that changed that needs to get included in the Release N
...
+
+## Haskell Language plugin
+
+Haskell is now detected in a new haskell language plugin:
+
+```javascript
+"languages": {
+  "haskell": {
+    "stack": {
+      "version": "1.2.",
```
btm: Does this really return `1.2.` with the trailing dot? Is that weird?

tas50: that is a bit odd. Let me see what's up with that before we ship this thing

tas50: Fixed in #914

PR #916, opened 2016-12-01
Pull the complete version string of Erlang

tas50:
Pull the complete version string of Erlang

Improve the shellout with help of StackOverflow and srenatus. This requires less fragile work in Ruby by formatting the data in erlang properly.

### DIFF
--- a/lib/ohai/plugins/erlang.rb
+++ b/lib/ohai/plugins/erlang.rb
@@ -24,13 +24,11 @@ Ohai.plugin(:Erlang) do
     erlang = Mash.new
 
     begin
-      so = shell_out("erl -eval 'erlang:display(erlang:system_info(otp_release)), erlang:display(erlang:system_info(version)), erlang:display(erlang:system_info(nif_version)), halt().'  -noshell")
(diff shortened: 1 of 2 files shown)


lib/ohai/plugins/erlang.rb
```diff
@@ -24,13 +24,13 @@
     erlang = Mash.new
 
     begin
-      so = shell_out("erl -eval 'erlang:display(erlang:system_info(otp_release)), erlang:display(erlang:system_info(version)), erlang:display(erlang:system_info(nif_version)), halt().'  -noshell")
+      so = shell_out("erl -eval '{ok, Ver} = file:read_file(filename:join([code:root_dir(), \"releases\", erlang:system_info(otp_release), \"OTP_VERSION\"])), Vsn = binary:bin_to_list(Ver, {0, byte_size(Ver) - 1}), io:format(\"~s,~s,~s\", [Vsn, erlang:system_info(version), erlang:system_info(nif_version)]), halt().' -noshell")
       # Sample output:
-      # "18"
-      # "7.3"
-      # "2.10"
+      # 19.1
```
thommay: sample output now is surely `19.1,8.1,2.11` ?

tas50: true

tas50:
@thommay Example output fixed

PR #921, opened 2016-12-14
Properly detect Cumulus Linux platform / version

tas50:
### Description

Detect cumulus Linux systems if they have /etc/cumulus directory. Use platform of ‘cumulus’ and platform_family debian since it’s Debian 8.5. Parse version from their config directory.

### Issues Resolved

N/A

### Check List

- [x] All tests pass

### DIFF
--- a/lib/ohai/plugins/linux/platform.rb
+++ b/lib/ohai/plugins/linux/platform.rb
@@ -73,18 +73,26 @@ Ohai.plugin(:Platform) do
   end
 
   #
+  # Determines the platform version for Cumulus Linux systems
+  #
+  # @returns [String] cumulus Linux version from /etc/cumulus/etc.replace/os-release
+  #
+  def cumulus_version
+    release_contents = File.read("/etc/cumulus/etc.replace/os-release")
+    release_contents.match(/VERSION_ID=(.*)/)[1]
+  rescue NoMethodError, Errno::ENOENT, Errno::EACCES # rescue regex failure, file missing, or permission denied
+    Ohai::Log.warn("Detected Cumulus Linux, but /etc/cumulus/etc/replace/os-release could not be parsed to determine platform_version")
+    nil
+  end
+
+  #
   # Determines the platform version for Debian based systems
   #
   # @returns [String] version of the platform
   #
   def debian_platform_version
     if platform == "cumulus"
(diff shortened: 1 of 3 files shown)


spec/unit/plugins/linux/platform_spec.rb
```diff
@@ -187,6 +225,12 @@
       expect(@plugin[:platform]).to eq("slackware")
       expect(@plugin[:platform_family]).to eq("slackware")
     end
+
+    it "should set platform_version on slackware" do
```
tas50: This was missing so I figured I should add it while I was here

lib/ohai/plugins/linux/platform.rb
```diff
@@ -72,6 +72,54 @@ def os_release_file_is_cisco?
...
+  #
+  # Determines the platform version for Debian based systems
+  #
+  # @returns [String] version of the platform
+  #
+  def debian_platform_version
+    if platform == "cumulus"
+      if File.exist?("/etc/cumulus/etc.replace/os-release")
+        release_contents = File.read("/etc/cumulus/etc.replace/os-release")
+        release_contents.match(/VERSION_ID=(.*)/)[1] rescue nil
```
btm: maybe you want the warning below here as well instead of a nil rescue

tas50: Probably. I keep going back and forth with how much Ohai should warn when things don't go right. Recently leaning more to towards Ohai being verbose when things fail so that people see that things are failing and file issues. We have a lot of hidden failures right now, which doesn't help anyone.

tas50:
@btm does that last commit work for you?

lib/ohai/plugins/linux/platform.rb
```diff
@@ -72,6 +72,62 @@ def os_release_file_is_cisco?
...
+  def debian_platform_version
+    if platform == "cumulus"
+      cumulus_version
+    else # not cumulus
+      File.read("/etc/debian_version").chomp
+    end
+  end
+
+  #
+  # Determines the platform_family based on the platform_family
```
thommay: "based on the platform"

tas50: I like my wording better ;)

btm: Determines the platform_family based on the platform family? :)

PR #922, opened 2016-12-14
Windows guest support on Hyper-v and Xen.

### LINKED ISSUE #664: Detect Hyper-V guests running Windows
I added the logic to properly detect Hyper-V guests running Linux.  We just need to move that same logic over to Windows.

rdean716:
### Description

This change allows Windows guest systems to detect the virtualization platform when running on both Hyper-v and Xen.

### Issues Resolved

#664 

### Check List

- [ X] New functionality includes tests
- [ X] All tests pass
- [ X] All commits have been signed-off for the Developer Certificate of Origin. See <https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco>

Signed-off-by: Bob Dean <radean@hotmail.com>

### DIFF
--- a/lib/ohai/plugins/windows/virtualization.rb
+++ b/lib/ohai/plugins/windows/virtualization.rb
@@ -44,6 +44,14 @@ Ohai.plugin(:Virtualization) do
       virtualization[:system] = "kvm"
       virtualization[:role] = "guest"
       virtualization[:systems][:kvm] = "guest"
+    when "American Megatrends Inc."
+      virtualization[:system] = "hyper-v"
+      virtualization[:role] = "guest"
+      virtualization[:systems][:hyperv] = "guest"
+    when "Xen"
+      virtualization[:system] = "xen"
+      virtualization[:role] = "guest"
+      virtualization[:systems][:xen] = "guest"
     end
 
     # vmware fusion detection
--- a/lib/ohai/plugins/windows/virtualization.rb
+++ b/lib/ohai/plugins/windows/virtualization.rb
@@ -45,9 +45,11 @@ Ohai.plugin(:Virtualization) do
       virtualization[:role] = "guest"
       virtualization[:systems][:kvm] = "guest"
     when "American Megatrends Inc."
-      virtualization[:system] = "hyper-v"
-      virtualization[:role] = "guest"
-      virtualization[:systems][:hyperv] = "guest"
+      if bios[0]["version"] =~ /VRTUAL -/
+        virtualization[:system] = "hyper-v"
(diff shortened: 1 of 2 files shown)


lib/ohai/plugins/windows/virtualization.rb
```diff
@@ -44,6 +44,14 @@
       virtualization[:system] = "kvm"
       virtualization[:role] = "guest"
       virtualization[:systems][:kvm] = "guest"
+    when "American Megatrends Inc."
```
tas50: This is probably going to catch some systems that are also using American Megatrends BIOS and identify them as VMs when they're actually hardware. We need to key off this and something else that is unique to a hyper-v system.

rdean716: In reviewing a few different guests running on different version of hyper-v, the smbiosbiosversion property appears to be '090006' consistently.  Reasonable to evaluate that when the manufacturer is 'American Megatrends Inc.'?

tas50: We could probably use that.

rdean716: The next property that seems to be semi consistent is version.  On one version of hyper-v its value is 'VRTUAL - 5001223' and on another it is 'VRTUAL - 4001628'.  I have no idea what factors cause this to change. But, it seems messy to try to match on a partial string.

rdean716: I can!

tas50:
@rdean716 Also make sure to rebase that last commit with the DCO sign-off

lib/ohai/plugins/windows/virtualization.rb
rdean716: I am fairly new to ruby, are you ok with this - if bios[0]["version"].include? "VRTUAL -" ...

tas50: I think what you want here is a match with a regex. There's two ways to do that with ruby, but the easiest here is:

if bios[0]["version"] =~ /VRTUAL -/

Here's a nice doc with all your options:
https://ruby-doc.org/core-2.2.0/Regexp.html

rdean716: Thanks for the reference.

tas50:
@rdean716 You have a Chefstyle warning showing up in travis. Run chefstyle -a and that should get autocorrected.

rdean716:
@tas50 - everything looks good now.

tas50:
Nice stuff. Thanks

PR #930, opened 2016-12-20
Remove usage of the Sigar gem

tas50:
Sigar was a really cool idea, but it’s a dead project now. The data is questionable at best and not getting any better as time goes on. We should figure out how to support these platforms vs. hoping that Sigar is doing the right thing. This does remove a lot of support for HP-UX, but I seriously doubt it worked to begin with. We should do that right sometime.


thommay:
There's no ruby on HP-UX anymore AIUI so this doesn't really matter!

tas50:
Well I don't feel bad then

tas50:

PR #931, opened 2016-12-20
Move the “Virtualization” plugin into libvirt namespace

tas50:
This plugin isn’t actually a virtualization plugin. It’s actually a libvirt plugin. Move the data out of the Virtualization mash and stick it into its own location. Also rename the plugin libvirt. I need to get a system setup with libvirt to test this and probably tweak it further / add tests.


tas50:

thommay:
the capitals were bugging me, but the code is fine. Tests?

tas50:
LibVirt -> Libvirt for the plugin name now

PR #933, opened 2016-12-21
Remove the deprecated run_command and popen4 methods

tas50:
This is very much legacy and has long since been replaced by shell_out


tas50:

thommay:
as a comment, let's stop adding author tags.

tas50:
I'm all for killing author tags. In this case all the code those authors wrote was deleted and everything left was mine.

PR #934, opened 2016-12-21
Bump ruby versions and test on ruby head

tas50:
Test ruby 2.2.6 and 2.3.3 and now ruby-head, which is allowed to fail

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,8 +14,12 @@ before_install:
   - gem --version
 rvm:
   - 2.1.9
-  - 2.2.5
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -14,8 +14,12 @@ before_install:
   - gem --version
 rvm:
   - 2.1.9
```
lamont-granquist: we can drop this right?  but whatevs

tas50:
We didn't actually drop 2.1 support in Ohai (yet)

PR #938, opened 2017-01-05
Require Ruby 2.2+

tas50:
chef/chef already requires Ruby 2.2. We've removed support for Ruby 2.1
in our other projects and community tooling.

I think this is a pretty minor impact, but we can wait till Ohai 13 in April if we want.


thommay:
need to fix appveyor too

tas50:
every time appveyer trolls me

PR #939, opened 2017-01-08
Remove deprecated config logic

tas50:
Start of the deprecation of the old config logic. Removes a ton of confusing duplicate logic and resolves 2 bogus deprecation warnings that were thrown every time Ohai ran. Specs are going to need to be cleaned up still.

Also bonus gift: This adds /etc/chef/ohai/plugins to the plugin path so the ohai_plugin resource will work without configuration mess by the user now. I'll probably port that same 1 line fix over to Ohai 8.X so we can get that in for Chef 12 users.

### DIFF
--- a/lib/ohai/config.rb
+++ b/lib/ohai/config.rb
@@ -19,7 +19,6 @@
 
 require "chef-config/config"
 require "ohai/exception"
-require "ohai/log"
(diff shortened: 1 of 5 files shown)


tas50:

PR #941, opened 2017-01-15
fix gcc and cc for darwin (no cli tools installed)

erikng:
If darwin is found:
1. shell out to /usr/bin/xcode-select -p
2a. If it returns an exit code other than 0, assume xcode command line tools are not installed and do not run these commands.
2b. If it returns an exit code of 0, run the commands.

Signed-off-by: Erik Gomez <e@eriknicolasgomez.com>

### Description

This removes the requirement for both Ohai and Chef to have the Xcode command line tools installed on macOS machines.

This is a huge win for managing client devices as the Xcode Command Line tools were previously a dependency. Companies have worked around these issues, but they are not ideal and require additional management overhead/engineering effort.

One such example: https://github.com/facebook/IT-CPE/blob/master/chef/tools/chef_bootstrap.py#L239-L284

### Issues Resolved

Fixes https://github.com/chef/ohai/issues/940

### Check List
Since this only impacts Darwin, I have tested both use cases where xcode command line tools are installed and when they are not installed.

### DIFF
--- a/lib/ohai/plugins/c.rb
+++ b/lib/ohai/plugins/c.rb
@@ -48,30 +48,6 @@ Ohai.plugin(:C) do
     Ohai::Log.debug("xcode-select binary could not be found. Skipping data.")
   end
 
-  collect_data(:darwin) do
-    if xcode_installed?
-      collect_gcc
-      collect_cc
-    end
-  end
-
-  collect_data(:windows) do
-    check_for_cl
-    check_for_devenv
-  end
-
-  collect_data(:default) do
-    collect_gcc
-    collect_glibc
-    check_for_cl
-    check_for_devenv
-    collect_xlc
-    collect_cc
-    collect_hpux_cc
-  end
-
-  c = Mash.new
-
   def collect_gcc
     #gcc
     collect("gcc -v") do |so|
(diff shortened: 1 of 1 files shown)


natewalck:
This would be an amazing fix to get merged. Obviously the tests must be updated, but is the overall form acceptable? (CC  @jaymzh @mcquin @tas50 )

erikng:
To be clear, the integration tests are probably failing due to the current code and not my changes as the CI tools do not run on macOS.

thommay:
@erikng the tests are failing due to expectations in the specs that are no longer true, so you'll need to fix those up. Chefstyle also has some errors that you'll need to fix (run `bundle exec rake style:auto_correct`).

In general though, I think each check ought to be more split out. For darwin, linux, etc, the checks that matter are gcc and glibc, so we could wrap those in a function, and then do something like:

```
collect_data(:darwin) do
  if xcode_installed
    collect_gcc
    collect_glibc
  end
end

collect_data(:windows) do
  # check for cl
  # check for devenv.com
end

collect_data(:default) do
  collect_gcc
  collect_glibc
end
```

If that seems like a too major refactor, that's ok, just let us know and we can take it on from here (but please tick the box to "Allow edits from maintainers" on this PR - https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/ )

erikng:
I would certainly be willing to do it for the Darwin side, but as this PR was designed to fix the specific issues presented, I had no intention of doing other operating systems.

If you were to do the refactor, what are the timetables here? This would infinitely make our deployments easier and I would hate for other factors to delay this fix.

erikng:
I went ahead and completely refactored it, based on the feedback you gave @thommay (wasn't as bad as I expected) and the Chefstyle is no longer producing errors.

That said, I have very little ideas on how to fix the spec tests, especially now. Some feedback would be greatly appreciated.

erikng:
No problem. I've tried various things today to understand the spec, but I'm not getting any closer to writing it. If someone could give me a little feedback on where to start here it would be appreciated, otherwise I may not be able to finish this PR.

I would hate to close it, but it may be infinitely easier for me to just deploy this as a hotfix rather than spend time I don't have fixing the specs. 

If I sound bitter, I apologize, but I've wasted about 4 hours of time today and gotten almost zero progress.

erikng:
Closing due to https://github.com/chef/ohai/pull/944

Issue #878, opened 2016-09-19
ohai no longer prints debug info in standalone executions


jaymzh:
## Description

When I run `ohai -l debug` I expect to see plugin failures.

But I do not.
## Ohai Version

8.19.1.

Chef 12.13.37
## Platform Version

CentOS7
## Ohai Output

N/A

tas50:
The log file location flag seems to be broken as well

erikng:
This really makes it hard to debug Ohai and contribute code.

tas50:
This commit cause the logging to not be reconfigured with the CLI arguments, which breaks the CLI logging configs:

https://github.com/chef/ohai/pull/863/files#diff-dee5766eb9ab7555535dae03a4674f28R223

If you comment out the return line things work, but I know there was some backstory as to why we did this. @thommay can probably provide the backstory.

PR #942, opened 2017-01-20
Only explicitly initialize a logger when needed

thommay:
When we're running Ohai as a standalone application, we need to fire up
the logger with our desired location, but when we're run from inside
Chef, we're passed the client's loggers.

( https://github.com/chef/chef/blob/master/lib/chef/application/client.rb#L398 )

Fixes: #878


tas50:
Works great locally for me.

PR #944, opened 2017-01-21
C Plugin: Expand GCC data & only shellout to gcc if Xcode is installed on macOS

tas50:
Builds on the work that @erikng did in #941 

Additions:

Rework the specs to be per platform and remove some the guards that existed for *nix plugins on Windows
Add more useful fields to the gcc data
Use the same logging format everywhere
Add specs for the xcode change and rework other specs to run per platform

### DIFF
--- a/lib/ohai/plugins/c.rb
+++ b/lib/ohai/plugins/c.rb
@@ -48,21 +48,23 @@ Ohai.plugin(:C) do
   end
 
   def collect_gcc
-    #gcc
-    collect("gcc -v") do |so|
-      # Sample output:
-      # Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
-      # Apple LLVM version 7.3.0 (clang-703.0.29)
-      # Target: x86_64-apple-darwin15.4.0
-      # Thread model: posix
-      # InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
-      description = so.stderr.split($/).last
-      output = description.split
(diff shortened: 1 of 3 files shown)


erikng:
Awesome job @tas50. Looks like the CI is failing, but not due to the PR.

I don't want to speak for too many people, but both @natewalck and I would love to see this merged into the next chef release.

lib/ohai/plugins/c.rb
```diff
@@ -34,94 +31,163 @@ def collect(cmd, &block)
...
+    languages[:c] = @c unless @c.empty?
+  end
 
-    languages[:c] = c unless c.empty?
+  collect_data(:default) do
+    @c = Mash.new
+    collect_gcc
+    collect_glibc
+    collect_xlc
+    collect_cc
```
thommay: `collect_cc` is pretty badly named here, and is solaris specific.

lib/ohai/plugins/c.rb
```diff
@@ -34,94 +31,163 @@ def collect(cmd, &block)
...
+    collect_hpux_cc
+    languages[:c] = @c unless @c.empty?
+  end
 
-    languages[:c] = c unless c.empty?
+  collect_data(:default) do
+    @c = Mash.new
+    collect_gcc
+    collect_glibc
+    collect_xlc
```
thommay: believe this is AIX specific.

lib/ohai/plugins/c.rb
tas50: Fixed

lib/ohai/plugins/c.rb
tas50: Fixed

PR #945, opened 2017-01-23
Release 8.23.0

tas50:


tas50:
Appveyer is failing on rainbow gem. We probably need to pull in the same fix that was used in chef/chef and dk to pin that back for now.

tas50:
Ready @thommay other than the Appveyer fix, which can be fixed outside this PR

PR #947, opened 2017-01-24
Config is used in system so require it there.

tas50:
COOL team found this when using ohai/system directly. It worked in 8.X, but not post deprecation of the old config logic.

### DIFF
--- a/lib/ohai/system.rb
+++ b/lib/ohai/system.rb
@@ -28,6 +28,7 @@ require "ohai/mixin/constant_helper"
 require "ohai/provides_map"
 require "ohai/hints"
 require "mixlib/shellout"
+require "ohai/config"
 
 module Ohai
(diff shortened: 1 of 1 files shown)


lib/ohai.rb
```diff
@@ -17,6 +17,5 @@
 #
 
 require "ohai/version"
-require "ohai/config"
```
thommay: Removing this seems bad; it's needed for Ohai::Application.

tas50: We can keep it here as well, but it's getting included with Application still:

Application -> Ohai -> System -> Config

thommay: require what you use. Implicit requirements have caused us so many bugs over the years.

tas50:
Required in both locations now @thommay

PR #951, opened 2017-02-03
Move duplicate http logic into a helper

tas50:
We probably want to use this in a few other places, but we certainly want to encourage people to use this in their plugins vs. reinventing the wheel.


tas50:
End goal here is to have a http helper than can be used by anyone writing an ohai plugin that makes http calls. This is just the first part and allows deduping the codebase a bit.

tas50:
Tests will happen before we announce this as a helper for public consumption. They're long overdue here.

PR #952, opened 2017-02-03
Rescue exception in DMI plugin

tas50:
This gets us non-exploding output on Linux systems now. This was the
last plugin.


btm:
Is it likely for `dmidecode` to fail for a reason other than it doesn't exist? if so, should we be including the exception in the debug output?

tas50:
I'm fairly certain this is the only failure. At least I haven't seen / heard of another failure with the dmi plugin

thommay:
https://github.com/chef/ohai/pull/952/files?w=1 makes this review much clearer.

PR #955, opened 2017-02-28
fix logger issues

lamont-granquist:
fix ohai logger problems.

    i believe this gets the logic correct in that we want to configure the
    logger when we're coming from the command line but want to skip it when
    we're coming from chef or from other APIs that directly construct an
    Ohai::System object.

    i suspect this is what thom was trying to do by moving the
    Ohai::Log.init call into the Ohai::Application class (which avoids it
    being called entirely when Chef creates its Ohai::System object) here:

    https://github.com/chef/ohai/pull/942/files

    but that change broke the behavior where we were also supposed to skip
    the rest of the configure_logging method in Ohai::System when run under
    chef client.

    I tried going down the route of having the Ohai::Application class
    construct the Ohai::System object and then having it be the
    responsibility of that caller to do configure_logging work.  However,
    I suspect that the initializer in Ohai::System does way too much and
    that the purpose of configuring the logger where it is, is that it
    must be initialized in the middle of object creation before it goes on
    and creates the Loader, the Runner, creates Hints and removes constants.

    So, I went the route of threading a flag through the initializer so that
    Ohai::System can know if its coming from the cli or not and behave
    appropriately.

    There's also quite a mess with the Ohai::Log class being passed around
    to the workstation loader, and it initializes itself at class loading
    time and Chef::Application will inject state into Ohai::Log.  I think
    the eager initialization at class loading time is to not lose early
    messages when run from the cli, but it means the Ohai::Log object is always
    initialized, so that isn't useful for determining if we have come from the
    cli or not.

### DIFF
--- a/lib/ohai/application.rb
+++ b/lib/ohai/application.rb
@@ -88,7 +88,8 @@ class Ohai::Application
   end
 
   def run_application
-    ohai = Ohai::System.new(config, cli: true)
+    config[:invoked_from_cli] = true
+    ohai = Ohai::System.new(config)
     ohai.all_plugins(@attributes)
 
     if @attributes
--- a/lib/ohai/application.rb
+++ b/lib/ohai/application.rb
@@ -88,7 +88,7 @@ class Ohai::Application
   end
 
   def run_application
-    ohai = Ohai::System.new(config)
+    ohai = Ohai::System.new(config, cli: true)
     ohai.all_plugins(@attributes)
 
     if @attributes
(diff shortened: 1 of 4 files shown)


lamont-granquist:
@tas50 @cheeseplus i think this is more-or-less correct

lamont-granquist:
optional hashes and named arguments in ruby #FML

lamont-granquist:
@tas50 @cheeseplus once more unto the breach, dear friends, once more

PR #960, opened 2017-03-03
move 'amazon' to 'fedora' platform-family from 'rhel'

lamont-granquist:
closes chef/chef#4665

### DIFF
--- a/lib/ohai/plugins/linux/platform.rb
+++ b/lib/ohai/plugins/linux/platform.rb
@@ -1,6 +1,6 @@
 #
 # Author:: Adam Jacob (<adam@chef.io>)
-# Copyright:: Copyright (c) 2015-2016 Chef Software, Inc.
+# Copyright:: Copyright (c) 2015-2017, Chef Software Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -107,11 +107,11 @@ Ohai.plugin(:Platform) do
     case platform
     when /debian/, /ubuntu/, /linuxmint/, /raspbian/, /cumulus/
       "debian"
-    when /oracle/, /centos/, /redhat/, /scientific/, /enterpriseenterprise/, /amazon/, /xenserver/, /cloudlinux/, /ibm_powerkvm/, /parallels/, /nexus_centos/ # Note that 'enterpriseenterprise' is oracle's LSB "distributor ID"
+    when /oracle/, /centos/, /redhat/, /scientific/, /enterpriseenterprise/, /xenserver/, /cloudlinux/, /ibm_powerkvm/, /parallels/, /nexus_centos/ # Note that 'enterpriseenterprise' is oracle's LSB "distributor ID"
       "rhel"
     when /suse/
       "suse"
-    when /fedora/, /pidora/, /arista_eos/
+    when /fedora/, /pidora/, /arista_eos/, /amazon/
       "fedora"
     when /nexus/, /ios_xr/
       "wrlinux"
--- a/spec/unit/plugins/linux/platform_spec.rb
+++ b/spec/unit/plugins/linux/platform_spec.rb
@@ -1,6 +1,6 @@
 #
 # Author:: Adam Jacob (<adam@chef.io>)
-# Copyright:: Copyright (c) 2008-2016 Chef Software, Inc.
+# Copyright:: Copyright (c) 2008-2017, Chef Software Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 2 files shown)


lamont-granquist:
Once again, this is the change which produces less work, not more work:

Both before and after we have to change this logic, which is the point of the change:

```ruby
if node[:platform_family] == "rhel"
  if node[:platform] == "amazon"
    # special case amazon
  else
    # do stuff like rhel
  end
end
```

Both before and after this change though this logic still works fine:

```ruby
case node[:platform_family]
when "rhel", "fedora"
  # we don't care about the differences here
end
```

If we give amazon its own family then we also generate additional work to change all those occurrences to:

```ruby
case node[:platform_family]
when "rhel, "fedora", "amazon"
  # we don't care about the differences here
end
```

Special cases where they're all different are still special cases, and most likely look like case #1 above.

This construct is recommended for backwards compatibility between < 13 and >= 13 when behaviors differ:

```ruby
if node[:platform] == "amazon"
  # do amazon stuff before we check for rhel
elsif node[:platform_family] == "rhel"
  # now check for rhel
else
  # we must be fedora
end
```

thommay:
I think the position I'm coming to on this is that Amazon linux is clearly neither RHEL nor Fedora, but it's own thing. This is only likely to get more pronounced as the systems age.
Thinking about this change in terms of work caused is cruel empathy, and we should do the hard right thing here - as you say, for a large proportion of cases users will already have to update their cookbooks, and we should ensure that they only need do that once.
I vote we add an `amazon` `platform_family`.
cc @chef/maintainers

chris-rock:
@stevendanna We manage os families different in InSpec. In Train, amazon is not part of the redhat family as you see is https://github.com/chef/train/blob/master/lib/train/extras/os_detect_linux.rb#L17. We also discouraged the use of os[:family] in train/inspec since this is a hard-coded string. Eg. is a bsd a unix? We introduced helper method to do that like `os.unix?, os.bsd? os.redhat?`. Except from that we stick with os `name` and os `release`. `family` is mainly there for Serverspec backward-compatibility.

lamont-granquist:
so @thommay we keep going around and around on this.

platform_family was defined to group somewhat similarly derived platforms which are often similar enough to ignore the differences.  so for example, debian, ubuntu and linuxmint are all platform_family of "debian"

we're heading towards a world where the only useful platform family is "rhel", which is a perverse outcome and renders it largely useless.

i actually agree that helpers like `rhel?` are better -- but having amazon categorized as 'rhel' is clearly wrong.

so having a platform_family of fedora here does not mean that amazon is like fedora -- it simply means that it is derived from fedora or it is "fedoraish".

and if you look at all the places where we use a platform_family of fedora internally in chef/chef:

```ruby
lib/chef/provider/ifconfig/redhat.rb:        provides :ifconfig, platform_family: %w{fedora rhel}
lib/chef/provider/package/dnf.rb:        provides :package, platform_family: %w{rhel fedora} do
lib/chef/provider/package/yum.rb:        provides :package, platform_family: %w{rhel fedora}
lib/chef/provider/route.rb:        when "centos", "redhat", "fedora"
lib/chef/provider/service/insserv.rb:        provides :service, platform_family: %w{debian rhel fedora suse} do |node|
lib/chef/provider/service/redhat.rb:        provides :service, platform_family: %w{rhel fedora suse} do |node|
lib/chef/provider/service.rb:        Chef.set_provider_priority_array :service, [ Systemd, Insserv, Redhat ], platform_family: %w{rhel fedora suse}
lib/chef/resource/dnf_package.rb:      provides :package, platform_family: %w{rhel fedora} do
lib/chef/resource/yum_package.rb:      provides :package, os: "linux", platform_family: %w{rhel fedora}
```

we always use 'fedora' together with 'rhel' because we simply don't care.  there's no added value to adding 'amazon' to all of these and its just more typing.

so again, describe to me how the amazon linux lineage from fedora is substantially different from the ubuntu or linuxmint lineage from debian?

i think there's still things like systemd-vs-initscripts and yum-vs-dnf which are core differences, but amazon is mostly just lagging behind and can be expected to switch at some point, and we've done the right thing there and duck-typed the O/S via `provides-with-a-block` so that we don't have to carry forwards if-else conditional around platform_version.

lamont-granquist:
and @thommay its not cruel empathy.  the point is that there's no utility to creating that distinction that isn't already captured by the one-off platform of "amazon".  in fact, i would argue that one-off platform_families are largely "code smell" (and really the BSD and Solaris one-off platform_families are also useless and code smell).   the point here is that nobody really needs this abstraction in order to get work done, and it doesn't help anyone, and that making the change would cause measurable amounts of harm (lots more fixing).  that's actual empathy, not cruel empathy.

thommay:
OK, so let's remove amazon from RHEL, and not add it to Fedora? I think the fact we've gone back and forth so many times is indication enough that we're not really convinced that we have a good solution. 
I'm not sure I'd agree that the platform is likely to become more fedora-ish over time - there don't seem to be indications that systemd support is on the horizon, and I can't imagine they'll bother with DNF until it's a lot more mature.

PR #964, opened 2017-03-17
[ec2] Add additional data from identity document

webframp:
### Description

Include the region and availability zone as information retrieved from the identity document. Since it's already fetching the identity document to retrieve the account id, this just adds the additional info from the same source.

[Other properties](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-identity-documents.html?shortFooter=true) could also be included but these were the two I needed most.

### Issues Resolved

None that I know of

### Check List

- [ ] All tests pass
- [ ] All commits have been signed-off for the Developer Certificate of Origin. See <https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco>

### DIFF
--- a/lib/ohai/plugins/ec2.rb
+++ b/lib/ohai/plugins/ec2.rb
@@ -99,6 +99,8 @@ Ohai.plugin(:EC2) do
       end
       ec2[:userdata] = fetch_userdata
       ec2[:account_id] = fetch_dynamic_data["accountId"]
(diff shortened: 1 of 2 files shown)


tas50:
Not sure if placement_availability_zone and availability_zone ever differ, but I guess we have both now. Thanks for the fix

PR #970, opened 2017-03-29
Make the DMI IDs we whitelist configurable

tas50:
A lot of people want their OEM / IPMI dmi data. It’s super handy to have.

### DIFF
--- a/lib/ohai/common/dmi.rb
+++ b/lib/ohai/common/dmi.rb
@@ -72,9 +72,9 @@ module Ohai
         127 => "end_of_table_marker",
       }
 
-      # list of IDs to collect, otherwise we generate pages of hashes about cache chip size and whatnot
-      # See OHAI-260. When we can give the user a choice, this will be a default.
-      ID_TO_CAPTURE = [ 0, 1, 2, 3, 4, 6, 11 ]
+      # list of IDs to collect from config or default to a sane list that prunes
+      # away some of the less useful IDs
+      ID_TO_CAPTURE = Ohai.config[:plugin][:dmi][:ids] || [ 0, 1, 2, 3, 4, 6, 11 ]
 
       # look up DMI ID
       def id_lookup(id)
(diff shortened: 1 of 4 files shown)


lib/ohai/common/dmi.rb
```diff
@@ -69,9 +69,9 @@ module DMI
         127 => "end_of_table_marker",
       }
 
-      # list of IDs to collect, otherwise we generate pages of hashes about cache chip size and whatnot
-      # See OHAI-260. When we can give the user a choice, this will be a default.
-      ID_TO_CAPTURE = [ 0, 1, 2, 3, 4, 6, 11 ]
+      # list of IDs to collect from config or default to a sane list that prunes
+      # away some of the less useful IDs
+      ID_TO_CAPTURE = Ohai.config[:plugin][:dmi][:ids] || [ 0, 1, 2, 3, 4, 6, 11 ]
```
thommay: hm, i don't think we should force any user that wants to add a single DMI ID to have to specify the full list of defaults too.

tas50: I could make this additional IDs pretty easily

tas50:
@thommay does this sort of thing make sense to allow the user to provide both IDs and the description we use to build the Mash?

thommay:
This seems good; you might want to think about accepting an array too and just returning the code if there's no description.

spec/unit/plugins/dmi_spec.rb
```diff
@@ -128,21 +129,28 @@
...
+        expect(plugin[:dmi][id][attribute]).to eql(value)
       end
     end
   end
 
-  it "should correctly ignore unwanted data" do
-    @plugin.run
-    expect(@plugin[:dmi][:base_board]).not_to have_key(:error_correction_type)
+  it "allows capturing additional DMI data" do
+    Ohai.config[:additional_dmi_ids] = [ 16 ]
```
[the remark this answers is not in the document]
thommay: You were shoving additional IDs on to a constant, so that lived forever.

PR #971, opened 2017-04-02
add amazon platform_family

lamont-granquist:
i think we've finally decided not to move amazon to fedora in #960

i still think everyone is wrong, but i got my `fedora_derived?`
chef-sugar helper so its a mute point now...

https://github.com/chef/chef-rfc/pull/252

### DIFF
--- a/lib/ohai/plugins/linux/platform.rb
+++ b/lib/ohai/plugins/linux/platform.rb
@@ -1,6 +1,6 @@
 #
 # Author:: Adam Jacob (<adam@chef.io>)
-# Copyright:: Copyright (c) 2015-2016 Chef Software, Inc.
+# Copyright:: Copyright (c) 2015-2017, Chef Software Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


lamont-granquist:
Grammar trolling in commit messages is about the best I can do for 4/1

lamont-granquist:
@chef/client-core agreed?

and it looks like faraday is screwed, but this should be green.

lamont-granquist:
Burn

eric-tucker:
Just ran into this trying out Chef 13.

Amazon Linux is based off CentOS 6, hence still running sysvinit and using epel 6.

Is there a documented discussion as to why this change was **necessary**?

This change pretty much broke every cookbook under the sun that was relying on it being under the "rhel" identifier (if moving to Chef 13).

coderanger:
If anyone has further questions on this change, feel free to ask in Slack or the mailing list but this issue is not the place.

lamont-granquist:
It was absolutely wrong to have it in the 'rhel' family -- everyone else in that family tracks the upstream and platform_versions are consistent such that "rhel 6.3 = centos 6.3 = oracle 6.3 = scientific 6.3".

amazon does not track a consistent release of rhel and it used to be rhel5 userland + updated rhel6 kernels.  then it largely shifted to rhel6 for a long time which somewhat masked the fact that it wasn't really rhel -- although platform_version comparisons to detect RHEL6 or RHEL7 would often go completely haywire.  now amazon linux is deviating from rhel7 in major ways and again its no longer correctly that its rhel6 or rhel7.

i argued that it should be changed to a platform_family of 'fedora' so that code which had a `fedora` check and only did the latest stuff would tend to still work correctly on `amazon`, which would have been less disruptive (and i'd argue makes `platform_family` as a concept more useful) but was pretty much the lone voice in that argument:

- https://github.com/chef/chef-rfc/pull/109
- https://github.com/chef/chef/issues/4665
- https://github.com/chef/ohai/pull/960

there were also public discussions in the chef community meetings (the thursday meetings) on several different occasions over this.

the way we wound up here was literally a 2 year long public ~~argument~~ discussion.

PR #972, opened 2017-04-03
Freeze all string values coming out of Ohai.

coderanger:
This both helps avoid weird mutation bugs and saves a bit of RAM. This does not freeze the hashes or arrays so replacement and whatnot should all work the same. Anyone doing in-place mutation of Ohai strings on purpose is getting what they deserve IMO.

Refs https://github.com/chef/chef/issues/4930 for original discussion.


thommay:
although travis is very red.

PR #973, opened 2017-04-03
Replace the cloud plugin with the cloud_v2 plugin

tas50:
The cloud plugin is the first thing a user would see and new users would gravitate to that plugin. The plugin format is flawed though and we long ago replaced it with cloud_v2. This plugin removes the exist cloud plugin and instead has the cloud_v2 plugin write to both node['cloud'] and node['cloud_v2']. This is going to need an additional test to ensure we get data in both namespaces.


spec/unit/plugins/cloud_v2_spec.rb
```diff
@@ -92,6 +92,18 @@
...
 
+  describe "with a cloud mash" do
+    before do
+      @plugin[:ec2] = Mash.new
+    end
+
+    it "populates cloud public ip" do
+      @plugin.run
+      expect(@plugin[:cloud_v2]).to_not be_nil
+      expect(@plugin[:cloud]).to_not be_nil
```
lamont-granquist: same here maybe check they're both equal.

tas50:
@lamont-granquist Fixed it for you

PR #974, opened 2017-04-03
Replace the existing filesystem plugin on Linux and Darwin with filesystem_v2

tas50:
Filesystem’s implementation leaves a lot to be desired and contains bugs that continually burn new users. This change removes the legacy filesystem plugin on Linux and changes the filesystem_v2 plugin to write to both node[‘filesystem’] and node[‘filesystem_v2’]


spec/unit/plugins/linux/filesystem2_spec.rb
```diff
@@ -52,6 +52,12 @@
     end
   end
 
+  it "sets both filesystem and filesystem_v2 attributes" do
+    plugin.run
+    expect(plugin[:filesystem]).to_not be_nil
+    expect(plugin[:filesystem_v2]).to_not be_nil
```
lamont-granquist: `expect(plugin[:filesystem]).to eq(plugin[:filesystem_v2])` as well maybe?

but this test also seems to be failing and filesystem_v2 seems to be nil for some reason...

jaymzh:
It seems confusing to me to go this way rather than renaming filesystem2 to filesystem and having it populate filesystem2 as a back-compat thing for now.

That said as the author of fs2, I'm 👍  in general. :)

tas50:
@jaymzh The end goal for this plugin and the cloud_v2 plugin is for them to just entirely assume the namespace of the existing plugins. We'll write to both locations with the new data through the Chef 13 release cycle. In Chef 14, the V2 namespaces go away and these just become the one and only filesystem and cloud plugins. It's just a bit staggered to lesson the impact as much as we can. Until we do the final cutover there's not a huge need to rename the plugin file itself.

tas50:
@jaymzh I take the above statement back. We should rename them so the enable/disable config makes sense.

jaymzh:
🥇

tas50:
Thanks again for writing this in the first place @jaymzh. Making it the default is going to prevent some 😢  out of box

PR #976, opened 2017-04-03
Shorten the sbt shellout timeout period

tas50:
30 seconds is way too high here. On my mac the sbt command fails (a whole different issue), but we take 30 seconds to skip the plugin. There’s no reason to wait 30 seconds here for a plugin that doesn’t need that.


lamont-granquist:
is that ever even likely to take a second?

PR #977, opened 2017-04-03
Fix sbt version detection in the scala plugin with modern sbt

tas50:
sbt —version isn’t a command anymore. We need to use the slower about, which also includes horrible color codes, which we have to regex around.


tas50:
Seems to still timeout for some reason on my mac. No sure exactly what's going on here, but I'll track it down eventually

tas50:
Due to sbt downloading the world there's just no way to do this. Existing detection will work for older releases, but unless we find some magic (don't install 100 megs of jar) method we're not getting this back.

thommay:
Let's think about removing the sbt plugin?

PR #991, opened 2017-04-25
Inject sane paths into shell_out

### LINKED ISSUE #990: Ohai should not assume networking commands are in the $PATH
## Description

Ohai network plugin for Linux fails to collect network related attributes
when networking commands like `ip`, `route`, `ifconfig` and `arp` are not in the `$PATH`.

Chef 12 and 13 both have the same issue but it is worse in Chef 13 because 
[it no longer enforces path sanity](https://docs.chef.io/release_notes.html#path-sanity-disabled-by-default-and-modified).

See https://github.com/chef/ohai/blob/master/lib/ohai/plugins/linux/network.rb

## Ohai Version

- 8.23.0 (included in chef-12.19.36)
- 13.0.1 (included in chef-13.0.118)

## Platform Version

- CentOS 6.8
- CentOS 7.3

## Ohai Output

```bash
$ PATH=/usr/local/bin:/usr/bin ohai > /dev/null
[2017-04-23T13:24:50+00:00] INFO: The plugin path /etc/chef/ohai/plugins does not exist. Skipping...
[2017-04-23T13:24:50+00:00] WARN: Plugin Network: unable to detect ipaddress
```

akitada:
### Description

Ohai network plugin for Linux fails to collect network related attributes
when networking commands like ip, route, ifconfig and arp are not in the $PATH.

### Issues Resolved

#990

### Check List

- [x] All tests pass

### DIFF
--- a/lib/ohai/mixin/command.rb
+++ b/lib/ohai/mixin/command.rb
@@ -27,8 +27,13 @@ module Ohai
       # DISCLAIMER: Logging only works in the context of a plugin!!
       # accept a command and any of the mixlib-shellout options
       def shell_out(cmd, **options)
+        options = options.dup
         # unless specified by the caller timeout after 30 seconds
         options[:timeout] ||= 30
+        unless RUBY_PLATFORM =~ /mswin|mingw32|windows/
+          options[:env] = options.key?(:env) ? options[:env].dup : {}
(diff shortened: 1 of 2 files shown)


tas50:
@akitada Thanks for taking a stab at this. You'll need to make sure the which command in the specs actually returns the expected data. It works on *nix systems that actually have those commands, but the specs blow up on Windows systems where we're autofilling "false"

akitada:
@tas50 Updated patch to inject sane paths, as in shell_out in Chef.

tas50:
Thanks for putting this together

Issue #994, opened 2017-04-27
ec2_metadata plugin uses way too much tcp sessions to fetch metadata


ViktorSchlaffer:
## Description

The ec2 metadata plugin in Chef/Ohai performs http requests to fetch ec2 metadata. The Openstack plugin of ohai uses the EC2 compatibility api of Openstack to fetch metadata. The way it is done right now in the ec2_metadata plugin is to create a new tcp connection for each http request, which is inefficient and does not scale well in Openstack environments.[1]

https://github.com/chef/ohai/blob/master/lib/ohai/plugins/openstack.rb
https://github.com/chef/ohai/blob/master/lib/ohai/mixin/ec2_metadata.rb

Proposal: 
There are a couple of alternatives to improve this behavior:
1. rewrite the openstack plugin, to use the config-drive (when available) to fetch metadata, and fall-back to the metadata server only if needed.
2. modify the ec2_metadata.rb plugin to reuse the tcp connection, instead of opening x parallel tcp connection for x http request.

The latter one would benefit both AWS and Openstack deployments.

[1]
In Openstack, all instances on a tenant network talk to the same metadata server. With the current behavior the ec2 plugin opens 20+ tcp connections (depends on the amount of metadata), as each http request is made in a separate session, and it leads to starvation very quickly, as the connection pool size of the metadata server is limited. Though the connection pool can be increased, it has it's impact on the rest of the infrastructure, and makes it very inconvenient to run hundreds of Chef instances on the same tenant network. Reducing the nr. of parallel tcp connection used by the ohai plugin to fetch the metadata would eliminate the scalability issue.

## Ohai Version

8.20.0

## Platform Version

Centos 6 / 7

ViktorSchlaffer:
Solution provided in #995  resolves the problem, so this issue can be closed, once that solution is merged.

PR #995, opened 2017-04-28
Improvements to EC2 metadata handling

tas50:
There's a few changes here so far.

Better logging when we reject a version that Amazon says they support
Drop our timeouts to a more reasonable time. These APIs are crazy fast there's no need to wait 30 seconds
Only setup the http client object once and use methods that allow connection reuse
Don't parse the best API version 3 times
Add all current EC2 metadata versions

### DIFF
--- a/lib/ohai/mixin/ec2_metadata.rb
+++ b/lib/ohai/mixin/ec2_metadata.rb
@@ -71,7 +71,10 @@ module Ohai
       end
 
       def http_client
-        Net::HTTP.start(EC2_METADATA_ADDR).tap { |h| h.read_timeout = 30 }
+        @conn ||= Net::HTTP.start(EC2_METADATA_ADDR).tap do |h|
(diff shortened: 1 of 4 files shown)


tas50:
@ViktorSchlaffer This should resolve #994 by only creating a single net/http object and using the appropriate methods that allow connection reuse within net/http. I'm not able to verify the change using AWS however since they only support http 1.0 w/o keepalives. Any chance you can pull this branch down somewhere and let me know if you see a reduction in connections?

ViktorSchlaffer:
@tas50 I've verified you changes in our Openstack environment, and now the ec2 plugin uses only one tcp connection for the http requests. I have not been able to deploy this on all of our instances, as that decision is not in my hands, but from a functional point of view, this is what I was looking for, so I expect, that it will resolve the connection pool starvation issue.
Thanks for your help Tim!

PR #1001, opened 2017-05-15
Update zpools plugin to work on Linux/BSD

tas50:
This plugin works on Linux/BSD for the most part. Let's do the extra work to get there.

Caveats:

The regex is still less than ideal here. Since zpool status for a zpool lists the zlevel and the volume itself along with the drives it's a bit tricky to extract just the drive information. We're relying on the starting characters to identify drives. So we're saying a line that starts with C is a drive on solaris, SD on Linux, and DA/AD on BSD. What happens if my volume is called "crap"? Well that would show up as a drive. That bug has always existed and this doesn't change that. We can use -P on Linux to get the full path which fixes this for Linux, but that doesn't exist on Solaris/BSD (of course).

### DIFF
--- a/lib/ohai/plugins/solaris2/zpools.rb
+++ b/lib/ohai/plugins/solaris2/zpools.rb
@@ -1,6 +1,6 @@
 #
 # Author:: Jason J. W. Williams (williamsjj@digitar.com)
-# Copyright:: Copyright (c) 2011-2016 Chef Software, Inc.
+# Copyright:: Copyright (c) 2011-2017 Chef Software, Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


tas50:
{
  "storage": {
    "pool_size": "7.25T",
    "pool_allocated": "1.20M",
    "pool_free": "7.25T",
    "capacity_used": "0%",
    "dedup_factor": "1.00x",
    "health": "ONLINE",
    "zpool_version": null,
    "devices": {
      "sda": {
        "state": "ONLINE",
        "errors": {
          "read": "0",
          "write": "0",
          "checksum": "0"
        }
      },
      "sdb": {
        "state": "ONLINE",
        "errors": {
          "read": "0",
          "write": "0",
          "checksum": "0"
        }
      },
      "sde": {
        "state": "ONLINE",
        "errors": {
          "read": "0",
          "write": "0",
          "checksum": "0"
        }
      },
      "sdf": {
        "state": "ONLINE",
        "errors": {
          "read": "0",
          "write": "0",
          "checksum": "0"
        }
      }
    }
  }
}

tas50:
so I actually think we should hold off on this because it won't work outside trivial ATA use cases. I'm going to fix that regex to work evreywhere, which will also allow it to work on FreeBSD.

tas50:
I'm merging this and I'm going to continue to gather zpool status example outputs to see if we can just rely on the spacing. If so we can eliminate the longstanding bugs and some of the awfulness of this regex.

PR #1010, opened 2017-06-07
cpu: add support for arm64 cpuinfo fields

davide125:

### Description

On arm64 the format of `/proc/cpuinfo` is different. This change adds basic support to expose `features` and `bogomips`, which are the most useful fields. Note that these fields are per-processor, while on S390 they're only set once, so we can't really reuse that logic.

### Issues Resolved

None that I could find, though I can file one if needed

### Check List

- [X] All tests pass
- [] RELEASE\_NOTES.md, has been updated if required (not required for bugfixes, required for API changes)


tas50:
Looks good to me, but you have a chefstyle warning.

PR #1014, opened 2017-06-13
Make filesystem more resilient

jaymzh:
### Description

Today lsblk and blkid each have their various shortcomings. Various hacks
and workarounds have been added and changed as we migrated to using lsblk
over blkid, but we can do better.

We run over both and combine the data.

This was caught when we noticed that somne of our AOE-type filesystems were
not recognized by lsblk when not mounted, but blkid did notice them. This meant
we thought the devices were not formatted and wanted to format them. That's bad.
Mega bad.

### Issues Resolved

Missing FS information

### Check List

- [X] All tests pass

### DIFF
--- a/lib/ohai/plugins/linux/filesystem.rb
+++ b/lib/ohai/plugins/linux/filesystem.rb
@@ -30,8 +30,9 @@ Ohai.plugin(:Filesystem) do
     name
   end
 
-  def parse_line(line, have_lsblk)
-    if have_lsblk
+  def parse_line(line, cmdtype)
+    case cmdtype
+    when "lsblk"
       regex = /NAME="(\S+).*?" UUID="(\S*)" LABEL="(\S*)" FSTYPE="(\S*)"/
       if line =~ regex
         dev = $1
@@ -41,7 +42,7 @@ Ohai.plugin(:Filesystem) do
         fs_type = $4
         return { :dev => dev, :uuid => uuid, :label => label, :fs_type => fs_type }
       end
-    else
+    when "blkid"
       bits = line.split
       dev = bits.shift.split(":")[0]
       f = { :dev => dev }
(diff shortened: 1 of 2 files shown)


jaymzh:
CC @chef/client-core

jaymzh:
ping?

lib/ohai/plugins/linux/filesystem.rb
```diff
@@ -142,36 +143,48 @@ def generate_mountpoint_view(fs)
...
     end
 
-    have_lsblk = File.exist?("/bin/lsblk")
+    # We used to try to decide if we wanted to run lsblk or blkid
+    # but they each have a variety of cases were they fail to report
+    # data. For example, there are a variety of cases where lsblk won't
+    # report unmounted filesystems, but blkid will. And vise-versa. Sweet.
+    # So for reliability, we'll run both, if we have them.
+
+    have_lsblk = File.exist?("/bin/lsblk") || File.exist?("/usr/bin/lsblk")
```
coderanger: If we're expanding the search path, should this use one of the various incarnations of `which()`?

lib/ohai/plugins/linux/filesystem.rb
```diff
@@ -142,36 +143,48 @@ def generate_mountpoint_view(fs)
...
-      parsed = parse_line(line, have_lsblk)
-      next if parsed.nil?
-      # lsblk lists each device once, so we need to update all entries
-      # in the hash that are related to this device
-      keys_to_update = []
-      fs.each_key do |key|
-        keys_to_update << key if key.start_with?("#{parsed[:dev]},")
-      end
+    cmds.each do |cmd|
+      cmdtype = File.basename(cmd.split.first)
```
coderanger: This could be handled more explicitly, like make `cmds` a hash of `command_type => command_string` or something?

lib/ohai/plugins/linux/filesystem.rb
jaymzh: I'm fine with that. Will update.

lib/ohai/plugins/linux/filesystem.rb
jaymzh: Um. Sure? I kinda prefer having it tied to the actual command rather than people having to keep things in sync by typing twice.

jaymzh:
Updated to use `which`. Keeping dynamic mapping, unless there's strong opposition.

jaymzh:
one more ping :)

tas50:
Looking forward to resolving some odd failures with this one. The switch to which alone is going to fix some issues.

PR #1016, opened 2017-06-17
fix AIX rpm data

### LINKED ISSUE #1015: AIX RPM packages have the version embedded in the package name
## Description

Ohai on AIX uses "lslpp -L -q -c" to capture package information on all package types, including LPP and RPM.

The LPPs are stored as expected:
    "bos.rte": {
      "version": "7.1.3.45"
    },
    
RPMs are stored with the version embedded in the package name:
    "wget-1.9.1-3": {
      "version": "1.9.1-3"
    },

In order to check the version attribute of the wget RPM, the version would need to be known ahead of time or additional parsing would be needed to check if wget is installed.  This is happening because the lslpp command being used outputs in the following format:

#Package Name:Fileset:Level:State:PTF Id:Fix State:Type:Description:Destination Dir.:Uninstaller:Message Catalog:Message Set:Message Number:Parent:Automatic:

The issue is in https://github.com/chef/ohai/blob/master/lib/ohai/plugins/packages.rb.
For LPP packages, the 2nd field containing the fileset name is used and the 1st field is discarded.  For RPMs, the 2nd field should not be used and the 1st field containing the package name should be used.

## Ohai Version

13.1.0

## Platform Version

AIX

msgarbossa:
Signed-off-by: Mike Sgarbossa michael.sgarbossa@cvshealth.com

### Description

Uses package type field from lslpp command to determine the correct field to use for the package name.  The 1st field needs to be used for RPM packages listed with lslpp -Lqc.

### Issues Resolved
#1015

### DIFF
--- a/lib/ohai/plugins/packages.rb
+++ b/lib/ohai/plugins/packages.rb
@@ -94,8 +94,14 @@ Ohai.plugin(:Packages) do
     # Package Name:Fileset:Level
     # On aix, filesets are packages and levels are versions
     pkgs.each do |pkg|
-      _, name, version = pkg.split(":")
-      packages[name] = { "version" => version }
+      name, fileset, version, _, _, _, pkg_type = pkg.split(":")
+      if pkg_type == 'R'
+        # RPM
+        packages[name] = { "version" => version }
+      else
+        # LPP
+        packages[fileset] = { "version" => version }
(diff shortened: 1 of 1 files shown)


tas50:
Closing this in favor of #1020

PR #1028, opened 2017-07-07
Fix solaris2 ipaddress attribute search

jaymalasinha:
```# ohai ipaddress ``` returns following warnings and replaces the node IP address with a blank IP
after chef-client run
```
[2017-07-07T20:36:32+00:00] WARN: [inet] no ip address on net0
[2017-07-07T20:36:32+00:00] WARN: unable to detect ipaddress
/opt/chef/embedded/lib/ruby/gems/2.1.0/gems/ohai-8.19.1/lib/ohai/system.rb:193:in `attributes_print': I cannot find an attribute named ipaddress! (ArgumentError)
```

Tested this fix on solaris11 client node and verfied IP address was retained after ccr.
Signed-off-by: Jaymala Sinha <jsinha@chef.io>

### Description

[Please describe what this change achieves]

### Issues Resolved

[List any existing issues this PR resolves, or any Discourse or
StackOverflow discussions that are relevant]

### Check List

- [x] All tests pass

### DIFF
--- a/lib/ohai/plugins/solaris2/network.rb
+++ b/lib/ohai/plugins/solaris2/network.rb
@@ -101,7 +101,7 @@ Ohai.plugin(:Network) do
     cint = nil
 
     so.stdout.lines do |line|
-      # if line =~ /^([0-9a-zA-Z\.\:\-]+)\S/
+      # regex: http://rubular.com/r/Iag7JLVTVe
       if line =~ /^([0-9a-zA-Z\.\:\-]+): \S+ mtu (\d+) index (\d+)/
         cint = $1
         iface[cint] = Mash.new unless iface[cint]
--- a/lib/ohai/plugins/solaris2/network.rb
+++ b/lib/ohai/plugins/solaris2/network.rb
@@ -101,7 +101,8 @@ Ohai.plugin(:Network) do
     cint = nil
 
     so.stdout.lines do |line|
-      if line =~ /^([0-9a-zA-Z\.\:\-]+)\S/
+      # if line =~ /^([0-9a-zA-Z\.\:\-]+)\S/
+      if line =~ /^([0-9a-zA-Z\.\:\-]+): \S+ mtu (\d+) index (\d+)/
         cint = $1
         iface[cint] = Mash.new unless iface[cint]
         iface[cint][:mtu] = $2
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by jaymalasinha** [comment]
@chef/engineering-services @scotthain

File: lib/ohai/plugins/solaris2/network.rb [hunk 1]
```diff
@@ -101,7 +101,8 @@ def full_interface_name(iface, part_name, index)
     cint = nil
 
     so.stdout.lines do |line|
-      if line =~ /^([0-9a-zA-Z\.\:\-]+)\S/
+      # if line =~ /^([0-9a-zA-Z\.\:\-]+)\S/
+      if line =~ /^([0-9a-zA-Z\.\:\-]+): \S+ mtu (\d+) index (\d+)/
```
scotthain: Like we discussed - need to update this regex to allow for but not require the last :

tas50: I find linking to a rubular gisst to be really helpful with ohai regex usage. It's really hard to tell what these are attempting to detect without one. Here's an example: https://github.com/chef/ohai/blob/484d8b538d547096ed4472ca5a795d21d0d902db/lib/ohai/plugins/zpools.rb#L69

lib/ohai/plugins/solaris2/network.rb
```diff
@@ -101,7 +101,8 @@ def full_interface_name(iface, part_name, index)
     cint = nil
 
     so.stdout.lines do |line|
-      if line =~ /^([0-9a-zA-Z\.\:\-]+)\S/
+      # if line =~ /^([0-9a-zA-Z\.\:\-]+)\S/
```
thommay: please remove the commented out line before merge

lib/ohai/plugins/solaris2/network.rb
tduffield: That's actually a really good idea. I'm going to start doing that.

jaymalasinha: Thanks.  Will include the link in the update.   
While looking at the history of this change it seems like the +mtu and +index was unintentionally removed while fixing the regex.

jaymalasinha:
Verified the changes are working on Solaris 10 and Solaris 11 both sparc and i86pc platforms

rhass:
Travis seems to be getting stuck trying to resolve gem dependencies. I tried kicking the build again just in case there was an issue in the downloading of the gems, but it did the same thing.

tas50:
This PR was started off a pretty stale master. We aren't testing 2.1 ruby anymore and we don't support it so this shouldn't be an issue to merge.

tas50:
Well apparently I missed the part where this was going into the 8-stable branch and not master. Was that intentional @jaymalasinha? It looks like we'd want this same change on master as well.

jaymalasinha:
@tas50  yes we need it in chef12, I will also cherry-pick it to master now.

PR #1029, opened 2017-07-11
Update Travis config

tas50:


.travis.yml
```diff
@@ -11,15 +12,15 @@ branches:
 before_install:
   - gem update --system
   - gem --version
-  - gem uninstall bundler -a -x -I
```
tas50: This fails in the trusty image

PR #1035, opened 2017-07-19
Detect F5 Big-IPs as platform bigip

tas50:
They're RHEL + a bunch of secret sauce. Put them in the RHEL platform family, detect the platform properly, and parse the platform version so it doesn't get the RHEL platform version instead.

### DIFF
--- a/lib/ohai/plugins/linux/platform.rb
+++ b/lib/ohai/plugins/linux/platform.rb
@@ -86,6 +86,19 @@ Ohai.plugin(:Platform) do
   end
 
   #
+  # Determines the platform version for F5 Big-IP systems
+  #
+  # @returns [String] bigip Linux version from /etc/f5-release
+  #
+  def bigip_version
+    release_contents = File.read("/etc/f5-release")
+    release_contents.match(/BIG-IP release (\S*)/)[1] # http://rubular.com/r/O8nlrBVqSb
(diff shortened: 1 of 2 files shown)


lamont-granquist:
angrysauce CI tests

tas50:
Oh I knew that would blow up. If there's no objections to this PR I'll go fix the specs

lib/ohai/plugins/linux/platform.rb
```diff
@@ -86,6 +86,19 @@ def cumulus_version
...
 
   #
+  # Determines the platform version for F5 Big-IP systems
+  #
+  # @returns [String] bigip Linux version from /etc/f5-release
+  #
+  def bigip_version
+    release_contents = File.read("/etc/f5-release")
+    release_contents.match(/BIG-IP release (\S*)/)[1] # http://rubular.com/r/O8nlrBVqSb
+  rescue NoMethodError, Errno::ENOENT, Errno::EACCES # rescue regex failure, file missing, or permission denied
```
stevendanna: An explanation since it took me a moment: The NoMethodError catches the regex failure because String#match returns `nil` if there was no match and `[]`  isn't a method on `nil`.

stevendanna:
Or just 1! :D

PR #1038, opened 2017-07-31
Handle situations where /proc/cpuinfo lacks core data

tas50:
This works around failures to collect core data on SLES <= 12 SP1. We need to support SLES and core counts are used in several of our product cookbooks.


tas50:
So I should add that this is technically a breaking change. When we ran on older systems previously we delivered completely bogus CPU info. Our tests even accounted for that. Now if we don't have count information we omit it vs. setting it to 0. The hope here is though that we run on these systems (like SLES) and fall back to lscpu, but if lscpu isn't there you're getting a missing value now. I think that's the right thing to do vs. falling back to 0s.

PR #1039, opened 2017-07-31
Added Devuan Compatibility

alzuin:
### Description
Added compatibility for Devuan distribution 

### Issues Resolved
Devuan is debian without systemd. As a complete fork of Debiain, the version is inside /etc/devuan_version rather than /etc/debian_version 

### Check List

- [ ] All tests pass
- [ ] All commits have been signed-off for the Developer Certificate of Origin. See <https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco>

### DIFF
--- a/lib/ohai/plugins/linux/platform.rb
+++ b/lib/ohai/plugins/linux/platform.rb
@@ -177,6 +177,9 @@ Ohai.plugin(:Platform) do
         end
         platform_version debian_platform_version
       end
+    elsif File.exist?("/etc/devuan_version")
+      platform "debian"
+      platform_version File.read("/etc/devuan_version").chomp
     elsif File.exist?("/etc/parallels-release")
       contents = File.read("/etc/parallels-release").chomp
       platform get_redhatish_platform(contents)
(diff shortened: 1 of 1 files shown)


tas50:
Sorry for the late response to this PR. At this point we’re trying to be more picky about the platforms we detect within Ohai. There are seemingly endless numbers of distros and each distro we add is yet another file op we have to perform during every Ohai / Chef run regardless of the platform ohai is running on. These additional platforms add latency to Ohai runs on IO starved systems. In order to determine when a system should be detected in Ohai we’re implementing these rules for platform inclusion:

- Platform is a server distribution (fedora/ubuntu/centos) or network device distribution (bigip, arista_eos)
- Distro has been existed for at least 6 months
- Distro is actively maintained by an organization (company/foundation) and not just a single individual
- Distro is continuously within the top 20 in Distrowatch if it’s a non-network distro

I don’t believe this platform meets these requirements so I’m going to close it out at this point.

It is however possible for you to ship this change to your local nodes and properly detect your distribution.  You can copy the platform plugin you modified and add it to /etc/chef/ohai/plugins/ where it will be loaded and used.

PR #1042, opened 2017-08-06
Add Arch Linux support to package plugin

oslerw:
### Description

This commit adds support for detecting packages on Arch Linux. This is my first time contributing to Ohai or any Chef project, so I'd appreciate any feedback.

### Issues Resolved

I wrote this mainly because I wanted to make some decisions in Chef that involved knowing if one package from a set were already installed. Specifically, I wanted a recipe that installs command line vim, but only if graphical vim isn't already installed.

### Check List

- [x] All tests pass

### DIFF
--- a/lib/ohai/plugins/packages.rb
+++ b/lib/ohai/plugins/packages.rb
@@ -32,7 +32,8 @@ Ohai.plugin(:Packages) do
 
   collect_data(:linux) do
     packages Mash.new
-    if %w{debian}.include? platform_family
+    case platform_family
+    when "debian"
       format = '${Package}\t${Version}\t${Architecture}\n'
       so = shell_out("dpkg-query -W -f='#{format}'")
       pkgs = so.stdout.lines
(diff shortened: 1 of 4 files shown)


oslerw:
Dang, I wasn't aware of chefstyle. Let me make some changes and try again.

lib/ohai/plugins/packages.rb
```diff
@@ -31,6 +31,7 @@
   end
 
   collect_data(:linux) do
+    require "date"
```
tas50: Can you move this into the Arch section since it's not needed for the other platforms

lib/ohai/plugins/packages.rb
```diff
@@ -31,6 +31,7 @@
...
     packages Mash.new
     if %w{debian}.include? platform_family
```
tas50: Any chance you can change these platform_family checks into a case statement. At 3 different checks of the same data it would be faster to use a single check in a case statement

tas50:
Overall this looks pretty solid. Thanks for putting it together. Just fix those 2 comments and this is good to go.

tas50:
Looks good

PR #1043, opened 2017-08-07
bumping timeout for lsblk and blkid to 60s

rmcleod8:
### Description

This change increases the default timeout for `lsblk` and `blkid` from 30 seconds to 60 seconds

### Issues Resolved

Ohai was not returning filesystem info for certain machines. After some research/debugging we determined that machines with large amounts of attached LUNs will take longer than 30 seconds to respond with output from `lsblk` and `blkid`. The machines in question here have >1000 LUNs. Bumping the timeout here will ohai to collect the data successfully for machines with these configurations. 

### Check List

- [x] All tests pass


tas50:
I wish we didn't have to do this, but I can see plenty of scenarios where we would timeout

Issue #1011, opened 2017-06-07
Filesystem plugin fails when missing command(s)


ant1441:
## Description

When running Ohai on a minimal system lacking the required commands used by the 'filesystem' plugin, no output is produced.
  
For example, (on linux) if the `mount` command is missing, the `shell_out` function will fail at [filesystem.rb#L133](https://github.com/chef/ohai/blob/03a9cf30487f4ef281fb46a7e5c266fd778a0f93/lib/ohai/plugins/linux/filesystem.rb#L133), resulting in the `ohai filesystem` command giving an error like 

    /usr/local/rvm/gems/ruby-2.4.0/gems/ohai-13.1.0/lib/ohai/system.rb:211:in `attributes_print': I cannot find an attribute named filesystem! (ArgumentError)

I believe that Ohai can still give valuable information collected by the other available commands if some are not available. For example if I alias `mount` and `blkid` to `true` in the centos:6 Docker container I still get useful output via the `df` commands.
If not, it may be valuable to report the lack of suitable commands to the user, to allow simpler resolution without requiring the Ohai debug output.

I originally encountered this issue running a Kitchen test suite under the dokken provisioner, where the `node["filesystem"]` attribute is missing due to this issue.

The following sequence of commands are suitable to reproduce this issue:

```
docker run -it centos:6 /bin/bash
command curl -sSL https://rvm.io/mpapis.asc | gpg2 --import -
curl -L get.rvm.io | bash -s stable
source /etc/profile.d/rvm.sh
rvm install ruby
gem install ohai
ohai filesystem -l debug
```

## Ohai Version

    [root@89d17baf2cea /]# ohai --version
    Ohai: 13.1.0

## Platform Version

cento6 (Specifically, the centos:6 Docker image - with image id 27f7cebb4137)

## Ohai Output

```
[root@89d17baf2cea /]# ohai filesystem -l debug
[2017-06-07T15:19:13+00:00] DEBUG: Running Ohai with the following configuration: {:log_location=>#<IO:<STDERR>>, :log_level=>:debug, :invoked_from_cli=>true}
[2017-06-07T15:19:13+00:00] DEBUG: Searching for Ohai plugins in /usr/local/rvm/gems/ruby-2.4.0/gems/ohai-13.1.0/lib/ohai/plugins
[2017-06-07T15:19:13+00:00] INFO: The plugin path /etc/chef/ohai/plugins does not exist. Skipping...
[...]

PR #1047, opened 2017-08-23
Add error handling in Linux filesystem plugin

jeunito:
### Description

Add error handling in Linux filesystem plugin for when a command does not exist. 

### Issues Resolved

https://github.com/chef/ohai/issues/1011

- [x] All tests pass
- [x] All commits have been signed-off for the Developer Certificate of Origin.

### DIFF
--- a/lib/ohai/plugins/linux/filesystem.rb
+++ b/lib/ohai/plugins/linux/filesystem.rb
@@ -95,7 +95,7 @@ Ohai.plugin(:Filesystem) do
     fs = Mash.new
 
     # Grab filesystem data from df
-    if which("df")
+    begin
       so = shell_out("df -P")
       so.stdout.each_line do |line|
         case line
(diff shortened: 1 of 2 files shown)


tas50:
Instead of performing the which before the shell_out (which does the which again) could you look into rescuing the shell_out exceptions. We do that for a lot of the language plugins. It avoids a pile of disk IO every time we run the plugin.

jeunito:
@tas50 ok. I have changed the implementation

tas50:
Thanks

jeunito:
No problem

jaymzh:
This is actually super terrifying. If `df` stops working for any reason, you get *partial* data which can be lying to you.

I'm not a fan of this. I'm actually going to make a PR to make this optional and default to failing the plugin.

jeunito:
@jaymzh I just fixed the issue but the reporter who wanted this feature is someone else. I have informed that person already in the issue was filed https://github.com/chef/ohai/issues/1011

PR #1052, opened 2017-09-06
Improve detection of Windows EC2 nodes by using UUID information

tas50:
We're currently relying on "amazon" Organization data being present in the Windows installation. This doesn't work if the VM was migrated from VMware and it doesn't work if the org builds their own AMIs that include their organization. Amazon recommends checking UUID data and we're already doing that on Linux. Lets do that on Windows.

### DIFF
--- a/lib/ohai/plugins/ec2.rb
+++ b/lib/ohai/plugins/ec2.rb
@@ -39,6 +39,7 @@ Ohai.plugin(:EC2) do
 
   # look for amazon string in dmi bios data
   # this gets us detection of HVM instances that are within a VPC
+  # @return [Boolean] do we have Amazon DMI data?
   def has_ec2_dmi?
     # detect a version of '4.2.amazon'
     if get_attribute(:dmi, :bios, :all_records, 0, :Version) =~ /amazon/
@@ -52,6 +53,7 @@ Ohai.plugin(:EC2) do
 
   # looks for a xen UUID that starts with ec2
   # this gets us detection of Linux HVM and Paravirt hosts
+  # @return [Boolean] do we have a Xen UUID or not?
   def has_ec2_xen_uuid?
     if ::File.exist?("/sys/hypervisor/uuid")
       if ::File.read("/sys/hypervisor/uuid") =~ /^ec2/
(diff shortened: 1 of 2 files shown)


lib/ohai/plugins/ec2.rb
```diff
@@ -51,9 +51,18 @@ def has_ec2_dmi?
...
-  # this gets us detection of Linux HVM and Paravirt hosts
+  # uses the sys tree on Linux and a WMI query on windows
+  # this gets us detection of HVM and Paravirt hosts
+  # @return [Boolean] do we have a Xen UUID or not?
   def has_ec2_xen_uuid?
-    if ::File.exist?("/sys/hypervisor/uuid")
+    if RUBY_PLATFORM =~ /mswin|mingw32|windows/
+      require "wmi-lite/wmi"
+      wmi = WmiLite::Wmi.new
+      if wmi.query("select uuid from Win32_ComputerSystemProduct")[0]["identifyingnumber"] =~ /^ec2/
```
adamleff: Isn't there a chance that `wmi.query(...)[0]` will return `nil` (in the event of an empty result set) and then you'd be trying to call `:[]` on `NilClass`?

adamleff: Isn't there a chance that wmi.query(...)[0] will return nil (in the event of an empty result set) and then you'd be trying to call :[] on NilClass?

tas50: I was thinking about that after I pushed it up. If we get an empty result set we didn't have a hint and we have no way to know we're on EC2. The plugin will fail and get skipped and the node won't be recognized as EC2, which is actually what we want. Ohai rescues the world (for better or worse) so we do sorta awful things like that throughout.

tas50: Also I'm not sure how we'd not have a UUID, although I suppose anything is possible

tas50:
Appveyor failure is valid. Right now the code is trying to run on windows in Appveyor and failing. Needs some mocking love.

lib/ohai/plugins/ec2.rb
btm: What does `Get-WmiObject Win32_ComputerSystemProduct  | Select-Object IdentifyingNumber, UUID` return on a Windows system in EC2?

You're searching for UUID but actually returning 'IdentifyingNumber' which is typically a serial number. That makes more sense than UUID if we're looking for a string with /^ec2/ in it.

In which case I think you just want this:
```
if wmi.first_of("Win32_ComputerSystemProduct")["identifyingnumber"] =~ /^ec2/
```

tas50: That's a much better query. Thanks

PR #1060, opened 2017-09-19
Detect Rackspace on Windows

tas50:
This uses the same method we previously used on EC2. It won't poll any data since that requires the xen cli tools, but it will show up as being on rackspace, which is an important first step.

### DIFF
--- a/lib/ohai/plugins/rackspace.rb
+++ b/lib/ohai/plugins/rackspace.rb
@@ -43,13 +43,27 @@ Ohai.plugin(:Rackspace) do
     false
   end
 
+  # Checks for the rackspace manufacturer on Windows
+  # === Return
+  # true:: If the rackspace cloud can be identified
+  # false:: Otherwise
+  def has_rackspace_manufacturer?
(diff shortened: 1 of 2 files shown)


lib/ohai/plugins/rackspace.rb
```diff
@@ -43,13 +43,27 @@ def has_rackspace_metadata?
...
+  # true:: If the rackspace cloud can be identified
+  # false:: Otherwise
+  def has_rackspace_manufacturer?
+    return false unless RUBY_PLATFORM =~ /mswin|mingw32|windows/
+    require "wmi-lite/wmi"
+    wmi = WmiLite::Wmi.new
+    if wmi.first_of("Win32_ComputerSystem")["PrimaryOwnerName"] == "Rackspace"
+      Ohai::Log.debug("Plugin Rackspace: has_rackspace_manufacturer? == true")
+      return true
+    end
```
coderanger: Should have a final false for good measure :)

tas50: Nil evals perfectly here and this is the same pattern we're using in all our other methods

PR #1061, opened 2017-09-22
Expedite-ify ohai

thommay:
@chef/engineering-services I'd like to get to the point where we can turn up a jenkins pipeline for ohai, and I think this is the necessary first step. Where do we go from here?


schisamo:
@thommay Looking at the actions you want to take I don't think we need Jenkins in the mix...unless you were hoping to build Omnibus packages (I don't think that is your intention).

adamleff:
@schisamo a Jenkins pipeline would still be necessary for us to be able to do automatic publishing to rubygems the way things currently are, correct? Once the concept of a generic "release" is introduced this gets easier, but we still publish from artifactory to rubygems.

tas50:
I very much want this. Just thought I'd drop that here

thommay:
@schisamo yeah no omnibus packages but @adamleff is correct; we would like to push each version to artifactory's gem repo and then promote to rubygems.

tduffield:
@thom how is the gem built now?

tas50:
@tduffield I rake release from my desktop

schisamo:
@adamleff @thommay We can devise a way to build and publish gems to artifactory.chef.co following a merge (along with automatic version bump) which would set up being able to promote things at a later time. I think right now the best thing to do would be to describe your ideal workflow and @tduffield and I can make that a reality with Expeditor. The reason I say let us do it is it's a new pipeline shape which we could turn into something generally useful across projects.

adamleff:
Yup.. I'd love to do this for Train as well.

thommay:
Here's what I'd like:

Each merged PR generates a version bump and a built gem that is pushed to artifactory. As part of the chef release cycle we promote a version of ohai (ie, push it to rubygems) so that we can then include that in chef.

Here's what I'd love:

each merged PR triggers a test run on every platform that we support chef on, prior to building and releasing a gem to artifactory.

tduffield:
@thommay @adamleff So, I think we need to close this PR as it won't work as you want it to. There's some functionality to Expeditor we need to add, which we can sort out and plan. I'm going to write up a few cards and put them on the JEX board to track.

thommay:
totally fine with that; thanks @tduffield

PR #1064, opened 2017-10-03
Critical Plugins

### LINKED ISSUE #879: make a required/critical plugins list
## Description

(perhaps this should be an RFC... if so, let me know and I'll write one).

It would be really nice if we could have a list of plugins that are required. We currently have a resource early on in our run that checks for `node[:fqdn]`, `node[:filesystem2]`, etc. - and apparently @coderanger  does as well.

Why not have an `ohai.config[:required_plugins]` where if those plugins fail, Ohai `abort`s and causes the chef run to fail. This list can be empty by default preserving compatibility.
## Ohai Version

N/A
## Platform Version

N/A
## Ohai Output

N/A

jaymzh:
### Description

This adds a new Ohai configuration `critical_plugins` which will cause
Ohai to exit (or `fail` if inside of Chef) if the plugins in the list do
not run successfully.

This feature only supports Ohai 7-style named plugins, not the ancient
Ohai 6-style stuff.

Tests to follow, but:

```
$ cat /home/phild/ohai.rb
ohai.critical_plugins << :Filesystem
$ ./bin/ohai -c /home/phild/ohai.rb
[2017-10-03T13:43:14-07:00] INFO: The plugin path /etc/chef/ohai/plugins does not exist. Skipping...
[2017-10-03T13:43:14-07:00] ERROR: The following plugins marked as critical failed: [:Filesystem]
```

And when I run Chef:
```
[2017-10-03T13:44:11-07:00] FATAL: RuntimeError: The following plugins marked as critical failed: [:Filesystem2]. Failing Chef run.
```

### Issues Resolved

This fixes #879.

### Check List

- [x] All tests pass

### DIFF
--- a/lib/ohai/system.rb
+++ b/lib/ohai/system.rb
@@ -47,6 +47,7 @@ module Ohai
       @cli = config[:invoked_from_cli]
       @plugin_path = ""
       @config = config
+      @failed_plugins = []
       reset_system
     end
 
@@ -106,6 +107,16 @@ module Ohai
       rescue Ohai::Exceptions::AttributeNotFound, Ohai::Exceptions::DependencyCycle => e
         Ohai::Log.error("Encountered error while running plugins: #{e.inspect}")
         raise
+      end
+      critical_failed = Ohai::Config.ohai[:critical_plugins] & @runner.failed_plugins
+      unless critical_failed.empty?
+        msg = "The following Ohai plugins marked as critical failed: #{critical_failed}"
+        if @cli
+          Ohai::Log.error(msg)
+          exit(true)
+        else
+          raise Ohai::Exceptions::CriticalPluginFailure, "#{msg}. Failing Chef run."
+        end
       end
 
       # Freeze all strings.
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,3 +1,13 @@
+# Unreleased
+
+### Critical Plugins
+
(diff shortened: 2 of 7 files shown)


lib/ohai/plugins/linux/filesystem.rb
```diff
@@ -92,6 +92,7 @@ def generate_mountpoint_view(fs)
   end
 
   collect_data(:linux) do
+    fail "i iz failing"
```
coderanger: Oops :)

lib/ohai/system.rb
```diff
@@ -107,6 +108,21 @@ def run_plugins(safe = false, attribute_filter = nil)
         Ohai::Log.error("Encountered error while running plugins: #{e.inspect}")
         raise
       end
+      critical_failed = []
+      @runner.failed_plugins.each do |failed|
+        if Ohai::Config.ohai[:critical_plugins].include?(failed)
```
coderanger: You can use a set intersection to do this, `critical_failed = Ohai::Config.ohai[:critical_plugins] & @runner.failed_plugins`.

danielsdeleo:
I like this feature. Would be worth thinking about adding some plugins to the list by default for Chef 14

jaymzh:
OK, review comments, now failing with a unique exception, and a test.

jaymzh:
Updated message for @tas50

jaymzh:
come on Travis, you can do it!

jaymzh:
Travis seems better, so updating this PR with a RELEASE_NOTES.md update to kick travis.

jaymzh:
Fixing lint

PR #1077, opened 2017-10-20
Properly allow ruby-head to fail in Travis

tas50:


lamont-granquist:
we may just want to `|| true` the rvm gem uninstall.  then if travis helpfully ships a new bundler we'll uninstall it.

tas50:
I can add that. I was going to remove the uninstall entirely since we're on a newer travis image, but a true probably future proofs us better.

Issue #1082, opened 2017-11-07
Detect New EC2 Instances


nate:
## Description

There was a [new EC2 instance class announced yesterday](https://aws.amazon.com/about-aws/whats-new/2017/11/introducing-amazon-ec2-c5-instances-the-next-generation-of-compute-optimized-instances/). They apparently use some new hypervisor tech (`C5 instances are built using a new light-weight hypervisor`) that fails Ohai's EC2 detection.

## Ohai Version

8.15.1 and 13.6.0

## Platform Version

Amazon Linux AMI release 2017.03

## Ohai Output

8.15.1:

```
[2017-11-07T19:14:36+00:00] DEBUG: ec2 plugin: has_ec2_xen_uuid? == false
[2017-11-07T19:14:36+00:00] DEBUG: ec2 plugin: has_ec2_dmi? == false
[2017-11-07T19:14:36+00:00] DEBUG: ec2 plugin: has_amazon_org? == false
[2017-11-07T19:14:36+00:00] DEBUG: ec2 plugin: looks_like_ec2? == false
```

13.6.0:

```
[2017-11-07T19:52:11+00:00] DEBUG: Plugin EC2: has_ec2_xen_uuid? == false
[2017-11-07T19:52:11+00:00] DEBUG: Plugin EC2: has_ec2_dmi? == false
[2017-11-07T19:52:11+00:00] DEBUG: Plugin EC2: has_ec2_identifying_number? == false
[2017-11-07T19:52:11+00:00] DEBUG: Plugin EC2: looks_like_ec2? == false
```

nate:
`/sys/class/dmi/id/` has some things that might be useful for detecting EC2. 

```
irb(main):001:0> File.read('/sys/class/dmi/id/bios_vendor')
=> "Amazon EC2\n"
irb(main):002:0> File.read('/sys/class/dmi/id/sys_vendor')
=> "Amazon EC2\n"
irb(main):003:0> File.read('/sys/class/dmi/id/product_name')
=> "c5.9xlarge\n"
```

nate:
Although, I'm not sure if these are specific to Amazon Linux or not...

tas50:
We have some new options for detecting the instances. I'm taking a look at it now.

nate:
Thanks!

tas50:
I need to redo all the tests, but this fixes the detection issue: https://github.com/chef/ohai/pull/1083

mswilson:
I left this same comment in #1083 

[...]

PR #1083, opened 2017-11-07
Detect new Amazon hypervisor used by the C5 instances

tas50:
C5 instances are not Xen based. They use Amazon's own hypervisor so our existing detection methods fail. I suspect we could probably get rid of the old DMI detection, but I get nervous removing detection logic. This checks the bios vendor DMI data. I also removed the depedency on  the DMI plugin since a lot of systems don't have dmidecode. This will catch far more systems.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ branches:
     - 8-stable
 
 before_install:
-  - gem update --system
+  #  - gem update --system
   - gem --version
   - rvm @global do gem uninstall bundler -a -x -I || true
   - gem install bundler
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,6 @@ branches:
     - 8-stable
 
 before_install:
-  #  - gem update --system
   - gem --version
   - rvm @global do gem uninstall bundler -a -x -I || true
   - gem install bundler
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -20,7 +20,6 @@ install:
   - SET PATH=C:\Ruby%ruby_version%\bin;%PATH%
   - echo %PATH%
   - ruby --version
-    #  - gem update --system
   - gem --version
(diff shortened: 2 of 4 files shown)


tas50:
Fixes https://github.com/chef/ohai/issues/1082

.travis.yml
```diff
@@ -10,7 +10,7 @@ branches:
     - 8-stable
 
 before_install:
-  - gem update --system
+  #  - gem update --system
```
thommay: let's just delete this

appveyor.yml
```diff
@@ -19,7 +20,7 @@ install:
   - SET PATH=C:\Ruby%ruby_version%\bin;%PATH%
   - echo %PATH%
   - ruby --version
-  - gem update --system
+    #  - gem update --system
```
thommay: and this

mswilson:
This change does not exclusively use the documented detection methods found here:  http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/identify_ec2_instances.html

"Amazon" in a BIOS vendor field is **not** guaranteed to **not** change.

If you are looking for an additional piece of information on which to key EC2 instance detection, the EC2 instance ID is provided as the asset tag of the base board.

> Handle 0x0002, DMI type 2, 15 bytes
   Base Board Information
	Manufacturer: Amazon EC2
	Product Name: Not Specified
	Version: Not Specified
	Serial Number: Not Specified
	Asset Tag: i-05f877841fb9d8b51  <---
	Features: None
	Location In Chassis: Not Specified
	Chassis Handle: 0x0003
	Type: Other
	Contained Object Handles: 0

While this is not currently covered in the documentation, it is something that can be relied upon for the latest generation C5 and future EC2 instances that run on the same technology. Though I don't understand why the "EC2" prefix in the UUID is insufficient.

PR #1087, opened 2017-11-08
Test the 8-stable branch in Appveyor

tas50:
We missed this when we started the stable branch


tas50:
Merging this test only PR since it's nice and green now

PR #1097, opened 2017-12-03
Add .idea to gitignore for Rubymine

tas50:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ coverage/
 .bundle
 *~
 vendor
+.idea


jonlives:
I dunno, on the other hand there are a relatively small number of really commonly used IDEs, and it'll make it easier to avoid accidentally ```git add```'ing them. I don't feel particularly strongly on this though.

tas50:
I'll shove it in my global. I'd say Rubymine is probably pretty rarely used in chef dev

PR #1099, opened 2017-12-03
Remove the rubygems require in the main app

tas50:
This is Ruby 1.8 code and also pretty heavily assumes the the user is installing into a system (debian) ruby install. There's no need for any of this with modern ruby releases.

### DIFF
--- a/bin/ohai
+++ b/bin/ohai
@@ -19,17 +19,8 @@
 # limitations under the License.
 #
 
-begin
-  # if we're in a source code checkout, we want to run the code from that.
-  # have to do this *after* rubygems is loaded.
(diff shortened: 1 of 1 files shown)


bin/ohai
```diff
@@ -19,12 +19,6 @@
 # limitations under the License.
 #
 
-begin
-  require "rubygems"
-rescue LoadError
-  # must be debian! ;)
-  missing_rubygems = true
```
thommay: we still check `missing_rubygems` below (line 34)

tas50: 🤕 Fixed up now

Issue #1092, opened 2017-11-17
FIPS support of shard plugin


sonic0002:
## Description

Should we change the digest algorithm from MD5 to something else to support FIPS?

## Ohai Version

ohai-8.25.0

## Platform Version

All Linux

## Ohai Output

Initially, the Chef client run output with FIPS enabled (**fips true** in **client.rb**) is:

> Starting Chef Client, version 12.12.13
> OpenSSL FIPS 140 mode enabled
> md5_dgst.c(75): OpenSSL internal error, assertion failed: Low level API call to digest MD5 forbidden in FIPS mode!  

Then we noticed that **shard** is a built-in plugin of Chef and it uses MD5 to generate the seed and it failed the OpenSSL assertion of FIPS compatable algorithms.

tas50:
We've marked the shard plugin as optional which means out the box it won't load and this resolves the FIPS warnings

PR #1106, opened 2017-12-21
Ohai 14: Use SHA256 for the shard seed for FIPS compliance

tas50:
MD5 is not allowed in FIPS140/180. We can use SHA256 and then this
passes FIPS and all the government users can use this plugin. Yes I
realize we just trim the value anyways. Still can't use it in FIPS.

### DIFF
--- a/lib/ohai/plugins/shard.rb
+++ b/lib/ohai/plugins/shard.rb
@@ -17,7 +17,7 @@
 #
 
 Ohai.plugin(:ShardSeed) do
-  require "digest/md5"
+  require "digest/sha2"
   depends "hostname", "dmi", "machine_id", "machinename"
   provides "shard_seed"
 
@@ -52,7 +52,7 @@ Ohai.plugin(:ShardSeed) do
                 yield(src)
               end
     end
-    shard_seed Digest::MD5.hexdigest(data)[0...7].to_i(16)
+    shard_seed Digest::SHA256.hexdigest(data)[0...7].to_i(16)
   end
 
   collect_data(:darwin) do
--- a/spec/unit/plugins/shard_spec.rb
+++ b/spec/unit/plugins/shard_spec.rb
@@ -16,7 +16,7 @@
 # limitations under the License.
 #
 
-require "digest/md5"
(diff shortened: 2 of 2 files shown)


tas50:
Fixes https://github.com/chef/ohai/issues/1092

tas50:
@jaymzh any issue with switching this to SHA for FIPS?

coderanger:
This would change the seed value for a host so should wait for 14.

coderanger:
@jaymzh Just for the sake of saying it, there is no feature for which we, the maintainers, promise to keep compat on _indefinitely_. Probably shouldn't break it just for this (we can either disable the plugin under fips or make it use sha only under fips) but if FB is relying on this being a truly forever thing, y'all should definitely fork the plugin locally.

coderanger:
Also as mentioned way back when this was added, Ruby has a very fast Adler CRC32 implementation as part of the Zlib binding. Could be used instead of a crypto hash to avoid any such future problems.

jaymzh:
@coderanger I get that, but we're not talking some hard-to-maintain complex code we want to get rid of. We have ways of passing a config to a plugin, there's no reason to ever not let people choose their hash algorithm

lamont-granquist:
FIPS compliance globally disables MD5 and doesn't care if its being used security-sensitive or not, it just isn't allowed.

Seems bad to break backcompat like this, though, we should probably produce shard seeds using different algorithms via specific ohai attributes.   Use `shard_seed256` and just make sure that when MD5 fails on a box with FIPS enabled that it rescues the exception and `shard_seed` is blank.

coderanger:
Using multiple attributes is probably more of a user trap than checking if FIPS is enabled inside the plugin. Though the downside then is switching FIPS on or off will change the seed value, but that doesn't seem unreasonable?

jaymzh:
Agree with @coderanger  - multiple seeds is a bad idea. But FIPS forcing the default to be different is probably fine.

tas50:
We worked around this by marked Shard as an optional plugin in Ohai 14

PR #1107, opened 2017-12-23
Add a plugin to provide PCI bus information using `lspci`.

jaymzh:
### Description

This is based on an original plugin from:
https://github.com/GLOBAL2000/chef-ohai-plugins/

With modifications from the fork here:
https://github.com/GLOBAL2000/chef-ohai-plugins/

### Check List

- [X] All tests pass
- [X] RELEASE\_NOTES.md, has been updated if required (not required for bugfixes, required for API changes)

### DIFF
--- a/lib/ohai/plugins/linux/lspci.rb
+++ b/lib/ohai/plugins/linux/lspci.rb
@@ -1,0 +1,76 @@
+#
+# Author:: Joerg Herzinger <joerg.herzinger@oiml.at>
+# Author:: Phil Dibowitz <phil@ipom.com>
+# Copyright:: Copyright (c) 2011 GLOBAL 2000/Friends of the Earth Austria
+# Copyright:: Copyright (c) 2017 Facebook, Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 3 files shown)


lib/ohai/plugins/linux/lspci.rb
```diff
@@ -0,0 +1,76 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+Ohai.plugin(:Lspci) do
+  depends "platform"
+  provides "pci"
+
+  collect_data(:linux) do
+    $devices = Mash.new
```
adamleff: Can we use an instance variable, or just defining this as a local variable and passing it along to methods that need it, instead of using a global?

lib/ohai/plugins/linux/lspci.rb
```diff
@@ -0,0 +1,76 @@
...
+    lspci = shell_out("lspci -vnnmk")
+
+    $h = /[0-9a-fA-F]/ #any hex digit
```
adamleff: Again re: global variables. Can we use instance variables, or define methods that return these regex patterns?

thommay:
I think we should merge this but add it to the list of plugins that will be disabled by default in chef-14.

jaymzh:
@adamleff  - yeah, dropping the globals is a good call. That's from the original author. I'll go ahead and clean it up.

jaymzh:
rebased and nuked all the globals. Waiting for tests to run.

PR #1121, opened 2018-01-24
build ohai gems automagically

thommay:


.expeditor/config.yml
```diff
@@ -0,0 +1,39 @@
...
+  version_tag_format: "v{{version}}"
+  # The Github Team primarily responsible for handling incoming Pull Requests.
+  maintainer_group: chef/client-maintainers
+  # Which Github branches to build Omnibus releases from, and what versions
+  # (as determined by the value in the VERSION file) those branches are responsible
+  # for building.
+  release_branch:
+    - master:
+        version_constraint: 14.*
+    - 13-stable:
```
tduffield: You'll need to add a config file to the `13-stable` and `8-stable` branches if you want/need Expeditor to do stuff on those branches.

thommay: For now I think master only is ok. we can rethink at need

PR #1127, opened 2018-01-30
Initial removal of support for Ohai V6 plugins

tas50:
This doesn't refactor anything. Just removes the old support in the most brutal way possible. With this in we can start the work to refactor the loader and runner code.


tas50:
I think this may be a MVP of ripping out V6 w/o refactoring any code. I'd appreciate any feedback.

lib/ohai/loader.rb
```diff
@@ -133,11 +124,8 @@ def load_plugin_class(plugin_path, plugin_dir_path = nil)
         load_v7_plugin_class(contents, plugin_path)
       else
         Ohai::Log.warn("[DEPRECATION] Plugin at #{plugin_path} is a version 6 plugin. \
```
thommay: I don't think this is a warn any more, this is a fatal error.

lib/ohai/runner.rb
```diff
@@ -45,8 +45,6 @@ def run_plugin(plugin)
           case plugin.version
           when :version7
             run_v7_plugin(plugin)
-          when :version6
-            run_v6_plugin(plugin)
```
thommay: we'd be better off just raising unless this is a v7 plugin, and ditching this whole case statement i think.

lib/ohai/system.rb
```diff
@@ -87,19 +85,6 @@ def load_plugins
     end
 
     def run_plugins(safe = false, attribute_filter = nil)
```
thommay: this whole function can get cleaned up a lot.

tas50: There's several methods where that's true

PR #1128, opened 2018-01-30
Simplify path logic by requiring chef-config 12.8+

tas50:
We can safely assume chef-config 12.8+ now.

### DIFF
--- a/ohai.gemspec
+++ b/ohai.gemspec
@@ -25,7 +25,7 @@ Gem::Specification.new do |s|
   s.add_dependency "ipaddress"
   s.add_dependency "wmi-lite", "~> 1.0"
   s.add_dependency "ffi", "~> 1.9"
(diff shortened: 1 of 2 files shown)


ohai.gemspec
```diff
@@ -25,7 +25,7 @@ Gem::Specification.new do |s|
   s.add_dependency "ipaddress"
   s.add_dependency "wmi-lite", "~> 1.0"
   s.add_dependency "ffi", "~> 1.9"
-  s.add_dependency "chef-config", ">= 12.5.0.alpha.1", "< 15"
+  s.add_dependency "chef-config", ">= 12.8", "< 15"
```
thommay: just make this `~>` the ohai major version?

tas50: We need it to work with chef-config 14, which isn't out though. It needs 13-14 at a minimum

PR #1131, opened 2018-02-07
account for snap based installations of LXD

thewyzard44:
Signed-off-by: Sean Zachariasen <thewyzard@hotmail.com>

### Description

As of Dec 2017, LXD will no longer be publishing new versions under either backports, or via their own PPA.  All future versions (except for 3.0.x coming in April 2018 with Bionic) will only be available via snapcraft via `snap install lxd` and as such, will have a new installation location.

The old installation location remains valid for base ubuntu installations (including 3.0.x in bionic) as well as past backports on past releases

### Issues Resolved

Allow virtualization 'LXD host' to be detected for non-ubuntu, and old-ubuntu installations done via snap, as well as any future releases of LXD (3.x +) post-bionic

### Check List

- [x] All tests pass


lib/ohai/plugins/linux/virtualization.rb
```diff
@@ -222,10 +222,15 @@ def nova_exists?
       Ohai::Log.debug("Plugin Virtualization: /dev/lxd/sock exists. Detecting as lxd guest")
       virtualization[:system] = "lxd"
       virtualization[:role] = "guest"
-    elsif File.exist?("/var/lib/lxd/devlxd")
-      Ohai::Log.debug("Plugin Virtualization: /var/lib/lxd/devlxd exists. Detecting as lxd host")
-      virtualization[:system] = "lxd"
-      virtualization[:role] = "host"
+    else
+      ["/var/lib/lxd/devlxd", "/var/snap/lxd/common/lxd/devlxd"].each do |devlxd|
```
tas50: Can you put a brief note about what versions we're looking for in each of these locations. It'll help years from now when someone reads this and wonders if we still need to check both locations

thewyzard44: incoming...

it's more about 'how it's installed' as both methods will remain in use for the foreseeable future.  But yea breaking it up by version(s) will lead to a more concise explanation I believe.

thommay:
Thanks @thewyzard44

PR #1134, opened 2018-02-16
fix failing tests on non linux platforms

thommay:


tas50:
This has been on my todo rage fix list. Thanks

PR #1136, opened 2018-02-16
Support optional plugins

thommay:
Some plugins are useful enough to ship in Ohai, but should be off by
default. Those plugins can be (and are) marked as optional. Implements
RFC 103

### DIFF
--- a/lib/ohai/dsl/plugin.rb
+++ b/lib/ohai/dsl/plugin.rb
@@ -32,15 +32,8 @@ module Ohai
       name.is_a?(Symbol) && name.to_s.match(/^[^A-Z]|_/).nil?
     end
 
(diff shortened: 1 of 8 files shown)


lib/ohai/dsl/plugin.rb
```diff
@@ -32,15 +32,8 @@ def self.valid_name?(name)
       name.is_a?(Symbol) && name.to_s.match(/^[^A-Z]|_/).nil?
     end
 
-    # dealing with ruby 1.8
```
tas50: bye bye 1.8

tas50:
Looks pretty straight forward. Lets just make sure we get this added to the Ohai plugin docs so people know the new DSL exists.

PR #1139, opened 2018-02-26
Simplify & optimize the kernel plugin

tas50:
Speed up some of the windows kernel helpers methods and remove the duplicate dragonfly plugin that was just a copy of freebsd.

### DIFF
--- a/lib/ohai/plugins/kernel.rb
+++ b/lib/ohai/plugins/kernel.rb
@@ -59,14 +59,14 @@ Ohai.plugin(:Kernel) do
 
   # windows
   def os_lookup(sys_type)
-    return "Unknown" if sys_type.to_s.eql?("0")
-    return "Other" if sys_type.to_s.eql?("1")
-    return "MSDOS" if sys_type.to_s.eql?("14")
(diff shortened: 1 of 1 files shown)


lib/ohai/plugins/kernel.rb
```diff
@@ -52,21 +52,21 @@ def bsd_modules(path)
...
   def os_lookup(sys_type)
-    return "Unknown" if sys_type.to_s.eql?("0")
-    return "Other" if sys_type.to_s.eql?("1")
-    return "MSDOS" if sys_type.to_s.eql?("14")
-    return "WIN3x" if sys_type.to_s.eql?("15")
-    return "WIN95" if sys_type.to_s.eql?("16")
-    return "WIN98" if sys_type.to_s.eql?("17")
-    return "WINNT" if sys_type.to_s.eql?("18")
-    return "WINCE" if sys_type.to_s.eql?("19")
+    return "WINNT" if sys_type == 18 # most likely so first
```
thommay: why on earth is this not a case statement?

Issue #1054, opened 2017-09-08
Switch to /etc/os-release to identify SLES


tas50:
From the SLES 12 SP3 release notes:

```
9.2.2.2 Use /etc/os-release Instead of /etc/SuSE-release

Starting with SLE 12, the /etc/SuSE-release file has been deprecated. Do not use it to identify a SUSE Linux Enterprise system anymore. This file will be removed in a future Service Pack or release.
To determine the release, use the file /etc/os-release instead. This file is a cross-distribution standard to identify Linux systems. For more information about the syntax, see the os-release man page (man os-release).
```

tas50:
We're currently doing Suse detection here: 

https://github.com/chef/ohai/blob/master/lib/ohai/plugins/linux/platform.rb#L197

We'll need to take the output of /etc/os-release and make sure we don't misdetect as another platform, but overall this should be a fairly straight forward change.

tas50:
/etc/SuSE-release
```
openSUSE 42.1 (x86_64)
VERSION = 42.1
CODENAME = Malachite
```


/etc/os-release
```
NAME="openSUSE Leap"
VERSION="42.1"
VERSION_ID="42.1"
PRETTY_NAME="openSUSE Leap 42.1 (x86_64)"
ID=opensuse
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:opensuse:opensuse:42.1"
BUG_REPORT_URL="https://bugs.opensuse.org"
HOME_URL="https://opensuse.org/"
ID_LIKE="suse"
```

tas50:
@yeoldegrove Did this one for us.

PR #1140, opened 2018-02-26
SUSE: Use /etc/os-release if present for all platform attributes

yeoldegrove:
### Description

Use /etc/os-release (if present) for all platform attributes.

### Issues Resolved

Newer linux distributions all have this file, but might not have the legacy *-release fles anmore.
This is for example true for the upcoming SUSE SLES 15.

#1054 This is also related (SLES12)

### Check List

- [x] All tests pass


yeoldegrove:
@thommay I think you're right.
This is the more conservative approach.

 Use os-release (present on all modern linux distros) BUT use old *-release files as fallback.

os-release will only be used if no other *-release file is present.
We have to do this for compatibility reasons, or older OS releases might get different
"platform" or "platform_version" attributes (e.g. SLES12, RHEL7).

platform_family also does not need to be hardcoded anymore.
This would be the correct way, but we stick with "determine_platform_family" for compatibility reasons.

yeoldegrove:
@tas50 @thommay Is anything missing or can this be merged?

thommay:
@yeoldegrove needs one more reviewer and then it's good to go.

tas50:
Thanks for putting this together and reworking it for us @yeoldegrove

PR #1141, opened 2018-02-26
Don't stacktrace if we can't shell_out to zpool

tas50:

### DIFF
--- a/lib/ohai/plugins/zpools.rb
+++ b/lib/ohai/plugins/zpools.rb
@@ -28,21 +28,26 @@ Ohai.plugin(:Zpools) do
 
   def gather_pool_info
     pools = Mash.new
-    # Grab ZFS zpools overall health and attributes
-    so = shell_out("zpool list -H -o name,size,alloc,free,cap,dedup,health,version")
-    so.stdout.lines do |line|
-      case line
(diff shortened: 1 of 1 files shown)


lib/ohai/plugins/zpools.rb
```diff
@@ -28,21 +28,26 @@ def sanitize_value(value)
...
-        Ohai::Log.debug("Plugin Zpools: Parsing zpool list line: #{line.chomp}")
-        pools[$1] = Mash.new
-        pools[$1][:pool_size] = sanitize_value($2)
-        pools[$1][:pool_allocated] = sanitize_value($3)
-        pools[$1][:pool_free] = sanitize_value($4)
-        pools[$1][:capacity_used] = sanitize_value($5)
-        pools[$1][:dedup_factor] = sanitize_value($6)
-        pools[$1][:health] = sanitize_value($7)
-        pools[$1][:zpool_version] = sanitize_value($8)
+    begin
```
thommay: don't need `begin`. Since you care about returning `pools` you can just
```
ensure
 return pools
end
```

thommay: No, the rescue doesn't change (besides moving left two characters)

tas50:
Merging based on triage go ahead

PR #1143, opened 2018-02-27
WIP: Use let to load plugins and remove "should" from the specs

tas50:
We're far from done here, but this pushes us in the right direction


tas50:
This is going to require the reworking of some of the helpers. Should have it wrapped up today then I can get some windows specs into kernel

PR #1154, opened 2018-03-07
Fix bug in azure plugin/update to recent metadata version.

kriszentner:
### Description

This fixes a bug discovered for the Azure plugin, and updates to the recent azure metadata api date/version.

### Issues Resolved

This fixes a bug I discovered which will crash the azure plugin if it's missing metadata. I discovered, at least on an ND24rs instance I was on, that no network metadata is received.

Also updated to the recent Azure metadata API which gives additional metadata including the resource group name, availability set info, subscription id, and VM scale set info.

### DIFF
--- a/lib/ohai/mixin/azure_metadata.rb
+++ b/lib/ohai/mixin/azure_metadata.rb
@@ -22,7 +22,7 @@ module Ohai
     module AzureMetadata
 
       AZURE_METADATA_ADDR = "169.254.169.254" unless defined?(AZURE_METADATA_ADDR)
-      AZURE_METADATA_URL = "/metadata/instance?api-version=2017-04-02" unless defined?(AZURE_METADATA_URL)
+      AZURE_METADATA_URL = "/metadata/instance?api-version=2017-12-01" unless defined?(AZURE_METADATA_URL)
 
       # fetch the meta content with a timeout and the required header
       def http_get(uri)
--- a/lib/ohai/mixin/azure_metadata.rb
+++ b/lib/ohai/mixin/azure_metadata.rb
@@ -22,7 +22,7 @@ module Ohai
     module AzureMetadata
 
       AZURE_METADATA_ADDR = "169.254.169.254" unless defined?(AZURE_METADATA_ADDR)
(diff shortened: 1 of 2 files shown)


tas50:
@kriszentner Thanks for taking the time to refactor this all. I'd love to get this into Chef 13/14. Can you run `chefstyle -a` to cleanup the style issues and then sign off the commits for the DCO so we can merge it in?

Thanks,
Tim

lib/ohai/mixin/azure_metadata.rb
```diff
@@ -22,7 +22,7 @@ module Mixin
     module AzureMetadata
 
       AZURE_METADATA_ADDR = "169.254.169.254" unless defined?(AZURE_METADATA_ADDR)
-      AZURE_METADATA_URL = "/metadata/instance?api-version=2017-04-02" unless defined?(AZURE_METADATA_URL)
+      AZURE_METADATA_URL = "/metadata/instance?api-version=2017-12-01" unless defined?(AZURE_METADATA_URL)
```
tas50: In the ec2 metadata resource we have some magic that looks for the latest supported metadata from an array of known supported versions. Do you think we should maybe look at doing something like that here or are we guaranteed to always have 2017-12-01?

kriszentner: I looked at the documentation and it seems that they support older versions as they age. I'll have to look into seeing if there's a per instance api version checker. Maybe I can bug someone in Azure. I found 12-01 isn't supported in UK so I backed it off to the one that's supported everywhere.

tas50:
@kriszentner There's a rspec failure if you check the Travis/Appveyor runs

tas50:
This looks good to me. We just need the DCO signoff on the commits so we can accept it. There's details here: https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco

kriszentner:
Thanks for helping me out with this. Learn something new every time when doing these contribs!

tas50:
Thanks for working through it. This is a nice step in the right direction for how we handle this metadata.

PR #1156, opened 2018-03-09
Fix bug in azure plugin/update to recent metadata version. [ backport #1154]

tas50:
### Description

This fixes a bug discovered for the Azure plugin, and updates to the recent azure metadata api date/version.

### Issues Resolved

This fixes a bug I discovered which will crash the azure plugin if it's missing metadata. I discovered, at least on an ND24rs instance I was on, that no network metadata is received.

Also updated to the recent Azure metadata API which gives additional metadata including the resource group name, availability set info, subscription id, and VM scale set info.


tas50:
These all had the DCO sign-off, but not the way I cherry-picked them. This can be verified on master. There was a merge that I didn't want to bring in

PR #1158, opened 2018-03-09
Pin Chefstyle and remove github_changelog_generator

tas50:
Pin chefstyle to 0.6.0 so we don't have to fix all the warnings on our
legacy branch


tas50:
Merging to fix CI

PR #1161, opened 2018-03-14
Move ohai to structured logging

thommay:
This does two things:
it moves away from the old model of using a log module throughout, to injecting the logger in to the various classes. This means that each class can attach metadata to the logger, allowing - in future - us to add tooling to drill into specific plugins or similar.
It converts (the second commit) everything to be at log level trace rather than debug. In general, a cookbook author does not give two hoots about which ohai plugins are loaded or in we failed to load a hint on azure, and we should respect their attention.

### DIFF
--- a/lib/ohai/plugins/mono.rb
+++ b/lib/ohai/plugins/mono.rb
@@ -31,7 +31,7 @@ Ohai.plugin(:Mono) do
       # 	Notification:  kqueue
       # 	Architecture:  amd64
       # 	Disabled:      none
-      # 	Misc:          softdebug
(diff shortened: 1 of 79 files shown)


lib/ohai/plugins/mono.rb
```diff
@@ -31,7 +31,7 @@
       # 	Notification:  kqueue
       # 	Architecture:  amd64
       # 	Disabled:      none
-      # 	Misc:          softdebug
+      # 	Misc:          softtrace
```
tas50: find/replace

PR #1169, opened 2018-03-27
Exclude example output from rubocop rules & remove changelog generator gem

tas50:
This gets our build green again


btm:
new travis job is https://travis-ci.org/chef/ohai/builds/359000671

PR #1170, opened 2018-03-28
SCSI plugin

jaymzh:
### Description

Add a SCSI plugin.

### Issues Resolved

Can know what devices are on the SCSI bus at what address.

### Check List

- [x] New functionality includes tests
- [x] All tests pass
- [ ] RELEASE\_NOTES.md, has been updated if required (not required for bugfixes, required for API changes)
- [x] All commits have been signed-off for the Developer Certificate of Origin. See <https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco>

PR #1171, opened 2018-03-28
SCSI plugin

jaymzh:
### Description

Add a SCSI plugin.

This is a backport of #1170 to Chef 13, it differs only in dropping
the `optional` keyword.

### Check List

- [X] All tests pass

### DIFF
--- a/lib/ohai/plugins/scsi.rb
+++ b/lib/ohai/plugins/scsi.rb
@@ -1,0 +1,51 @@
+#
+# Author:: Phil Dibowitz <phil@ipom.com>
+# Copyright:: Copyright (c) 2018 Facebook, Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


jaymzh:
Heya @tas50  you may be interested. :)

jaymzh:
@tas50 wut? Ohai ignores all failures no matter what. I added 2 new plugins to 12 like 2 months ago. There's no reason this is a breaking change.

tas50:
If this ships in 13.9 and is disabled in 14.1 that's a breaking change between 13 and 14. If we could ship this disabled in 13 I'd be fine with shipping it, but if a user relies on this with 13.9 and then it disappears in 14.1 that's a bad user experience.

jaymzh:
I thought part of 14 was that we were disabling tons of plugins. Should we enable this in 14 then and disable it in 15?

lib/ohai/plugins/scsi.rb
```diff
@@ -0,0 +1,51 @@
...
+
+Ohai.plugin(:Lsscsi) do
+  depends "platform"
+  provides "scsi"
+
+  require "mixlib/shellout"
+
+  collect_data(:linux) do
+    devices = Mash.new
+    lsscsi = shell_out("lsscsi")
```
coderanger: Should probably either use `shell_out!` or have an explicit `return if lsscsi.error?` so we don't try and parse any potential error output.

tas50:
Closing this out at this point based on the Ohai plugins RFC policy changes.

Issue #1174, opened 2018-04-15
Shard plugin probably shouldn't be optional


coderanger:
I think we crossed some wires wrt optional plugins, we advertised that it was for niche-and-problematic plugins like `passwd` but also marked `shard` as optional to avoid problems for FIPS. We should probably make it optional only if FIPS mode is active so that users don't have to jump through a bunch of hoops to use the shard plugin.

tas50:
It's a standard plugin again

PR #1175, opened 2018-04-15
Make the shard plugin work under FIPS by using SHA2 instead of MD5

coderanger:
This makes life easier for most users since there is no reason to disable this plugin by default other than FIPS, and having to set `optional_plugins` is more of a barrier than I think we planned (I couldn't find an implementation of the cookbook-metadata-driven path we had talked about).

### DIFF
--- a/lib/ohai/plugins/shard.rb
+++ b/lib/ohai/plugins/shard.rb
@@ -17,7 +17,7 @@
 #
 
 Ohai.plugin(:ShardSeed) do
-  depends "hostname", "dmi", "machine_id", "machinename", "fips", "hardware"
+  depends "hostname", "dmi", "machine_id", "machinename", "fips", "hardware", "kernel"
   provides "shard_seed"
 
   def get_dmi_property(dmi, thing)
@@ -94,6 +94,8 @@ Ohai.plugin(:ShardSeed) do
       case src
       when :serial
         wmi.first_of("Win32_BIOS")["SerialNumber"]
+      when :os_serial
+        kernel["os_info"]["serial_number"]
       when :uuid
         wmi.first_of("Win32_ComputerSystemProduct")["UUID"]
       else
--- a/lib/ohai/plugins/shard.rb
+++ b/lib/ohai/plugins/shard.rb
@@ -17,10 +17,13 @@
 #
 
 Ohai.plugin(:ShardSeed) do
+  require "openssl"
   require "digest/md5"
   depends "hostname", "dmi", "machine_id", "machinename"
   provides "shard_seed"
-  optional true
+  # Disable this plugin by default under FIPS mode because even though we aren't
+  # using MD5 for cryptography, it will still throw up an error.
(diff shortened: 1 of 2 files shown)


coderanger:
Fixes #1174

coderanger:
I disagree that the level of niche-ness is anywhere close, FIPS-mode is used by only places that actually require it, while shard_seed has been an advertised feature for a long time now. And the improved UX for almost everyone seems like it's worth the minor confusion for the tiny subset of FIPS users.

coderanger:
Summary of a chat on Slack, in an effort to reduce special-case-iness, I'm going to rework this so the plugin isn't disabled for any users, but will switch to SHA2 (or Adler?) if FIPS-mode is active. This means that enabling or disabling FIPS-mode will change the shard seed on all nodes, but since that is expected to happen approximately never, it's the least edge case approach.

coderanger:
I made this configurable for the potential edge case of actually needing to turn FIPS on and off, if you reeeeeally want to, you can force it to SHA2 even without FIPS. Also leaves room to add new stuff if needed.

tas50:
@coderanger You're missing a DCO signoff on a commit here

lib/ohai/plugins/shard.rb
```diff
@@ -34,6 +33,27 @@ def default_sources
     [:machinename, :serial, :uuid]
   end
 
+  def default_digest_algorithm
+    if defined?(OpenSSL.fips_mode) && OpenSSL.fips_mode
```
tas50: Is there a reason you're not just using the data we already gather from the FIPS plugin? We gather data in very specific ways within that plugin per platform.

coderanger: No specific reason, this seemed more accurate. It was only added in Ruby 2.5 which is why I'm checking `defined?` and probably why we didn't use it before :)

coderanger:
Fixed DCO, my script apparently forgot my name and email ¯\_(ツ)_/¯

coderanger:
By the power of scope creep, this now adds support for Windows to the Shard plugin, as well as best-effort support for all other OSes.

spec/unit/plugins/shard_spec.rb
```diff
@@ -26,35 +26,93 @@
...
+      expect(subject).to eq(178738102)
+    end
+  end
+
+  context "with a weird OS" do
+    let(:os) { :aix }
+
+    it "should provide a shard with a default-safe set of sources" do
+      # Note: this is different than the other defaults.
+      expect(subject).to eq(253499154)
```
jaymzh: above you `raise` an error in the default case

jaymzh: Oh, nevermind, only if it's not one of the default ones.

jaymzh:
Thanks for keeping the defaults compatible @coderanger !

coderanger:
@jaymzh With the current defaults baked down to actual ints in the tests now, we should very very notice if we ever break compat :)

PR #1178, opened 2018-05-01
Make the FIPS plugins use the new Ruby 2.5 accessor if present

coderanger:
We need to remember to clean this up when we drop Ruby 2.4.

The global variable for testing is gross, I welcome better ideas :) (we could just only run the pre-2.5 tests when actually running under 2.4 but that seemed risky)


coderanger:
@tyler-ball The accessor method didn't exist before 2.5 from what I can tell :) You used to be able to set the FIPS mode flag from inside Ruby, but not query it.

PR #1181, opened 2018-05-04
[filesystem] Unify plugins, bring BSD into the modern age

### LINKED ISSUE #1180: filesystem plugin never moved to new API on BSD
## Description

We moved the Filesystem plugin to a new better API. In 12 this was called `filesystem2` and both APIs were available. In 13 both `node['filesystem']` and `node['filesystem2']` on use this new API.

... unless you're on BSD. On BSD, `filesystem` is still the old API. :sadtrombone: :cry: 

Actually it may be worse: We implemented Darwin and Linux, and nothing else.

I now have access to a BSD box and will write fs2 for it, but @scotthain probably needs to do the conversion for aix, hpux, irix, and whatever else craziness there is.

## Ohai Version

12, 13 and 14, getting worse as you go

## Platform Version

BSD

jaymzh:
### Description

We never converted anything other than mac or linux because we're clowny, and
now BSD and other Unixes has a totally different API than Mac and Linux.

This creates a single plugin so they can leverage the helper classes, and moves
mac, linux and BSD into it, and adds full support for BSD.

This starts to fix #1180

### Check List

- [X] All tests pass


jaymzh:
I didn't do release notes because it's no longer ordered by release and so I don't know how...

thommay:
I'd honestly rather have kept them as separate files but I'm not sure it really matters. Thanks for updating BSD

PR #1190, opened 2018-05-18
WIP: Remove V6 plugin Struct to reduce memory consumption

tas50:
We built a struct to track plugin file location because we needed both the file and the directory containing the file. We don't need that data anymore with V7 plugins. Instead we can just grab a list of files and then load each of the classes.

This shaves about 500k off Ohai runs

### DIFF
--- a/lib/ohai/loader.rb
+++ b/lib/ohai/loader.rb
@@ -52,7 +52,7 @@ module Ohai
         Ohai::Log.trace("Searching for Ohai plugins in #{path}")
 
         escaped = ChefConfig::PathHelper.escape_glob_dir(path)
-        Dir[File.join(escaped, "**", "*.rb")]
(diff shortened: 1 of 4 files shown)


lib/ohai/loader.rb
```diff
@@ -24,53 +24,39 @@
...
-        plugin_files + PluginFile.find_all_in(plugin_path)
-      end
+    # Searches all plugin paths and returns an Array of plugins paths
+    # @param dir [Array] array of plugin directories
+    #
+    # @return [Array] plugin paths
+    def plugin_files_by_dir(plugin_dir = Ohai.config[:plugin_path])
+      Array(plugin_dir).map do |path|
+        unless Dir.exist?(path)
+          Ohai::Log.info("The plugin path #{path} does not exist. Skipping...")
```
thommay: not sure this needs to be `info`

PR #1221, opened 2018-07-07
Add support for passing multiple plugin directories on CLI/client.rb

tas50:
This allows you to run ohai -d /foo/bar,/bar/foo

### DIFF
--- a/lib/ohai/application.rb
+++ b/lib/ohai/application.rb
@@ -34,8 +34,13 @@ class Ohai::Application
   option :directory,
     short: "-d DIRECTORY",
     long: "--directory DIRECTORY",
-    description: "A directory to add to the Ohai plugin search path",
-    proc: lambda { |path| Ohai::Config.platform_specific_path(path) }
+    description: "A directory to add to the Ohai plugin search path. If passing multiple directories comma separate directories.",
+    proc: lambda { |paths|
+      paths = paths.split(",")
+      paths.map do |path|
+        Ohai::Config.platform_specific_path(path)
+      end
+    }
 
   option :log_level,
     short: "-l LEVEL",
--- a/lib/ohai/application.rb
+++ b/lib/ohai/application.rb
@@ -34,12 +34,10 @@ class Ohai::Application
   option :directory,
     short: "-d DIRECTORY",
     long: "--directory DIRECTORY",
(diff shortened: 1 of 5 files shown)


lib/ohai/system.rb
```diff
@@ -176,9 +176,12 @@ def attributes_print(a)
     def configure_ohai
       Ohai.config.merge!(@config)
 
-      if Ohai.config[:directory] &&
-          !Ohai.config[:plugin_path].include?(Ohai.config[:directory])
-        Ohai.config[:plugin_path] << Ohai.config[:directory]
+      # add any additional CLI passed directories to the plugin path excluding duplicates
+      unless Ohai.config[:directory].nil?
+        Ohai.config[:directory].each do |dir|
```
lamont-granquist: This should probably be `Array(Ohai.config[:directory]).each` so it still supports scalars, otherwise you break people who might have been setting this value in config.

jaymzh:
Other than @lamont-granquist 's comment, I'm :+1:

tas50:
I'm not sure where those chefstyle warnings are coming from. All 3 are bogus and don't show up locally with bundle exec chefstyle. I'm on the same version too:

Chefstyle 0.10.0
  * RuboCop 0.55.0

lib/ohai/application.rb
```diff
@@ -34,8 +34,13 @@ class Ohai::Application
   option :directory,
     short: "-d DIRECTORY",
     long: "--directory DIRECTORY",
-    description: "A directory to add to the Ohai plugin search path",
-    proc: lambda { |path| Ohai::Config.platform_specific_path(path) }
+    description: "A directory to add to the Ohai plugin search path. If passing multiple directories comma separate directories.",
+    proc: lambda { |paths|
+      paths = paths.split(",")
```
btm: I think the last time I asked about this is lost in community slack now, but we've been using passing the option multiple times as the preferred UX over comma separated or other list parsing.

`--config-option` is one example of this: https://github.com/chef/chef/pull/5045
We switched knife ec2 to this behavior for security group IDs: https://github.com/chef/knife-ec2/pull/439
Also knife `--chef-tag` and `--aws-tag`: https://github.com/chef/knife-ec2/pull/520

tas50:
@btm @jaymzh @lamont-granquist This updated method work for you?

jaymzh:
Yeah that works! Thanks.

PR #1224, opened 2018-07-12
Load collect_data() even if we've already seen it

### LINKED ISSUE #1212: Ohai merges duplicate plugins in very dangerous ways
## Description

It turns out that Ohai `eval`s every plugin in every directory... and so all methods inside of `Ohai.plugin` blocks are evaluated... **however** the `collect_data` blocks are not - if a second one is hit on the same platform, it will not evaluate it, it runs on the first one it hits. This means the methods from the *last* plugin seen win, but the collector code from the *first* plugin seen wins. 

This means that you can into a problem where if you have two versions of the same module in different plugin paths, the **first** path will be run but call the methods from the **second** plugin.

So if you have `/root/ohai_repro/test.rb` with:

```
Ohai.plugin(:Test) do
  provides 'xtest'
  def test(a, b, c)
    Ohai::Log.warn("this is test() from test1")
  end
  collect_data do
    Ohai::Log.warn("this is collect_data from test1")
    test(1, 2, 3)
    xtest 'test1'
  end
end
```

So if you have `/root/ohai_repro2/test.rb` with:

```
Ohai.plugin(:Test) do
  provides 'xtest'
  def test(a, b, c)
    Ohai::Log.warn("this is test() from test2")
  end
  collect_data do
    Ohai::Log.warn("this is collect_data from test2")
    test(1, 2, 3)
    xtest 'test2'
  end
end
```

Your output will be the **very** unexpected:

```
[2018-07-06T10:36:27-07:00] WARN: this is collect_data from test1
[2018-07-06T10:36:27-07:00] WARN: this is test() from test2
```

**Woah**!! That's pretty weird.

## Ohai Version

13.8.0

works on 12 too.

## Platform Version

CentOS 7.5

btm:
If we happen to load a plugin with the same name twice, we refuse to run #collect_data but we still
load the rest of the plugin, which means you get #collect_data from the first plugin and rest of the
plugin from the last plugin.

This makes it so you get #collect_data from the last plugin as well.

I don't know how to use the logger from the class method so I fell back to `Ohai::Log`. Happy to fix that
if someone knows the magic.

Fixes #1212

Signed-off-by: Bryan McLellan <btm@chef.io>

### DIFF
--- a/lib/ohai/dsl/plugin/versionvii.rb
+++ b/lib/ohai/dsl/plugin/versionvii.rb
@@ -108,11 +108,8 @@ module Ohai
         # @param block [block] the actual code to collect data for the specified platforms
         def self.collect_data(platform = :default, *other_platforms, &block)
           [platform, other_platforms].flatten.each do |plat|
-            if data_collector.key?(plat)
-              raise Ohai::Exceptions::IllegalPluginDefinition, "collect_data already defined on platform #{plat}"
-            else
-              data_collector[plat] = block
-            end
+            Ohai::Log.warn("collect_data already defined on platform '#{plat}' for #{self}, last plugin seen will be used") if data_collector.key?(plat)
+            data_collector[plat] = block
           end
         end
 
--- a/spec/unit/dsl/plugin_spec.rb
+++ b/spec/unit/dsl/plugin_spec.rb
@@ -519,21 +519,20 @@ describe Ohai::DSL::Plugin::VersionVII do
     end
 
     it "fails a platform has already been defined in the same plugin" do
-      expect do
-        Ohai.plugin(:Test) do
-          collect_data {}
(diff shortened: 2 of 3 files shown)


lamont-granquist:
Yeah, the better way would be to fail it hard in some way or other, but there might be a chicken+egg with chef being unable to clean it up once it got into this state.  The ohai segment in cookbooks is probably the better way to fix this 'right'.

At least here there's some hope that the collect_data from the last-parsed plugin will use the methods from the last-parsed plugin and that will be halfway sane, although technically the object is going to get monkeypatched multiple times...

👍

btm:
Agree this isn't ideal, but it's also an edge case. hopefully anyone going down this path see's the warning and then avoids having multiple plugins with the same name.

Issue #1072, opened 2017-10-11
filesystem plugin timed out (df -P took over 30 seconds)


dkowis:
## Description

Apparently `df -P` takes over 30 seconds to complete on my VM, I think because I have NFS mounts. Is there any way to exclude those NFS mounts? I'm interested in the mounted local filesystems, not the network mounted filesystems.

```
[2017-10-11T17:07:51-05:00] DEBUG: Plugin Filesystem: ran 'df -P' and timed out after 30 seconds
[2017-10-11T17:07:51-05:00] DEBUG: Plugin Filesystem threw #<Ohai::Exceptions::Exec: Command timed out after 30s:
Command exceeded allowed execution time, process terminated
---- Begin output of df -P ----
STDOUT:
STDERR:
---- End output of df -P ----
Ran df -P returned >
[2017-10-11T17:07:51-05:00] DEBUG: /opt/chef/embedded/lib/ruby/gems/2.3.0/gems/ohai-8.23.0/lib/ohai/mixin/command.rb:46:in `rescue in shell_out'
[2017-10-11T17:07:51-05:00] DEBUG: /opt/chef/embedded/lib/ruby/gems/2.3.0/gems/ohai-8.23.0/lib/ohai/mixin/command.rb:37:in `shell_out'
[2017-10-11T17:07:51-05:00] DEBUG: /opt/chef/embedded/lib/ruby/gems/2.3.0/gems/ohai-8.23.0/lib/ohai/plugins/linux/filesystem.rb:53:in `block (2 levels) in <main>'
[2017-10-11T17:07:51-05:00] DEBUG: /opt/chef/embedded/lib/ruby/gems/2.3.0/gems/ohai-8.23.0/lib/ohai/dsl/plugin/versionvii.rb:88:in `instance_eval'
[2017-10-11T17:07:51-05:00] DEBUG: /opt/chef/embedded/lib/ruby/gems/2.3.0/gems/ohai-8.23.0/lib/ohai/dsl/plugin/versionvii.rb:88:in `run_plugin'
[2017-10-11T17:07:51-05:00] DEBUG: /opt/chef/embedded/lib/ruby/gems/2.3.0/gems/ohai-8.23.0/lib/ohai/dsl/plugin.rb:100:in `run'
[...]

PR #1227, opened 2018-08-04
Make the default shell_out timeout of 30 seconds configurable

WheresAlice:
### Description

This makes the default timeout of 30 seconds configurable

### Issues Resolved

- #1072 

### Check List

- [x] All tests pass

### DIFF
--- a/lib/ohai/config.rb
+++ b/lib/ohai/config.rb
@@ -40,6 +40,7 @@ module Ohai
       default :run_all_plugins, false
       # optional plugins are the set of plugins that are marked optional but you wish to run.
       default :optional_plugins, []
+      default :timeout, 30
     end
   end
 
--- a/lib/ohai/config.rb
+++ b/lib/ohai/config.rb
@@ -40,7 +40,7 @@ module Ohai
       default :run_all_plugins, false
       # optional plugins are the set of plugins that are marked optional but you wish to run.
       default :optional_plugins, []
-      default :timeout, 30
+      default :shellout_timeout, 30
     end
   end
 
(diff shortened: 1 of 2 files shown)


lib/ohai/config.rb
```diff
@@ -40,6 +40,7 @@ class Config
       default :run_all_plugins, false
       # optional plugins are the set of plugins that are marked optional but you wish to run.
       default :optional_plugins, []
+      default :timeout, 30
```
tas50: Can you change this to shellout_timeout so it's more clear what it changes?

coderanger:
Is there a specific use case in mind? Are you trying to make it longer? Shorter? Not against the patch, but we should have a goal in mind :)

lib/ohai/config.rb
lamont-granquist: +1 on using a more specific name

WheresAlice:
I'm thinking specifically around the use-case in #1072 - like the original reporter, we are also seeing chef runs failing on some of our systems with a lot of large network storage mounts.  At the moment we are having to manually patch this code after every upgrade of Chef client in order to make it work on some of our servers.

lib/ohai/config.rb
WheresAlice: Fair point, will do.

cheeseplus:
I like this one and the use case seems legit - I know we've seen this with other customers and clients

WheresAlice:
The existing tests check that the timeouts are set to 30 seconds, which seems to cover this change.  So I'm not sure I need to add any additional tests.

Should I be adding something to RELEASE_NOTES.md for this?

tas50:
I'll handle the expeditor fixes in another PR

Issue #746, opened 2016-02-22
Issue with how IP Address is determined for Windows?


JCapriotti:
I'm experiencing an issue with how Ohai is returning the IP address of nodes in my Windows environment. Our nodes have multiple network interfaces, and it seems to be configured in [this manner](http://directaccess.richardhicks.com/2013/06/19/network-interface-configuration-for-multihomed-windows-server-2012-directaccess-servers/). In my case, using the article's lingo, we want Ohai to return the IP of the external adapter, and not the internal adapter.

I started poking at the source code and I'm not sure if it is functioning correctly, based on that article's configuration guidance. I've been focusing on this code section in [plugins/windows/network.rb](https://github.com/chef/ohai/blob/a15a010c351c12271d090a35c855c5fe8135297c/lib/ohai/plugins/windows/network.rb#L106-L109)

That code seems to be setting the default interface and gateway values to whatever the last adapter it loops over is, provided it contains a value for `default_ip_gateway`. However, when the default gateway for the "internal" adapter is configured according to that first link (not set), the value returned for `default_ip_gateway` is 0.0.0.0, which is not nil and has size > 0. 

It seems like the code is trying to only set the default interface when its gateway is set. I feel like it is incorrectly assuming the `default_ip_gateway` will return nil, but instead it is returning 0.0.0.0 and therefore incorrectly setting the default interface when the internal adapter happens to be the last in the loop.

I realize I'm a few layers deep in here... but hopefully I'm making some sense. I'm not great with configuring networking settings, but it does seem like the intent of the code agrees with that configuration article, it just isn't working...

Thoughts? I'd be happy to address this, but I'm just not clear on the intent of the code.

lamont-granquist:
default gateway of '0.0.0.0' seems wrong.  i'd be concerned that this host got configured with a netmask of '0.0.0.0' instead of '255.255.255.0' or whatever and its ARP'ing on its subnet for the entire internet and has its own interface as its default gateway essentially.

of course we could just be buggy when we have two NICs as well...

smurawski:
@JCapriotti I've tried setting up machine with multiple NICs and mulitple default gateways and I can't replicate the behavior.  

I also don't quite get why you would be getting 0.0.0.0 as a default gateway.  Could you share what version of ohai you are seeing this with, along with the output from ohai and `ipconfig /all` in a gist?

JCapriotti:
[...]

PR #1231, opened 2018-08-10
Prefer ipv4 for default_gateway over ipv6 on windows

Nimesh-Msys:
- Also added a minor fix to select default_interface based on the least metric value
- Added Test cases
- Fixes: MSYS-875

Signed-off-by: Nimesh <nimesh.patni@msystechnologies.com>

### Description
According to previous logic, there was an iteration over all the interfaces and the value of `default_gateway`  and `default_interface` was getting reset in each loop. Hence the last interface was chosen as default_interface and so its default_gateway.
`default_gateway` was selected as the first one out from the list of `default_ip_gateway` of an interface, hence there was a possibility that if this list contain IPV6 address at former place, then it was selected.

In case of multiple Interfaces, priority for `default_interface` should be given to the interface with least metric value. Also, we need to prefer `IPV4` value of `default_gateway` over `IPV6`

### Issues Resolved

#[746](https://github.com/chef/ohai/issues/746)

### Check List

- [x] All tests pass

### DIFF
--- a/lib/ohai/plugins/windows/network.rb
+++ b/lib/ohai/plugins/windows/network.rb
@@ -53,6 +53,65 @@ Ohai.plugin(:Network) do
     data
   end
 
+  # Returns interface code for an interface
+  # 
+  # Interface Index (if present, Index otherwise) will be converted in hexadecimal format
+  #
+  # @param int_idx [String or nil] the interface index of interface
+  # @param idx [String] the index of interface
+  #
+  # @return [String]
+  #
+  # @example Interface Code when interface index is present
+  #   plugin.interface_code("1", "1") #=> "ox1"
+  # @example Interface Code when interface index is not present
+  #   plugin.interface_code(nil, "2") #=> "ox2"
+  #
+  def interface_code(int_idx, idx)
+    sprintf("0x%x", (int_idx || idx)).downcase
+  end
+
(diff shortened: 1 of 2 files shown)


spec/unit/plugins/windows/network_spec.rb
```diff
@@ -0,0 +1,141 @@
...
+describe Ohai::System, "Windows Network Plugin" do
+  let(:plugin) { get_plugin("windows/network") }
+
+  before(:each) do
+    allow(plugin).to receive(:collect_os).and_return(:windows)
+  end
+
+  # it "Failed example !!!" do
+  #   expect(plugin.favored_ip).to eq('b')
+  # end
```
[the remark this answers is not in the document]
Nimesh-Msys: Sorry. That was a typo. Removing.

lib/ohai/plugins/windows/network.rb
```diff
@@ -53,6 +53,42 @@ def msft_adapter_data
     data
   end
 
+  # Returns the interface code in Hexadecimal Format
```
tas50: YARD format would be helpful here in the future

Nimesh-Msys: Updating them
Thank you

tas50:
Thanks for the comments @Nimesh-Msys

PR #1241, opened 2018-09-09
Fix providing relative paths to the config file

tas50:
We do an expand_path on the passed config value and we use Dir.pwd with the expand, but before we ever get here we change the directory to /, which means specifying a file in the working directory results in something like /client.rb. This isn't what the user wants. Instead change the working directory right before we run the application instead of before we work out the config situation.

### DIFF
--- a/lib/ohai/application.rb
+++ b/lib/ohai/application.rb
@@ -69,14 +69,6 @@ class Ohai::Application
     proc: lambda { |v| puts "Ohai: #{::Ohai::VERSION}" },
     exit: 0
 
-  def initialize
-    super
-
-    # Always switch to a readable directory. Keeps subsequent Dir.chdir() {}
-    # from failing due to permissions when launched as a less privileged user.
-    Dir.chdir("/")
-  end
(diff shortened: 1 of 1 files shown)


tas50:
I'm really not actually sure if we need this chdir anymore at all. I created a non-priv'd user on my Mac and ran ohai using that user from the home directory of my personal user without issue. The original reason we added this chdir was to get around errors running ohai from within a directory we didn't have privs on. https://tickets.opscode.com/browse/OHAI-174

lamont-granquist:
yeah that old ohai ticket is crazyballs.

this line of code ultimately needs to get removed.

callers should be responsible for doing `cd /` themselves.  this is caused by the old chef-solr cli which used to drop privs and apparently run ohai for some reason.  it should have been made responsible for chdir /'ing itself.  similarly any init scripts that run chef-client/ohai need to be responsible for doing that themselves.

tas50:
I created #1246 to track removing the chdir altogether.

PR #1259, opened 2018-09-27
Avoid gathering all data with sysctl which seems to hang

tas50:
While it works fine for me @bdausses was seeing chef-client hang
forever on macos 10.14. It turns out running sysctl -a on his VM was causing the
system to hang. Besides this crash it actually takes about 2x longer to
gather the additional data, which we never use. This shaves a whole 0.033255 seconds from
the total runtime of Ohai.

### DIFF
--- a/spec/unit/plugins/darwin/cpu_spec.rb
+++ b/spec/unit/plugins/darwin/cpu_spec.rb
@@ -1,6 +1,6 @@
 #
 # Author:: Nathan L Smith (<nlloyds@gmail.com>)
-# Copyright:: Copyright (c) 2013-2016 Chef Software, Inc.
+# Copyright:: Copyright (c) 2013-2018 Chef Software, Inc.
 # License:: Apache License, Version 2.0
 #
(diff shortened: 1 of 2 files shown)


spec/unit/plugins/darwin/cpu_spec.rb
```diff
@@ -78,7 +78,7 @@
 CTL
 
     allow(@plugin).to receive(:collect_os).and_return(:darwin)
-    allow(@plugin).to receive(:shell_out).with("sysctl -a").and_return(mock_shell_out(0, @stdout, ""))
+    allow(@plugin).to receive(:shell_out).with("sysctl cpu machdep").and_return(mock_shell_out(0, @stdout, ""))
```
tyler-ball: The code runs `sysctl hw` and the test expects `sysctl cpu` - looks like the source of the failures

tyler-ball:
@tas50 Updated the spec to match `sysctl hw machdep` in expectation

PR #1262, opened 2018-10-03
Unify the cpu plugin

jaymzh:
### Description

Utilize the Ohai 7 syntax to keep the code for all these variations of the plugin in the same place and consistent.

In this case I kept the specs in separate files since there's no actual overlap in tests AFAICT and each platform's tests are exceptionally complicated, it's cleaner to leave them separate.

### Check List

- [X] All tests pass
- [X] RELEASE\_NOTES.md, has been updated if required (not required for bugfixes, required for API changes)


tas50:
Another 100 lines down!

PR #1263, opened 2018-10-03
Trim out bogus data in system_profile plugin

tas50:
Also there's no need to rescue requiring plist since ohai depends on it. We don't do this for anything else in Ohai. This shaves 8k from the node data. I'm also going to open a PR to deprecate this plugin so we can remove it in Chef 15 since it doesn't actually work on modern systems and we already collect the same data in the hardware plugin.


jaymzh:
This looks good to me!

PR #1264, opened 2018-10-03
Deprecate the system_profile plugin

tas50:
1) It's broken
2) It's a huge amount of data on the node
3) It takes 3 seconds to run
4) Hardware plugin does the same thing, but better

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,3 +1,11 @@
+# Unreleased
+
+## New Deprecations
+
+### system_profile plugin
+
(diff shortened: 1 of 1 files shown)


jaymzh:
Hey @mikedodge04  - I'm pretty sure we're OK with this - you've replaced this functionality in other ohai plugins but can you just confirm?

natewalck:
Burn it with fire. We can already get this data from other sources as Phil said.

mikedodge04:
I'm in agreement. Lets kill it !

tas50:
Thanks for the comments everyone. Faster mac ohai runs here we come.

Issue #1282, opened 2018-10-29
Add support for xcp-ng in Ohai


heyjodom:
## Description

XCP-ng is the opensource release of XenServer

## Ohai Version

13.10.0

## Platform Version

XCP-ng (Centos base)

makemebacon:
I have a PR I can submit to add support. Patched it locally in our environment.

tas50:
Closing this out since we shipped this

PR #1283, opened 2018-10-29
Add support for XCP-ng platform

heyjodom:
Signed-off-by: Justin Odom <jodom@tripadvisor.com>

### Description

Add support for XCP-ng to be classified as rhel. Similar behavior to XenServer.

### Issues Resolved

Issue #1282 

### Check List

- [x] All tests pass

### DIFF
--- a/lib/ohai/plugins/linux/platform.rb
+++ b/lib/ohai/plugins/linux/platform.rb
@@ -127,7 +127,7 @@ Ohai.plugin(:Platform) do
     when /debian/, /ubuntu/, /linuxmint/, /raspbian/, /cumulus/
       # apt-get+dpkg almost certainly goes here
       "debian"
-    when /oracle/, /centos/, /redhat/, /scientific/, /enterpriseenterprise/, /xenserver/, /cloudlinux/, /ibm_powerkvm/, /parallels/, /nexus_centos/, /clearos/, /bigip/ # Note that 'enterpriseenterprise' is oracle's LSB "distributor ID"
+    when /oracle/, /centos/, /redhat/, /scientific/, /enterpriseenterprise/, /xcp/, /xenserver/, /cloudlinux/, /ibm_powerkvm/, /parallels/, /nexus_centos/, /clearos/, /bigip/ # Note that 'enterpriseenterprise' is oracle's LSB "distributor ID"
       # NOTE: "rhel" should be reserved exclusively for recompiled rhel versions that are nearly perfectly compatible down to the platform_version.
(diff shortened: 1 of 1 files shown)


lamont-granquist:
does the platform_version for this platform actually match the rhel numbering scheme and is 6.x guaranteed to be RHEL6 ABI compatible and 7.x guaranteed to be RHEL7 ABI compatible, etc?  Because if the underlying system is a mix+match of RHEL RPMs (similar to the Amazon RHEL6-mixed-with-RHEL7-without-systemd mutants) or even if the platform_version just disagrees, then it may be better to put this under the "fedora" umbrella (and I have the same questions about XenServer as well -- although without any complaints about it, I wouldn't bother changing that one).

makemebacon:
I'm not the maintainer or contributor on XCP, so not able to answer with 100% certainty. That being said, everything that i've found seems to show it as matching up with the rhel numbering scheme and compatible (XCP-ng wasn't around in 6.x as it was just plain XenServer then, so not able to look on one of those systems).

As with XenServer, major and minor versions follow the overall centos release they are built on, however, there is no patch version defined. 

XenServer 7.2 example:
  "platform": "xenserver",
  "platform_version": "7.2.0",
  "platform_family": "rhel",

XCP-ng 7.5 example:
  "platform": "xcp",
  "platform_version": "7.5.0",
  "platform_family": "rhel",

makemebacon:
@lamont-granquist Actually, it looks like things line up now but that is not something guaranteed for the future (just had a quick chat with the folks working on xcp-ng). It's possible there might be divergence with the rhel/centos numbering scheme in the future with what is returned in platform_version.

That being said, do you see any possible issues if platform_family were to be changed to fedora with a 7.x or 8.x platform_version numbering?

lib/ohai/plugins/linux/platform.rb
```diff
@@ -139,7 +139,7 @@ def determine_platform_family
       "amazon"
     when /suse/, /sles/, /opensuse/
       "suse"
-    when /fedora/, /pidora/, /arista_eos/
+    when /fedora/, /pidora/, /arista_eos/, /xcp/
```
tas50: Looking at the xcp docs it looks like we should actually classify this as RHEL. Can you change that back and then we can merge this.

Also can you open a PR against https://github.com/chef/os_release. We're updating ohai to use the /etc/os-release file for system detection and it will help us to have an idea what this system looks like.

heyjodom:
Reverted change. Tests are running now. Will open a PR against os-release as well as requested.

tas50:
Thanks for working with us despite the back and forth here

PR #1297, opened 2018-11-23
Correctly detect openSUSE leap 15+

### LINKED ISSUE #1289: Ohai not detecting openSUSE Leap properly
## Description

Ohai does not identify openSUSE Leap properly. I'm using the official Docker Image from https://hub.docker.com/r/opensuse/leap/
I've collected detection relevant info in this issue: https://github.com/inspec/inspec/issues/3574

I assume it fails to pass this line https://github.com/chef/ohai/blob/master/lib/ohai/plugins/linux/platform.rb#L294 because its ID value is different `ID="opensuse-leap"`.

## Ohai Version

14.6.2

## Platform Version

openSUSE Leap 15

## Ohai Output

See linked issue.

tas50:
We need to correctly read in the data in /etc/os-release and do the right thing. We already did this for SLES 15.

Fixes #1289. Thanks for great issue report @megamorf


tas50:
This is part of #1299 which was approved so merging here so we can backport just this bit to ohai 14

PR #1299, opened 2018-11-24
/etc/os-release based OS detection

tas50:
/etc/os-release is the modern way of identifying Linux platforms and versions. It's far more reliable than the brittle regex mess that we have built up over the last decade. It also allows us to easily detect unknown / obscure distros since they have platform IDs in the os-release file.

Changes to call out:
- Platform detection will now default to /etc/os-release if that file is present
- The majority of the legacy logic is retained within a legacy method
- Unless we override the value we just use the platform value from os-release. Going forward for new platforms we'll just use this value. The overrides are just to retain legacy compatibility
- Unless it's missing we just use the version found in the os-release file. If it's not found we grab the kernel version with uname as we did before (gentoo/arch/etc)
- Some platforms where we know users stay more up to date have been switched over to the os-release logic entirely: arch, alpine, and cumulus, mint, raspbian, and clearlinux. This also speeds up the remaining legacy platforms.
- Testing has been simplified and expanded since we're mostly testing the platform remapping method, and the platform to platform_family methods. There's no need to run the plugin for each possible platform when the logic is entirely identical sans the value in the os-release file now.
- The new logic should be faster as we don't have to check for the existence of a large number of files. We just read out a single file and then lookup a few things in a hash table. 
- Adds manjaro to the arch platform_family and kali to the debian platform_family.

Things that change:
- Nexus switch versions are now properly identified with the shorter version string such as 7.0(3)I2(2b) instead of 7.0(3)I2(0.475E.6) which includes the build number.

### DIFF
--- a/spec/unit/plugins/linux/platform_spec.rb
+++ b/spec/unit/plugins/linux/platform_spec.rb
@@ -312,7 +312,6 @@ OS_DATA
     let(:have_usr_lib_os_release) { false }
     let(:have_cisco_release) { false }
     let(:have_f5_release) { false }
-    let(:have_cumulus_dir) { false }
 
     before(:each) do
       allow(plugin).to receive(:collect_os).and_return(:linux)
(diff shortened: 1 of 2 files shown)


spec/unit/plugins/linux/platform_spec.rb
```diff
@@ -19,974 +19,768 @@
...
-    end
-
-    it "should set platform to amazon and platform_family to rhel when [:lsb][:id] contains Amazon" do
-      @plugin[:lsb][:id] = "AmazonAMI"
-      @plugin[:lsb][:release] = "2011.09"
-      @plugin.run
-      expect(@plugin[:platform]).to eq("amazon")
-      expect(@plugin[:platform_family]).to eq("amazon")
+  describe "#read_os_release_info" do
+    let(:file_contents) { "COW=MOO\nDOG=\"BARK\"" }
```
lamont-granquist: but what does the fox say?

Issue #1302, opened 2018-11-25
Plugin to load hyper_v hostname from guest


safematix:
## Description

Plugin to load hyper_v hostname from guest, to see on which node the guest is running.

## Ohai Version

```
Ohai: 14.5.5
```

## Platform Version

Linux systems with:

```bash
lsmod | grep hv
hv_balloon             13836  0 [permanent]
hv_netvsc              28115  0
hv_utils               14888  0
hv_storvsc             12817  4
hv_vmbus              280508  7 hv_balloon,hv_netvsc,hv_utils,hid_hyperv,hyperv_keyboard,hv_storvsc,hyperv_fb
```

## Ohai Output

example:

`ohai virtualization`
```json
{
  "systems": {
    "hyperv": "guest"
  },
  "system": "hyperv",
  "role": "guest",
  "host": "hyper_v.example.com"
}
```

PR #1303, opened 2018-11-25
Plugin to load hyper_v hostname from guest

safematix:
### Description

Plugin to load hyper_v hostname from guest, to see on which node the guest is running.

### Issues Resolved

- Issue #1302

### Check List

- [x] All tests pass

### DIFF
--- a/lib/ohai/plugins/linux/virtualization.rb
+++ b/lib/ohai/plugins/linux/virtualization.rb
@@ -158,7 +158,7 @@ Ohai.plugin(:Virtualization) do
       virtualization[:system] = "hyperv"
       virtualization[:role] = "guest"
       virtualization[:systems][:hyperv] = "guest"
-      virtualization[:host] = hyperv_host
+      virtualization[:hypervisor_host] = hyperv_host
     end
 
     # Detect Linux-VServer
--- a/lib/ohai/plugins/linux/virtualization.rb
+++ b/lib/ohai/plugins/linux/virtualization.rb
@@ -152,12 +152,12 @@ Ohai.plugin(:Virtualization) do
 
     # Detect Hyper-V gest and host hostname
     if File.exist?("/var/lib/hyperv/.kvp_pool_3")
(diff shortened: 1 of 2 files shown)


tas50:
There's a small chefstyle warning that you can resolve by running `chefstyle -a .` on the ohai dir

lib/ohai/plugins/linux/virtualization.rb
```diff
@@ -150,6 +150,17 @@ def docker_exists?
       end
     end
 
+    # Detect Hyper-V gest and host hostname
```
tas50:
```suggestion
    # Detect Hyper-V guest and host hostname
```

lib/ohai/plugins/linux/virtualization.rb
```diff
@@ -150,6 +150,17 @@ def docker_exists?
...
 
+    # Detect Hyper-V guest and the hostname of the host
+    if File.exist?("/var/lib/hyperv/.kvp_pool_3")
+      logger.trace("Plugin Virtualization: /var/lib/hyperv/.kvp_pool_3 contains string indicating Hyper-V guest")
+      data = File.read("/var/lib/hyperv/.kvp_pool_3")
+      hyperv_host = data[/\HostName(.*?)HostingSystemEditionId/, 1].scan(/[[:print:]]/).join.downcase
+      virtualization[:system] = "hyperv"
+      virtualization[:role] = "guest"
+      virtualization[:systems][:hyperv] = "guest"
+      virtualization[:host] = hyperv_host
```
tas50: I'm not sure how I feel about this attribute name. It seems a bit ambiguous. How would you feel about hypervisor_host instead? I'm open to other ideas as well.

safematix: That's a good idea, I think "hypervisor_host" is better than "host". That's more understandable and less confusing. :blush:

tas50:
@safematix thanks for putting this together. This will be quite handy for a lot of users

PR #1311, opened 2018-11-28
Add Ohai 14.8 release notes

tas50:
There may be a few more things, but this is the general list. I took out
the GCE change since that's not visible to end users. That's more for
Google.


tas50:
This same change was already approved by docs team on chef/chef

PR #1318, opened 2018-11-29
WIP: Auto detect softlayer hosts based on DMI data

tas50:
We currently rely on the hint data, which is fine if we bootstrap the
system ourselves, but for a lot of users this hint data won't exist and
we can auto detect a system based on the DMI data.


tas50:
Closing this out since this doesn't get us the VMs and I think this is of limited value for the hardware systems

PR #1319, opened 2018-11-30
bsd virtualization: Detect amazonec2 hypervisor + kvm without dmidecode

tas50:
Refactor a chunk of code. There's no need to split the sysctl output
since it's just a single line. There's also a ton of repetition in how
we detect guests from sysctl kernel data. Add KVM and amazonec2
detection there.


tas50:
root@freebsd:/usr/home/ec2-user/ohai # bundle exec ohai virtualization
[2018-11-30T04:54:41+00:00] INFO: The plugin path /etc/chef/ohai/plugins does not exist. Skipping...
{
  "systems": {
    "amazonec2": "guest"
  },
  "system": "amazonec2",
  "role": "guest"
}

PR #1336, opened 2019-01-23
Fully remove support for HP-UX

tas50:
There hasn't been a working Ruby release for HP-UX for many years. We shouldn't continue to execute this logic and ship it around to all our customers when it's not possible that any of them are running HP-UX.


tas50:
Pour one out for @juliandunn

PR #1342, opened 2019-03-19
Loosen the mixlib version dependencies

tas50:
Allow us to bring in the new major releases which just drop Ruby 2.1/2.2
support.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,7 @@ gemspec
 
 # NOTE: do not submit PRs to add pry as a dep, add to your Gemfile.local
 group :development do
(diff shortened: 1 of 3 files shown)


Gemfile
```diff
@@ -4,7 +4,7 @@ gemspec
 
 # NOTE: do not submit PRs to add pry as a dep, add to your Gemfile.local
 group :development do
-  gem "chefstyle", git: "https://github.com/chef/chefstyle.git", branch: "master"
+  gem "chefstyle", "=0.11.2"
```
lamont-granquist: ah right this is the 14-stable branch so it should never get updated.

PR #1346, opened 2019-04-04
Fixes ec2 adding json? attribute

burtlo:
## Description

Because of the method_missing approach to setting data the use of json?
here without the function serves only to assign the response.body to a
new attribute named `json?`

Signed-off-by: Franklin Webber <franklin.webber@gmail.com>

## Types of changes

## Checklist:

### DIFF
--- a/lib/ohai/mixin/ec2_metadata.rb
+++ b/lib/ohai/mixin/ec2_metadata.rb
@@ -202,6 +202,20 @@ module Ohai
         key.gsub(/\-|\//, "_")
       end
 
+      # @param [String] data that might be JSON
+      #
+      # @return [Boolean] is the data JSON or not?
+      def json?(data)
+        data = StringIO.new(data)
+        parser = FFI_Yajl::Parser.new
+        begin
+          parser.parse(data)
+          true
+        rescue FFI_Yajl::ParseError
+          false
+        end
+      end
+
     end
   end
 end


lamont-granquist:
You're double-parsing the data here though, this would probably be better written by removing the `json?` method entirely and parsing the response.body and directly rescuing the exception.

lib/ohai/mixin/ec2_metadata.rb
```diff
@@ -202,6 +202,20 @@ def metadata_key(key)
         key.gsub(/\-|\//, "_")
       end
 
+      # @param [String] data that might be JSON
+      #
+      # @return [Boolean] is the data JSON or not?
+      def json?(data)
+        data = StringIO.new(data)
```
tas50: Instead of creating the objects twice why not just change the above if / elseif to rescue to the 2nd condition? It would save a lot of duplicate object setup. Also this same issue exists in gce mixin as well from my quick glance.

tas50: There's got to be some sort of jinx when you review the same thing without seeing it because github didn't update while you were typing out a comment.

burtlo:
https://github.com/chef/ohai/blob/master/lib/ohai/mixin/gce_metadata.rb#L60-L69

lamont-granquist:
yeah, our existing code is not necessarily an indication of acceptable code.

aaronlippold:
I love it when a community is dead honest with itself :-) kudos

PR #1354, opened 2019-04-24
Add :no_df_remote configuration flag to filesystem plugin

malmond77:
## Description
`df` runs statfs across all filesystems including remote ones. When run at scale, Ohai adds extra load simply to obtain block / inode usage free information. This is rarely useful. Turn on this option to disable remote filesystem enumeration in `df`. Note that device + mountpoint will still be obtained from `mount` and `/proc/mounts`

## Related Issue
Issue #1357 

## Types of changes

## Checklist:

### DIFF
--- a/lib/ohai/plugins/filesystem.rb
+++ b/lib/ohai/plugins/filesystem.rb
@@ -159,8 +159,8 @@ Ohai.plugin(:Filesystem) do
       # free information. This is rarely useful. Turn on this option to
       # disable remote filesystem enumeration in df. Note that device +
       # mountpoint will still be obtained from mount and /proc/mounts.
-      params = if Ohai.config[:plugin][:filesystem][:no_remote]
-        ' --local'
+      params = if Ohai.config[:plugin][:filesystem][:no_df_remote]
+        ' -l'
       else
         ''
       end
@@ -168,7 +168,7 @@ Ohai.plugin(:Filesystem) do
       fs.merge!(parse_common_df(so.stdout))
 
       # Grab filesystem inode data from df
-      so = shell_out("df -iP#{local}")
+      so = shell_out("df -iP#{params}")
       so.stdout.each_line do |line|
         case line
         when /^Filesystem\s+Inodes/
--- a/lib/ohai/plugins/filesystem.rb
+++ b/lib/ohai/plugins/filesystem.rb
@@ -154,11 +154,21 @@ Ohai.plugin(:Filesystem) do
 
     # Grab filesystem data from df
     run_with_check("df") do
-      so = shell_out("df -P")
+      # df runs statfs across all filesystems including remote ones. When run
+      # at scale, Ohai adds extra load, simply to obtain block / inode usage
+      # free information. This is rarely useful. Turn on this option to
+      # disable remote filesystem enumeration in df. Note that device +
+      # mountpoint will still be obtained from mount and /proc/mounts.
+      params = if Ohai.config[:plugin][:filesystem][:no_remote]
+        ' --local'
+      else
+        ''
+      end
+      so = shell_out("df -P#{params}")
       fs.merge!(parse_common_df(so.stdout))
 
       # Grab filesystem inode data from df
-      so = shell_out("df -iP")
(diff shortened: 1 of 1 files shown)


lib/ohai/plugins/filesystem.rb
```diff
@@ -154,11 +154,21 @@ def run_with_check(bin, &block)
 
     # Grab filesystem data from df
     run_with_check("df") do
-      so = shell_out("df -P")
+      # df runs statfs across all filesystems including remote ones. When run
+      # at scale, Ohai adds extra load, simply to obtain block / inode usage
+      # free information. This is rarely useful. Turn on this option to
+      # disable remote filesystem enumeration in df. Note that device +
+      # mountpoint will still be obtained from mount and /proc/mounts.
+      params = if Ohai.config[:plugin][:filesystem][:no_remote]
```
jaymzh: no_remote is a lie, we definitely still get remote fs info. How about no_df_remote

lib/ohai/plugins/filesystem.rb
```diff
@@ -154,11 +154,21 @@ def run_with_check(bin, &block)
...
+        ' --local'
+      else
+        ''
+      end
+      so = shell_out("df -P#{params}")
       fs.merge!(parse_common_df(so.stdout))
 
       # Grab filesystem inode data from df
-      so = shell_out("df -iP")
+      so = shell_out("df -iP#{local}")
```
jaymzh: did you mean `params` ?

lib/ohai/plugins/filesystem.rb
malmond77: Agreed

lib/ohai/plugins/filesystem.rb
malmond77: yeah, not enough coffee yesterday it seems.

lib/ohai/plugins/filesystem.rb
```diff
@@ -154,11 +154,21 @@ def run_with_check(bin, &block)
...
     # Grab filesystem data from df
     run_with_check("df") do
-      so = shell_out("df -P")
+      # df runs statfs across all filesystems including remote ones. When run
+      # at scale, Ohai adds extra load, simply to obtain block / inode usage
+      # free information. This is rarely useful. Turn on this option to
+      # disable remote filesystem enumeration in df. Note that device +
+      # mountpoint will still be obtained from mount and /proc/mounts.
+      params = if Ohai.config[:plugin][:filesystem][:no_df_remote]
+        ' -l'
```
jaymzh: FYI this should be aligned like:

```
params = if Ohai.config[...
           ' -l'
         else
           ''
         end
```

Which is terrible, so instead use terniary:

```
params = Ohai.config[:plugin][:filesystem][:no_df_remote] ? ' -l' : ''
```

lib/ohai/plugins/filesystem.rb
```diff
@@ -154,11 +154,21 @@ def run_with_check(bin, &block)
...
+      else
+        ''
+      end
+      so = shell_out("df -P#{params}")
```
jaymzh: since you're butting it up against this, why not just make the param `l` instead of ` -l` so that way you end up with "df -l" or "df -Pl" ?

lib/ohai/plugins/filesystem.rb
malmond77: My original code used the long option. I think it's been supported for years. I much prefer it, especially in code. However, this leads it to be inconsistent with the `-P` / `--portability` option and all things considered, I want to keep the style close to the original.

My original code exceeded 80 columns, so I split it across lines using the longer if. Now the option is short, it all fits.

tas50:
@malmond77  I'm going to close this out now since we need the DCO sign-of in order to merge PRs. If you still want to commit this then rebase it on master and sign the commits for the DCO

jaymzh:
@tas50  to be clear, it also needed a other changes before being merged... per my comments. I think he's waiting to have time to fix the code. Closing this in the interim seems fine.

PR #1360, opened 2019-05-03
Avoid constant warnings in windows/filesystem plugin

tas50:
This is very noisy in DK specs.

### DIFF
--- a/lib/ohai/plugins/windows/filesystem.rb
+++ b/lib/ohai/plugins/windows/filesystem.rb
@@ -24,7 +24,7 @@ Ohai.plugin(:Filesystem) do
   # @see https://docs.microsoft.com/en-us/windows/desktop/SecProv/getconversionstatus-win32-encryptablevolume#parameters
   #
   CONVERSION_STATUS = %w{FullyDecrypted FullyEncrypted EncryptionInProgress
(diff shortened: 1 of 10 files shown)


lib/ohai/plugins/windows/filesystem.rb
```diff
@@ -24,7 +24,7 @@
   # @see https://docs.microsoft.com/en-us/windows/desktop/SecProv/getconversionstatus-win32-encryptablevolume#parameters
   #
   CONVERSION_STATUS = %w{FullyDecrypted FullyEncrypted EncryptionInProgress
-                         DecryptionInProgress EncryptionPaused DecryptionPaused}.freeze
+                         DecryptionInProgress EncryptionPaused DecryptionPaused}.freeze unless defined?(CONVERSION_STATUS)
```
zenspider: 1. I feel like this is a bandaid that hides a problem in `plugin.rb` (`NamedPlugin.plugin` is where this seems to be coming from).
2. You can change that to `||=` instead of `unless defined?`

tas50: I updated this to improve our bandaid to use ||= everywhere

Issue #1368, opened 2019-06-03
RPM packages can have multiple packages with the same `name`.


jjustice6:
### Describe the Enhancement:
It's legal to have two `rpm` packages with the _exact same name_ installed. not (for example) `python` and `python2.7`, but literally the exact same _name_. Unfortunately, Ohai uses the package name as the [key of the hash](https://github.com/chef/ohai/blob/v15.0.37/lib/ohai/plugins/packages.rb).

### Describe the Need:
The need here is evident from #968 , however, kernel and gpg key packages are not the only packages which have this problem. In theory, any package can have this problem as long as it is legal for the RPM packages to live side by side, such as installing libs for another architecture, or using certain vendor-provided packages.

### Current Alternative
There is no current alternative.

### Can We Help You Implement This?:
It's very evident that changing the Ohai plugin to use the full combination of name+version+release+architecture as the key would break tons of existing code. I suggested to @tas50 that we could put an 'other_versions' array in the hash, as in `node['packages'][name]['other_versions']` and put the information for every other version of the package installed in there. This may render #968 unnecessary as this would be a generic solution to the problem.

I have written custom ohai plugins with tests before, so I believe I can have a PR for you in a few days that implements this functionality.

tas50:
Closing this out since this was merged in

PR #1369, opened 2019-06-03
Add other_versions subfield for RPM packages.

jjustice6:
Signed-off-by: Joshua Justice <jjustice6@bloomberg.net>

## Description
It's legal to have two rpm packages with the exact same name installed. not (for example) python and python2.7, but literally the exact same name. Unfortunately, Ohai uses the package name as the key of the hash. This means Ohai currently only keeps the information of the last package in the list. 

This now produces a list of all versions of the package if there is more than one version present.

## Related Issue
https://github.com/chef/ohai/issues/1368

Please note that we strongly desire that this be backported into chef-client 14.x after its merge.

## Types of changes

## Checklist:


lib/ohai/plugins/packages.rb
```diff
@@ -50,7 +50,16 @@
 
       pkgs.each do |pkg|
         name, epoch, version, release, installdate, arch = pkg.split
-        packages[name] = { "epoch" => epoch, "version" => version, "release" => release, "installdate" => installdate, "arch" => arch }
+        if packages[name]
+          # We have more than one package with this exact name!
+          # Create an "other_versions" array for tracking all versions of packages with this name.
+          # The top-level package information will be the first one returned by rpm -qa,
+          # all others go in this list, with the same information they'd normally have.
+          packages[name]["other_versions"] = [] unless packages[name]["other_versions"]
```
tas50:
```suggestion
          packages[name]["other_versions"] ||= []
```

lib/ohai/plugins/packages.rb
```diff
@@ -50,7 +50,16 @@
...
+          packages[name]["other_versions"] << Mash.new({ "epoch" => epoch, "version" => version, "release" => release, "installdate" => installdate, "arch" => arch })
```
tas50: Any chance any of these would ever be unique so we could just use them as a key for a hash?

lib/ohai/plugins/packages.rb
jjustice6: Ah, perfect.

lib/ohai/plugins/packages.rb
jjustice6: Not unless you want to concatenate stuff. By definition you need name+version+release+architecture to guarantee uniqueness. You could roll the dice with epoch or installdate, but those aren't _guaranteed_ unique (and I suspect there are edge cases where they break, like install discs or similar). An array seems like the least bad option.

lib/ohai/plugins/packages.rb
```diff
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
```
tas50: shouldn't this be encoding?

lamont-granquist: unnecessary now (this is pre-2.0 i think)

tas50: Don't you still want it if you're trying to force the encoding.

lamont-granquist: only if you want to set it to something other than utf-8

lamont-granquist:
i think this would be more useful as `versions` rather than `other_versions` and just list all of them, which can then be used to sidestep the 'best' one to use as the 'main' one.

lib/ohai/plugins/packages.rb
jjustice6: Did a `(setq ruby-insert-encoding-magic-comment nil)` and cleaned this up. Sorry.

jjustice6:
I'm not picking a "best" one, but rather keeping the first one out of the list. If we move it into an array of versions, that extra level would break existing code that just assumes there's one package. (Also would make the interface for RPMs different than for everything else without doing work on those.)

jjustice6:
Looks like the buildkite failures have nothing to do with this, incidentally.

lib/ohai/plugins/packages.rb
lamont-granquist: the NEVRA itself (name-epoch-version-release-arch) should be unique

tas50:
@jjustice6 if you rebase on master that build failure will go away

jjustice6:
All right, there we go. I've also incorporated the return of the clobbering per discussion with @lamont-granquist in slack.

jjustice6:
This approach now puts _all_ versions of a package in a version table, if and only if there are multiple versions of the package present (so no duplication of information in the majority of cases).

jjustice6:
Do you have any further requests for this PR?

glensc:
the final version adds `versions` subkey. the title should be updated. also this got merged into 14.8.14 as well (#1386)

jjustice6:
@glensc Thanks for the catch, I've updated the title of this for posterity's sake.

PR #1371, opened 2019-06-06
Try running a smaller container in BK

tas50:
We'll see if this if faster


tas50:
This is not worth the effort at the moment

PR #1373, opened 2019-06-11
Use virtualization attributes to run or skip the virtualbox plugin

tas50:
Instead of attempting to shellout on every system we can just check the attributes we already have. This is slightly faster and improves logging since you won't see a failure.


tas50:
Thanks @robbkidd

PR #1376, opened 2019-06-13
Merge the new vbox plugin into the existing virtualbox plugin

tas50:
It makes sense to handle both guest and host logic in the same plugin since you can't be both. This gives a more predictable namespace and makes the code slightly smaller. I also removed the ostypes field since that doesn't change by system and is an enormous amount of data to collect with pretty limited value from a cookbook standpoint.


lib/ohai/plugins/virtualbox.rb
```diff
@@ -20,8 +22,128 @@
...
+        left, right = line.split("=")
+
+        # remove enclosing quotes, if needed
+        key =
+          case left
+          when /^"(.*)"$/
+            Regexp.last_match(1)
+          else
+            left
+          end
```
lamont-granquist: probably:

```
key = left
key = Regexp.last_match(1) if left ~= /^"(.*)"$/
```

tas50: Updated using delete_prefix / delete_suffix per our review

lib/ohai/plugins/virtualbox.rb
```diff
@@ -20,8 +22,113 @@
...
+    if so.exitstatus == 0
+      so.stdout.lines.each do |line|
+        line.chomp!
+        left, right = line.split("=")
+
+        # remove enclosing quotes, if needed
+        key = left.delete_prefix('"').delete_suffix('"')
+
+        # skip the name attribute since that is the parent key
+        next if left == "name"
```
robbkidd: Maybe want to test `key` here after the `"` chomping on `left`. Is the value always lowercase `name` or should this test be after downcasing?

tas50: Updated

Issue #1343, opened 2019-03-26
Backport #1299 to 14.x?


michaelklishin:
Looks like https://github.com/chef/ohai/pull/1299 (and thus https://github.com/chef/ohai/pull/1299/commits/6b14655d54bef0f4f6be28f624aa1819e2393f16) is only available in Chef 15.x series. Any chance it can be backported to 14.x so that OpenSUSE Leap 15 users can, ahem, use Chef?

michaelklishin:
Any feedback from Ohai maintainers?

michaelklishin:
@tas50 I'm truly sorry to bother you directly but you seem to be by far most active person in the Chef project space on GitHub. No response in ~ 2 weeks in April is rough. Any opinion on this?

tas50:
I'm uncomfortable backporting #1299 to Chef 14. It's a huge improvement to our platform detection login, but it's probably a breaking change for someone since it reworks 10 years of historical platform detection. We'd need to rework SLES 15 within the existing Chef 14 codebase.

tas50:
I'm going to close this out at this point since #1377 has the fix for opensuse

PR #1375, opened 2019-06-12
Define CONVERSION_STATUS only if it wasn't defined earlier

balasankarc:
Note: This PR targets 14-stable branch, as it is looking at fixing a problem in Chef-14

Closes: https://github.com/chef/ohai/issues/1276

When ohai is reloaded, Ruby warns about `CONVERSION_STATUS` constant being already defined. This PR fixes those warnings.

## Types of changes
- [x] Chore (non-breaking change that does not add functionality or fix an issue)

## Checklist:

- [x] I have read the **CONTRIBUTING** document.
- [ ] I have run the pre-merge tests locally and they pass.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [x] All commits have been signed-off for [the Developer Certificate of Origin](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).

PR #1377, opened 2019-06-14
Fix openSUSE leap and SLES detection in Chef 14

balasankarc:
Detection of openSUSE leap and SLES distributions fail with Chef 14 because these distributions satisfy the `if File.exist?('/usr/lib/os-release')` check. However, in that check, we only consider Clear Linux and not anything else, thus causing detection to fail for any other OS.

In my experiments, all the Clear Linux distributions provided  `/etc/os-release` file also. So, moving Clear Linux detection to be based on that file, and using `/usr/lib/os-release` only if `/etc/os-release` is not found (to handle any weird edge cases out there) seems like a safe option to handle Clear Linux, openSUSE Leap and SLES.

## Related Issue

https://github.com/chef/ohai/issues/1343

## Types of changes

## Checklist:

### DIFF
--- a/lib/ohai/plugins/linux/platform.rb
+++ b/lib/ohai/plugins/linux/platform.rb
@@ -266,7 +266,8 @@ Ohai.plugin(:Platform) do
     elsif File.exist?("/etc/alpine-release")
       platform "alpine"
       platform_version File.read("/etc/alpine-release").strip
-    elsif File.exist?("/usr/lib/os-release")
+    # If /etc/os-release file exists, we take that as source of truth
+    elsif File.exist?("/usr/lib/os-release") && !File.exist?("/etc/os-release")
       contents = File.read("/usr/lib/os-release")
       if /Clear Linux/ =~ contents
         platform "clearlinux"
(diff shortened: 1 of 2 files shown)


lib/ohai/plugins/linux/platform.rb
```diff
@@ -266,7 +266,8 @@ def determine_platform_family
     elsif File.exist?("/etc/alpine-release")
       platform "alpine"
       platform_version File.read("/etc/alpine-release").strip
-    elsif File.exist?("/usr/lib/os-release")
+    # If /etc/os-release file exists, we take that as source of truth
+    elsif File.exist?("/usr/lib/os-release") && !File.exist?("/etc/os-release")
```
balasankarc: @tas50 Does this make sense? ` && !File.exist?("/etc/os-release")` essentially ensures openSUSE Leap and SLES not match with this check.

balasankarc: CI failures doesn't seem related to my change. :thinking:

tas50:
Thanks @balasankarc

balasankarc:
@tas50 Thanks for merging. Would it be possible to cut a new 14.x release with this PR and https://github.com/chef/ohai/pull/1375 included?

balasankarc:
> @tas50 Thanks for merging. Would it be possible to cut a new 14.x release to Rubygems with this PR and #1375 included?

I see 14.14.0 has been released. Thanks. Much appreciated.

Issue #1390, opened 2019-07-29
Openstack plugin doesn't detect windows


jjustice6:
### Describe the Enhancement:

If you are using Windows on openstack, the openstack plugin fails to detect that it is using openstack unless you use hints due to the lack of `dmi` on Windows.
It is possible to work around this using `'wmi-lite'` via data contained in `wmi.first_of('Win32_ComputerSystem')`, specifically, the 'manufacturer' and 'model' fields.

### Describe the Need:
Any openstack user.

### Current Alternative
Throwing an empty file into C:\chef\ohai\hints\openstack.json.

### Can We Help You Implement This?:

I believe I can send a patch for this in the near future.

PR #1391, opened 2019-07-29
Use virtualization plugin to detect if we're on openstack on Windows.

jjustice6:
Signed-off-by: Joshua Justice <jjustice6@bloomberg.net>

## Description

If you are using Windows on openstack, the openstack plugin fails to detect that it is using openstack unless you use hints due to the lack of `dmi` on Windows.
It is possible to work around this using `'wmi-lite'` via data contained in `wmi.first_of('Win32_ComputerSystem')`, specifically, the 'manufacturer' and 'model' fields, or data already collected by Ohai elsewhere.

## Related Issue

https://github.com/chef/ohai/issues/1390

## Types of changes

## Checklist:

- [X] I have updated the documentation accordingly.

### DIFF
--- a/lib/ohai/plugins/openstack.rb
+++ b/lib/ohai/plugins/openstack.rb
@@ -25,6 +25,7 @@ Ohai.plugin(:Openstack) do
   provides "openstack"
   depends "dmi"
   depends "etc"
+  depends "virtualization"
 
   # do we have the openstack dmi data
   def openstack_dmi?
@@ -38,6 +39,14 @@ Ohai.plugin(:Openstack) do
     else
       logger.trace("Plugin Openstack: has_openstack_dmi? == false")
       false
+    end
+  end
+
+  # use virtualization data
+  def openstack_virtualization?
+    if get_attribute(:virtualization, :system, :guest) == "openstack"
+      logger.trace("Plugin Openstack: has_openstack_virtualization? == true")
+      true
     end
   end
 
@@ -61,7 +70,7 @@ Ohai.plugin(:Openstack) do
 
   collect_data do
     # fetch data if we look like openstack
-    if openstack_hint? || openstack_dmi?
+    if openstack_hint? || openstack_dmi? || openstack_virtualization?
       openstack Mash.new
       openstack[:provider] = openstack_provider
 
--- a/spec/unit/plugins/openstack_spec.rb
+++ b/spec/unit/plugins/openstack_spec.rb
@@ -302,4 +302,24 @@ describe Ohai::System, "plugin openstack" do
       end
     end
   end
+
+  context "when openstack virtualization is present" do
+    context "and the metadata service is not available" do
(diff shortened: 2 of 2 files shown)


jjustice6:
I'm not sure if this is how Chef expects this to be structured, or if I should make an all-new plugin in the windows/ folder for this purpose.

lib/ohai/plugins/openstack.rb
```diff
@@ -41,6 +42,27 @@ def openstack_dmi?
     end
   end
 
+  # if we're on windows, use wmi since it doesn't have dmi
+  def openstack_wmi?
+    if os == "windows"
+      require "wmi-lite"
```
jjustice6: Note to self: poking around, I see I should be using `collect_os` here, and my require should be "wmi-lite/wmi"

jjustice6: Or should it be a separate `collect_data(:windows)`?

tas50:
I think you might want to take a look at depending on the virtualization plugin and then running if that plugin detected the openstack hypervisor. That plugin has pretty similar logic to this already, but generalized for all the different manufacturer IDs

https://github.com/chef/ohai/blob/master/lib/ohai/plugins/windows/virtualization.rb

https://github.com/chef/ohai/blob/master/lib/ohai/mixin/dmi_decode.rb

jjustice6:
Makes sense, will do.

jjustice6:
I've rewritten it to use the info from the virtualization plugin, and added a test accordingly. (A basic one, since there's no need to re-test all the other stuff.)

lib/ohai/plugins/openstack.rb
```diff
@@ -61,7 +70,7 @@ def openstack_provider
 
   collect_data do
     # fetch data if we look like openstack
-    if openstack_hint? || openstack_dmi?
+    if openstack_hint? || openstack_dmi? || openstack_virtualization?
```
tas50: I think this DMI check method becomes redundant as long as the virtualization plugin contains all the same manufacturer IDs. I’m on thumbs at the moment so I’ll check on this when I get to a real computer

jjustice6: I've got no problem making that change in this PR, I can even move the tests to make that work. Just let me know.

jjustice6: I think it's fine, since it calls the same function under the hood: https://github.com/chef/ohai/blob/master/lib/ohai/plugins/linux/virtualization.rb#L135 - and I've confirmed a linux openstack box does report `"system":"openstack"`.

tas50: I'll clean it up post merge. There's just no point in the extra logic now that they use the same logic under the hood.

jjustice6: Sure, thanks.

PR #1394, opened 2019-08-02
Simplify the Openstack plugin by not polling DMI data

tas50:
Since we rely on the Virtualization plugin now there's no reason to also poll the DMI plugin data. That's a waste of time since we'll have already matched if virtualization came back Openstack. I didn't move the manufacturer = Openstack detection since that seems wrong on all modern Openstack installs which identify product field: https://cloudinit.readthedocs.io/en/latest/topics/datasources/openstack.html


tas50:
This needs the specs refactored, but my battery is dead so that's for another day.

PR #1395, opened 2019-08-03
Simplify and fix Openstack detection on Linux

tas50:
Don't require the Etc plugin for detection of Dreamhost which was causing this plugin to run on all hosts and make Ohai take several minutes on large AD domains.

Fix detection of openstack on non-Redhat based Openstack setups.


tas50:
This needs test updates for this new functionality

tas50:
Tag @teknofire

tas50:
This breaks windows so I need to do some more testing and come up with a solution. I'll probably just assume openstack when on windows.

tas50:
Windows support fixed by bailing early

PR #1397, opened 2019-08-08
Use Benchmark.realtime for the main application time as well

tas50:
We probably want this in both locations


tas50:
Similar to #1397 1392

tas50:
Approved on Slack

PR #1412, opened 2019-11-22
Fix failures under Ruby 2.7

KrisShannon:
Ruby 2.7 removes the scanf library which is used in the network plugins for darwin and solaris to parse the hex netmask into a dotted decimal

Replaced the scanf library with simple regex and to_i(16) calls

## Related Issue
Pull #1410 added ruby 2.7 to the testing matrix which is now causing failures

## Types of changes

## Checklist:

### DIFF
--- a/lib/ohai/plugins/darwin/network.rb
+++ b/lib/ohai/plugins/darwin/network.rb
@@ -17,8 +17,12 @@
 #
 
 Ohai.plugin(:Network) do
+  require_relative "../../mixin/network_helper"
+
   provides "network", "network/interfaces"
   provides "counters/network", "counters/network/interfaces"
+
+  include Ohai::Mixin::NetworkHelper
 
   def parse_media(media_string)
     media = {}
@@ -87,8 +91,6 @@ Ohai.plugin(:Network) do
   end
 
   collect_data(:darwin) do
-    require "scanf"
-
     network Mash.new unless network
     network[:interfaces] ||= Mash.new
     counters Mash.new unless counters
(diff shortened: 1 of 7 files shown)


KrisShannon:
I'm not sure about the code duplication for parsing the hex netmasks.

Should this be extracted and if so where should I put it?

jaymzh:
I would probably rename `mixin/network_constants.rb` to `mixin/network_helper.rb` (ala `http_helper`), rename the module accordingly, and then add your method in there. Seem sane @tas50

lib/ohai/plugins/darwin/network.rb
```diff
@@ -50,6 +50,14 @@ def parse_media(media_string)
     media
   end
 
+  def darwin_parse_netmask(netmask)
```
tas50: Can you toss a YARD comment on these explaining what the arg and return is for anyone that needs to touch it in the future. Also thanks for keeping the names unique so we avoid collisions in the future.

jaymzh: @tas50  I"m pretty sure there these are the same, and thus should not have different names and should not be in different places, no?

tas50:
it would probably be nice to extract this. Short of that the unique names avoid the problems we had with def collisions in the past.

KrisShannon:
I like the idea of moving it into a renamed `mixin/network_helper.rb`

I'll give that a go.

KrisShannon:
I found another existing method `hex_to_dec_netmask` in [plugins/aix/network.rb](https://github.com/chef/ohai/blob/v15.6.0/lib/ohai/plugins/aix/network.rb#L26-L32)

I'll extract and patch that instead.

lib/ohai/plugins/darwin/network.rb
```diff
@@ -50,6 +50,14 @@ def parse_media(media_string)
...
+    if netmask =~ /^([0-9a-f]{2})([0-9a-f]{2})([0-9a-f]{2})([0-9a-f]{2})$/
+      return [$1, $2, $3, $4].map { |hex| hex.to_i(16) } * "."
```
jaymzh: btw. `join` is a much clearer than array-repetition-bizarre-side-use-if-it-isnt-a-number syntax :)

KrisShannon:
What should I do about the test of the `hex_to_dec_netmask` helper in `spec/unit/plugins/aix/network_spec.rb`?

It's no longer specific to AIX so should I create a new `spec/unit/mixin/network_helper.rb` and move the test across to it?

KrisShannon:
I created a `spec/unit/mixin/network_helper.rb` and moved the aix plugin test across to it but it doesn't look like it got executed by rspec on buildkite.

Is there something else that needs to be updated so it will run the new spec file?  I tried grepping around for the `command_spec` and the `azure_metadata_spec` which are bring run but I couldn't find anything.

I did notice that `spec/unit/mixin/dmi_decode.rb` also doesn't seem to be running.

lib/ohai/plugins/darwin/network.rb
KrisShannon: Yeah,  I agree.

The scanf line was using `... * "."` and I just didn't change it when I was editing to remove the scanf.

The new helper from AIX is completely different so ¯\_(ツ)_/¯

jaymzh:
@KrisShannon - I believe the reason those aren't working is because they don't end in `_spec.rb`

KrisShannon:
Doh!  Of course...

jaymzh:
Don't feel bad, at least you noticed yours wasn't running. Whoever wrote `dmi_decode.rb` never noticed it didn't run. :) We all miss stuff. :)

KrisShannon:
Ok.  Final push I hope.

Test file is renamed and buildkite has run it succesfully

jaymzh:
Oh, @KrisShannon you already opened a PR for that too! Sorry I missed that. Hadn't had coffee yet. :) <3 Thanks!

tas50:
Thanks @KrisShannon. You saved us some work next April when we ship with ruby 2.7

jaymzh:
@tas50 - now that you merged this you can merge https://github.com/chef/ohai/pull/1267 :)

Issue #1271, opened 2018-10-10
Chef 15: Write filesystem2 AIX/Solaris data to filesystem and filesystem2


tas50:
We're unifying the disk data and for Chef 14.6 we introduced filesystem for AIX and Solaris. In Chef 15 we'll begin writing data to both locations but the data will be the filesystem2 data. In Chef 16 we'll nuke the filesystem2 namespace. Painful, but the least painful option.

btm:
Looks like the plan is now to remove the `filesystem2` namespace in Chef 17: 069149c0d

tas50:
This part is done

PR #1426, opened 2020-01-29
Replace filesystem with filesystem2 on aix/solaris/bsd

tas50:
We added filesystem2 plugin data mid way through the Chef 14 release
cycle. This converts the filesystem plugin to write data in the
filesystem2 format. In Chef 17 we'll remove those filesystem2 namespaces
and just write the data out in the filesystem2 format.


tas50:
Fixes https://github.com/chef/ohai/issues/1271

PR #1427, opened 2020-01-30
Add logic to fetch the latest azure metadata version

tas50:
This is inspired by the way we do this in EC2, but it's different in
many ways since Azure presents the latest version in a different way.

Additional values exposed by this:

```json
      "azEnvironment": "AzurePublicCloud",
      "customData": "",
      "plan": {
        "name": "",
        "product": "",
        "publisher": ""
      },
      "provider": "Microsoft.Compute",
      "publicKeys": [
        {
          "keyData": "ssh-rsa 123",
          "path": "/home/tsmith/.ssh/authorized_keys"
        }
      ],
      "resourceId": "/subscriptions/123/resourceGroups/tsmith/providers/Microsoft.Compute/virtualMachines/timtest",
    "storageProfile": {
        "dataDisks": [

        ],
        "imageReference": {
          "id": "",
          "offer": "UbuntuServer",
          "publisher": "Canonical",
          "sku": "18.04-LTS",
          "version": "latest"
        },
        "osDisk": {
          "caching": "ReadWrite",
          "createOption": "FromImage",
          "diffDiskSettings": {
            "option": ""
          },
          "diskSizeGB": "30",
          "encryptionSettings": {
            "enabled": "false"
          },
          "image": {
            "uri": ""
          },
          "managedDisk": {
            "id": "/subscriptions/123/resourceGroups/tsmith/providers/Microsoft.Compute/disks/timtest_OsDisk_1_f68c8aee869f4668aff1f12e449be5ac",
            "storageAccountType": "Premium_LRS"
          },
          "name": "timtest_OsDisk_1_f68c8aee869f4668aff1f12e449be5ac",
          "osType": "Linux",
          "vhd": {
            "uri": ""
          },
          "writeAcceleratorEnabled": "false"
        }
      },
      "tagsList": [

      ],
      "vmScaleSetName": "",
      "zone": ""
```

### DIFF
--- a/lib/ohai/mixin/azure_metadata.rb
+++ b/lib/ohai/mixin/azure_metadata.rb
@@ -50,7 +50,7 @@ module Ohai
           end
 
           # azure returns a list of the 3 latest versions it supports
-          versions = parse_json(response.body)["newest-versions"]
+          versions = parse_json(response.body).fetch("newest-versions", [])
           versions.sort!
 
           until versions.empty? || AZURE_SUPPORTED_VERSIONS.include?(versions.last)
--- a/lib/ohai/mixin/azure_metadata.rb
+++ b/lib/ohai/mixin/azure_metadata.rb
@@ -1,6 +1,6 @@
 #
 # Author:: Tim Smith (<tsmith@chef.io>)
-# Copyright:: Copyright 2017 Chef Software, Inc.
+# Copyright:: 2017-2020 Chef Software, Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


lib/ohai/mixin/azure_metadata.rb
```diff
@@ -19,30 +19,87 @@
...
+      AZURE_SUPPORTED_VERSIONS ||= %w{ 2017-04-02 2017-08-01 2017-12-01 2018-02-01 2018-04-02
+                                   2018-10-01 2019-02-01 2019-03-11 2019-04-30 2019-06-01
+                                   2019-06-04 2019-08-01 2019-08-15 2019-11-01 }.freeze
+
+      def best_api_version
+        @api_version ||= begin
+          logger.trace("Mixin AzureMetadata: Fetching http://#{AZURE_METADATA_ADDR}/metadata/instance to determine the latest supported metadata release")
+          response = http_get("/metadata/instance")
+          if response.code == "404"
+            logger.trace("Mixin AzureMetadata: Received HTTP 404 from metadata server while determining API version, assuming #{AZURE_SUPPORTED_VERSIONS[-1]}")
```
phiggins: nitpick: It could be confusing to someone not familiar with ruby to use `[-1]` on this line and `.first` on the next to refer to the same value.

lib/ohai/mixin/azure_metadata.rb
```diff
@@ -19,30 +19,87 @@
...
+            return AZURE_SUPPORTED_VERSIONS.last
+          elsif response.code != "400" # 400 is actually what we want
+            raise "Mixin AzureMetadata: Unable to determine Azure metadata version (returned #{response.code} response)"
+          end
+
+          # azure returns a list of the 3 latest versions it supports
+          versions = parse_json(response.body)["newest-versions"]
```
phiggins: nitpick: If the shape of this response body changes `versions` could be nil here. Would it be worth changing this to something like `.fetch("newest-versions", [])`?

lib/ohai/mixin/azure_metadata.rb
tas50: yeah I can make the consistent

tas50: I actually fixed the code to make it consistent, but forgot the log

lib/ohai/mixin/azure_metadata.rb
```diff
@@ -19,30 +19,87 @@
...
+      #
+      # @param [String] the relative uri to fetch from the Azure Metadata Service URL
+      #
+      # @return [Net::HTTP]
       def http_get(uri)
+        full_uri = URI.join("http://#{AZURE_METADATA_ADDR}", uri)
         conn = Net::HTTP.start(AZURE_METADATA_ADDR)
         conn.read_timeout = 6
-        conn.get(uri, { "Metadata" => "true" })
+        conn.get(full_uri, { "Metadata" => "true" })
```
phiggins: I would have expected this to work with just the path, does that not work?

lib/ohai/mixin/azure_metadata.rb
tas50: TIL. That makes way more sense

lib/ohai/mixin/azure_metadata.rb
```diff
@@ -19,30 +19,87 @@
...
-        response = http_get(AZURE_METADATA_URL)
+      # parse JSON data from a String to a Hash
+      #
+      # @param [String] response_body json as string to parse
+      #
+      # @return [Hash]
+      def parse_json(response_body)
+        data = StringIO.new(response_body)
+        parser = FFI_Yajl::Parser.new
+        parser.parse(data)
```
phiggins: FWIW, a string or StringIO works here.

lib/ohai/mixin/azure_metadata.rb
tas50: Azure requires you to set a header or it fails

lib/ohai/mixin/azure_metadata.rb
tas50: That's probably just cargo culting from ec2 metadata

lib/ohai/mixin/azure_metadata.rb
```diff
@@ -19,30 +19,87 @@
...
+
+      def best_api_version
+        @api_version ||= begin
+          logger.trace("Mixin AzureMetadata: Fetching http://#{AZURE_METADATA_ADDR}/metadata/instance to determine the latest supported metadata release")
+          response = http_get("/metadata/instance")
+          if response.code == "404"
+            logger.trace("Mixin AzureMetadata: Received HTTP 404 from metadata server while determining API version, assuming #{AZURE_SUPPORTED_VERSIONS.last}")
+            return AZURE_SUPPORTED_VERSIONS.last
+          elsif response.code != "400" # 400 is actually what we want
+            raise "Mixin AzureMetadata: Unable to determine Azure metadata version (returned #{response.code} response)"
```
phiggins: I'm really unfamiliar with the conventions of this project, but should this return a specific error type rather than the generic error?

lib/ohai/mixin/azure_metadata.rb
phiggins: I meant just the path rather than constructing the URL out of the host + path.

lib/ohai/mixin/azure_metadata.rb
tas50: I updated all 3 of the metadata mixins to use just String

lib/ohai/mixin/azure_metadata.rb
tas50: Most places in Ohai are just generic errors

lib/ohai/mixin/azure_metadata.rb
tas50: oh yep I fixed that

PR #1438, opened 2020-03-03
Drop unneeded require in ec2 plugin

davide125:

## Description
The ec2 plugin requires `http_helper`, but it doesn't seem to actually use it anywhere, so we might as well drop it. This caused issues for us while backporting this plugin to an older Chef version.

## Types of changes

## Checklist:


davide125:
Oh, apparently this breaks Windows. Still, it looks like the dependency should be in `ec2_metadata`, not `ec2`, given that's where the call it happening. Thoughts?

tas50:
The can_socket_connect? method used in ec2 is from the http helper so this is needed here.

PR #1446, opened 2020-04-06
Make shard plugin more resilient and throw better errors

### LINKED ISSUE #1385: Shard plugin no longer creating a seed on Amazon Linux 1
In Chef Infra Client 15.1 the Shard plugin no longer creating a seed on Amazon Linux 1

jaymzh:
* Catch a variety of problems and Ohai::Log.error about them. Since this
  is an optional plugin, if we're hitting these, someone specifically
  requested this plugin, so an error is warranted.
* We raise anyway, so that you don't spew further errors

Fixes #1385

### DIFF
--- a/lib/ohai/plugins/shard.rb
+++ b/lib/ohai/plugins/shard.rb
@@ -22,10 +22,12 @@ Ohai.plugin(:ShardSeed) do
 
   def get_dmi_property(dmi, thing)
     %w{system base_board chassis}.each do |section|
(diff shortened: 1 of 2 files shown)


tas50:
@jaymzh any chance you can take a stab at adding some tests for this?

jaymzh:
tests that it warns properly? Sure!

jaymzh:
Tests added!

tas50:
Thanks @jaymzh

PR #1454, opened 2020-04-21
Gather CPU info from lscpu and merge data into cpu attribute namespace

ramereth:
## Description

This contains two changes:

### Gather CPU info from lscpu and ~create~  merge data into cpu attribute namespace

This ~adds a cpu/lscpu node data name space~ merges data containing information from the lscpu command (if it exists) into the cpu namespace. This augments and standardizes the information you can gather from ``/proc/cpuinfo`` which can be an issue on non-x86 architectures such as ppc64le and aarch64. Namely, when trying to determine whether or not a node is a guest or not. In addition, this properly fixes cpu counts on s390x nodes and uses lscpu data for determining cpu/total, cpu/real and cpu/cores if lscpu exists.

### Fixes for detecting KVM guests on certain platforms and architectures

On some architectures (i.e. ppc64le), dmidecode does not exist nor does the proper information in ``/proc/cpuinfo`` exist to detect KVM guests. To workaround that, we can use lscpu data to easily determine this as a fallback.

In addition on CentOS 6, ``/sys/devices/virtual/misc/kvm`` does not exist so this works around that problem as well.

I don't believe this is a breaking change however please let me know if it is.

### TODO:
- [x] Additional tests for other architectures and platforms for ``lscpu``

## Types of changes

## Checklist:


ramereth:
@tas50 I'm going to start working on this again with the hopes of getting it included in Chef 17. I'm trying to remember where we left off on this. I seem to recall that you suggested instead of making a new `cpu/lscpu` namespace, to just replace it with the current `cpu` name space. Is that still the case? If so, do I need to replicate the current structure that duplicates all of the cpu attributes for every single core?

lib/ohai/plugins/cpu.rb
```diff
@@ -46,13 +47,211 @@ def parse_bsd_dmesg(&block)
...
+  def range_to_a(range)
+    range_array = []
+    range.split(",").each do |cpu|
+      if /\d+-\d+/.match?(cpu.to_s)
+        range_array << Range.new(*cpu.split("-").map(&:to_i)).to_a
+      else
+        range_array << cpu.to_i
+      end
+    end
+    range_array.flatten
```
lamont-granquist, line 59: you can probably `range.split(",").each_with_object([]) { |cpu| if...else..end }.flatten` (only probably multiline)

lib/ohai/plugins/cpu.rb
```diff
@@ -46,13 +47,211 @@ def parse_bsd_dmesg(&block)
...
+  end
+
+  def parse_lscpu(cpu_info)
+    lscpu_info = Mash.new
+    lscpu_info[:numa_node_cpus] = Mash.new
+    lscpu_info[:vulnerability] = Mash.new
+    begin
+      so = shell_out("lscpu")
+      cpu_cores = shell_out("lscpu -p=CPU,CORE,SOCKET")
+      if so.exitstatus == 0 && cpu_cores.exitstatus == 0
```
lamont-granquist, line 69: can remove this and just `rescue` (anything, including ShellCommandFailed)

lib/ohai/plugins/cpu.rb
```diff
@@ -46,13 +47,211 @@ def parse_bsd_dmesg(&block)
...
+            numa_node = $1
+            cpus = $2
+            lscpu_info[:numa_node_cpus][numa_node] = range_to_a(cpus)
+          when /^Vulnerability (.+?):\s+(.+)/ # https://rubular.com/r/aKtSD1ypUlKbGm
+            name = $1.strip.downcase.tr(" ", "_")
+            description = $2.strip
+            lscpu_info[:vulnerability][name] = Mash.new
+            lscpu_info[:vulnerability][name] = description
+          when /^Flags:\s+(.+)/
+            lscpu_info[:flags] = $1.split(" ").sort.to_a
```
jaymzh, line 164:
```suggestion
            lscpu_info[:flags] = $1.split(" ").sort
```

lib/ohai/plugins/cpu.rb
```diff
@@ -68,7 +267,14 @@ def parse_bsd_dmesg(&block)
...
       when /model\s+:\s(.+)/
-        cpuinfo[current_cpu]["model"] = $1
+        model = $1
+        # ppc has "model" at the end of /proc/cpuinfo. In addition it should always include a include a dash or "IBM".
+        # So let's put this in cpu/model on ppc
+        if model.match?(/-|IBM/)
+          cpuinfo["machine_model"] = model
+        else
+          cpuinfo[current_cpu]["model"] = model
+        end
```
jaymzh, line 277:
```suggestion
        cpuinfo[current_cpu]["model"] = model
        if model.match?(/-|IBM/)
          cpuinfo["machine_model"] = model
        end
```

lib/ohai/plugins/cpu.rb
ramereth: Am I just changing `.each` with `.each_with_object` or is there something more I need to do? If so, chefstyle complains if I do that.

ramereth:
@jaymzh as requested, you can find my output gists here: https://gist.github.com/ramereth/6b9fdc88ba36af7aa3cc9f14b363e34a

@lamont-granquist @tas50 @jaymzh please have another look at this

lib/ohai/plugins/cpu.rb
jaymzh: The only change here is that you don't have to pre-define range_array, but I think this is fine as-is

lib/ohai/plugins/cpu.rb
jaymzh: I agree, @ramereth  but I'll see if there's something @lamont-granquist saw that I"m missing

lib/ohai/plugins/cpu.rb
jaymzh: So it could be this:
```
def range_to_a(range)
  range.split(",").each_with_object([]) do |cpu, arr|
    if /\d+-\d+/.match?(cpu.to_s)
      arr << Range.new(*cpu.split("-").map(&:to_i)).to_a
    else
      arr << cpu.to_i
    end
  end.flatten
end
```
is Lamont's point.

But also, this  comment is inaccurate. `range` is a thing in ruby and you can't pass this a range.  It takes a _string_ that looks like a range.

jaymzh (approved):

PR #1467, opened 2020-05-25
Convert some specs to use let

tas50:
Modernize our specs a bit. There's tons of other issues, but this gets us into a slightly better state.


tas50:
This is going to require a lot of extra fixes. Waiting on the next time I've very bored.

Issue #1469, opened 2020-06-23
node['platform'] will need to be updated for macOS 10.16


chilcote:
## Description
https://github.com/chef/ohai/blob/master/lib/ohai/plugins/darwin/platform.rb#L26-L30
Apple changed the output of `sw_vers -productName` from 
```
Mac OS X
```
to:
```
macOS
```

## Ohai Version

15.7.4, but it looks the same in 16.2.0.

## Platform Version

macOS 10.16 beta 1

## Ohai Output

n/a

kmf:
According to Apple's pages it will be called `macOS 11`

btm:
https://developer.apple.com/documentation/macos-release-notes/macos-big-sur-11-beta-release-notes
> The system ProductVersion is reported as 11.0 only on the Developer Transition Kit at this time. Existing Mac systems will temporarily report the value as 10.16.

lamont-granquist:
this is done

PR #1470, opened 2020-06-23
Make sure Darwin hosts are always mac_os_x

tas50:
mac_os_x_server isn't a thing anymore and no third party darwin based
operating systems ever took off. Fix this detectio and speed things up
a tiny bit by assuming darwin == mac_os_x just like we already do with
the platform_family

### DIFF
--- a/lib/ohai/plugins/darwin/hardware.rb
+++ b/lib/ohai/plugins/darwin/hardware.rb
@@ -41,8 +41,8 @@ Ohai.plugin(:Hardware) do
     hardware.merge!(hw_hash[0]["_items"][0])
 
     # ProductName:	Mac OS X
-    # ProductVersion:	10.12.5
-    # BuildVersion:	16F73
+    # ProductVersion:	10.15.6
+    # BuildVersion:	19G46c
     shell_out("sw_vers").stdout.lines.each do |line|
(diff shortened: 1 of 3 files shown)


tas50:
@chilcote does this look like a valid way to resolve #1469. Correct me if I'm wrong, but there isn't a supported MacOS X Server anymore so we shouldn't have to worry about that distinction. I think that's the only historical reason we had this other than some 2010-era hope that someone might come out with a Darwin mac clone OS.

chilcote:
Yep, macOS Server is just an app now. This should work, thanks! (will this be back-ported to chef15?)

tas50:
Yeah I'll kick this back to Ohai 15 and make sure we get a chef release going

erikng:
When this lands, pretty please make a new chef installer that supports 10.16.

Maybe another chef 16.2 build?

tas50:
We need 10.16 builders in Mac Stadium first. That's not going to happen overnight, but our releng folks know it's a request.

erikng:
@tas50 honestly that shouldn't matter too much. Build one for 10.15 and we can use that for now :).

We don't ship any of the "OS specific" versions of chef anyway.

Issue #1474, opened 2020-06-24
Unable to handle Linux IPv4 routes with IPv6 next hops - breaking network interface data


cooperlees:
## Description

When I add IPv4 routes with IPv6 next hops (available since early Linux 5.x releases) ohai has a bad time and can not populate the network interfaces JSON in the ohai output. I'd expect (but have not dug in yet as my ruby skills are low) some parsing is breaking somewhere.

Example:
```console
[cooper:~]$ ip -4 route add default nexthop via inet6 fe80::69 dev eth0

[cooper:~]$ ip -4 route
default via inet6 fe80::69 dev eth0
```

## Ohai Version

```console
[cooper:~]$ /opt/chef/embedded/bin/ohai --version
Ohai: 13.10.0
```

- This **also** repros on *master*

## Platform Version

Linux aserver.facebook.com 5.2.9 #1 SMP Mon Mar 23 09:04:23 PDT 2020 x86_64 x86_64 x86_64 GNU/Linu

## Ohai Output

**Bad** (after route added) missing data:

```
  "network": {
    "interfaces": {

    },
    "default_interface": "eth0",
    "default_gateway": "inet6"
  },
```

**Good** *(expected)*:

```
  "network": {
    "interfaces": {
      "lo": {
        "mtu": "65536",
        "flags": [
          "LOOPBACK",
          "UP",
          "LOWER_UP"
        ],
        "encapsulation": "Loopback",
        "addresses": {
          "127.0.0.1": {
            "family": "inet",
            "prefixlen": "8",
            "netmask": "255.0.0.0",
            "scope": "Node"
          },
          "10.127.30.4": {
            "family": "inet",
            "prefixlen": "32",
            "netmask": "255.255.255.255",
            "scope": "Global"
          },
          "2401:db00:eef0:b69::69": {
            "family": "inet6",
            "prefixlen": "64",
            "scope": "Global",
            "tags": [

            ]
          },
          "::1": {
            "family": "inet6",
            "prefixlen": "128",
            "scope": "Node",
            "tags": [

            ]
          }
        },
        "state": "unknown",
        "routes": [
          {
            "destination": "::1",
            "family": "inet6",
            "metric": "256",
            "proto": "kernel"
          },
          {
            "destination": "unreachable",
            "family": "inet6",
            "metric": "1024"
          },
          {
            "destination": "unreachable",
            "family": "inet6",
            "metric": "1024"
          },
          {
            "destination": "unreachable",
            "family": "inet6",
[...]

PR #1475, opened 2020-06-24
Linux network plugin: Handle IPv6 next hops for IPv4 routes

cooperlees:
## Description
- Add handling for `via` where the inet6 address family is specified
- Add rspec test to show working

I suspect this only fixes part of the issue for #1474 

## Related Issue
Issue: https://github.com/chef/ohai/issues/1474

## Types of changes
- [x ] Bug fix (non-breaking change which fixes an issue)

Signed-off-by: Cooper Lees <me@cooperlees.com>

### DIFF
--- a/lib/ohai/plugins/linux/network.rb
+++ b/lib/ohai/plugins/linux/network.rb
@@ -108,6 +108,7 @@ Ohai.plugin(:Network) do
           # http://rubular.com/r/pwTNp65VFf
           route_entry[k] = $1 if route_ending =~ /\b#{k}\s+([^\s]+)/
         end
+        # https://rubular.com/r/k1sMrRn5yLjgVi
         route_entry["via"] = $1 if route_ending =~ /\bvia\s+inet6\s+([^\s]+)/
 
         # a sanity check, especially for Linux-VServer, OpenVZ and LXC:
--- a/lib/ohai/plugins/linux/network.rb
+++ b/lib/ohai/plugins/linux/network.rb
@@ -105,7 +105,7 @@ Ohai.plugin(:Network) do
         route_entry = Mash.new(destination: route_dest,
                                family: family[:name])
         %w{via scope metric proto src}.each do |k|
-          if k == 'via' && route_ending =~ /\bvia\sinet6\s+([^\s]+)/
+          if k == "via" && route_ending =~ /\bvia\sinet6\s+([^\s]+)/
             route_entry[k] = $1
             next
           end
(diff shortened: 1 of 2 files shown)


lib/ohai/plugins/linux/network.rb
```diff
@@ -105,6 +105,10 @@ def check_routing_table(family, iface, default_route_table)
         route_entry = Mash.new(destination: route_dest,
                                family: family[:name])
         %w{via scope metric proto src}.each do |k|
+          if k == "via" && route_ending =~ /\bvia\sinet6\s+([^\s]+)/
+            route_entry[k] = $1
+            next
+          end
           # http://rubular.com/r/pwTNp65VFf
           route_entry[k] = $1 if route_ending =~ /\b#{k}\s+([^\s]+)/
         end
```
phiggins, line 114: There doesn't seem to be a reason this new code needs to be inside the loop.

```suggestion
        %w{via scope metric proto src}.each do |k|
          # http://rubular.com/r/pwTNp65VFf
          route_entry[k] = $1 if route_ending =~ /\b#{k}\s+([^\s]+)/
        end
        route_entry["via"] = $1 if route_ending =~ /\bvia\sinet6\s+([^\s]+)/          
```

phiggins: The other regex uses `\s+` for whitespace so the new regex probably should as well.

lib/ohai/plugins/linux/network.rb
```diff
@@ -105,6 +105,10 @@ def check_routing_table(family, iface, default_route_table)
         route_entry = Mash.new(destination: route_dest,
                                family: family[:name])
         %w{via scope metric proto src}.each do |k|
+          if k == "via" && route_ending =~ /\bvia\sinet6\s+([^\s]+)/
+            route_entry[k] = $1
+            next
```
tas50, line 110: Couldn't this just be an if / else and avoid the next?

lib/ohai/plugins/linux/network.rb
```diff
@@ -108,6 +108,7 @@ def check_routing_table(family, iface, default_route_table)
           # http://rubular.com/r/pwTNp65VFf
           route_entry[k] = $1 if route_ending =~ /\b#{k}\s+([^\s]+)/
         end
+        route_entry["via"] = $1 if route_ending =~ /\bvia\s+inet6\s+([^\s]+)/
```
tas50, line 111: can you add a rubular permalink like the one above for this regex so we can see exactly how it's matching?

cooperlees: Don't the unittests show that? And even show them much clearer. I feel that's way more useful than a URL comment ... But I can add, will try work it out.

tas50:
Thanks for the updates @cooperlees

PR #1480, opened 2020-07-08
Remote Ohai with Target Mode

lamont-granquist:
🚧 largely for discussion at this point 🚧


lib/ohai/mixin/os.rb
```diff
@@ -23,40 +23,36 @@ module Ohai
...
+      def os_family
+        connection.os.family
+      end
+
+      # Using train determine the platform we're running on and its ancestors
+      #
+      # @return [Array] the OS family hierarchy
+      def os_hierarchy
+        connection.os.family_hierarchy
       end
```
lamont-granquist, line 51: So this is a big old "nope nope nope".

Train's ideal of families and such is based off of serverspec and not ohai which is totally incompatible.  This would be a MASSIVE breaking change.

This was one of the major blockers to ohbye.  You have to leverage the train stuff only to replace the :RbConfig::CONFIG bootstrapping, then you have to run the normal platform plugins over train to populate the platform + platform_family.

lib/ohai/util/file_helper.rb
```diff
@@ -16,23 +16,39 @@
...
           end
         end
+
         logger.trace("Plugin #{name}: did not find #{cmd}")
         false
       end
+
+      def file_exist?(path)
+        connection.file(path).exist?
+      end
```
lamont-granquist, line 47: So instead of this, ohai should be using chef-util to get these helper functions:

https://github.com/chef/chef/blob/master/chef-utils/lib/chef-utils/dsl/train_helpers.rb

There's also a train-compatible which helper:

https://github.com/chef/chef/blob/master/chef-utils/lib/chef-utils/dsl/which.rb

Those should get injected here, with some "dependency injection" to wire up the location of the
train transport.  What I came up with is slightly janky but I documented it here:

https://github.com/chef/chef/tree/master/chef-utils#documentation-for-software-developers

The next big bit is that the shell_out helper inside of chef-client itself needs to get moved into chef-utils (and some of the tight coupling to Chef::Config and Chef::Log (and any other internal constructs, the timeout stuff in particular my be problematic) needs to be made optional / dep-injected), then the implementation of shell_out in ohai needs to go away and it just needs to get it from chef-utils:

https://github.com/chef/chef/blob/master/lib/chef/mixin/shell_out.rb

So for example you probably want to have a `def __config` similarly to `def __transport_connection` in here:

https://github.com/chef/chef/blob/master/chef-utils/lib/chef-utils/internal.rb

I don't know what that is by default, you may have to raise NotImplementedError and then in the chef-client classes override that to be Chef::Config (otherwise you take a circular dep on the Chef class).  You can try to use the same trick I did with protecting the Chef.run_context and having the Chef stuff be wired up automatically, but honestly I'm starting to hate that now that I'm looking at it with fresh eyes.  That injection wiring probably needs to be moved into the chef gem and everything there should throw some notimplerror.

So anyway then in ohai you'd include the helpers into a module in ohai, then override `def __config` and def __transport_connection`.

lamont-granquist: I'm going to go do some work now on the chef-side to try to extract the shell_out mixin properly

tas50: This was one of the concerns I wrote down reading this over this morning. We want to make sure we sane paths added when we shell out or Ohai is going to stop returning valid data on EL platforms. We either need to hack that back into this helper or we can use chef-utils to get that via which.

lib/ohai/plugins/ohai_remote.rb
```diff
@@ -0,0 +1,36 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+Ohai.plugin(:OhaiRemote) do
+  provides "ohai_remote"
```
tas50, line 20: I dig this plugin. Nice

tas50:
I think there's more platform conversions that are going to have to take place for this to work. Train is a weird combination of Chef 12 era Ohai platforms and ServerSpec. We've continued to update platform families on the Ohai side and those changes weren't made in Train. That was one of the original concerns with using Train for Ohai. You'll want to make sure that Fedora and Amazon linux are getting the correct families as well.  

In general Ohai has much more advanced platform detection for Linux and we'll need to update Train to use some of that same logic. It parses out /etc/os-release when available and skips over all the legacy configs unless it needs to. The switch from the legacy configs in Ohai to os-release, not only sped up platform detection, but made it nice and future proof as any Linux platform was automatically detected.

lamont-granquist:
IMO we don't do that.

Just use train's detection to replace the RBConfig bootstrapping.  Then launch the plugins normally.  Don't modify train at all to get this going.

If we wind up blocking this on sorting out the train-vs-ohai platform detection then this will never happen.

If train and ohai get sorted out at some point in the future, then it'll be straightforwards to clean it up here, so you can just push it all into train and then use it in ohai (and obviously this then means that you need good testing so that no merged PR into train can break and change platform detection in ohai).

lib/ohai/util/file_helper.rb
lamont-granquist: https://github.com/chef/chef/pull/10137

README.md
```diff
@@ -21,6 +21,18 @@ Chef distributes ohai as a RubyGem. This README is for developers who want to mo
 - General documentation: <https://docs.chef.io/ohai.html>
 - Writing Ohai Plugins documentation: <https://docs.chef.io/ohai_custom.html>
 
+## Remote Ohai
+
+Ohai can be executed on remote systems to collect data on external systems:
```
kagarmoe, line 26: Use `target` mode to execute Ohai and collect data on external systems:

README.md
```diff
@@ -21,6 +21,18 @@ Chef distributes ohai as a RubyGem. This README is for developers who want to mo
...
+
+```shell
+ohai --target ssh://user@198.51.100.23/?key_files=/etc/chef/remote_ohai.pem
+```
+
+This is using Train backends, which include SSH, WinRM, Docker and local execution by default. Third party Train plugins can be used to extend support.
```
kagarmoe, line 32:
```suggestion
Targeting uses Train backends, which include SSH, WinRM, Docker and local execution by default. In addition, you can use third-party Train plugins to extend support.
```

README.md
```diff
@@ -21,6 +21,18 @@ Chef distributes ohai as a RubyGem. This README is for developers who want to mo
...
+
+The URL and query parameters depend on the used backend.
+
 ## Development Environment:
 
 Ohai's development dependencies should be installed with bundler. Just run `bundle install` in the root of the repo.
```
kagarmoe, line 38:
```suggestion
Install Ohai's development dependencies with bundler. Just run `bundle install` in the root of the repository.
```

lib/ohai/dsl/plugin.rb
```diff
@@ -98,6 +100,14 @@ def initialize(data, logger)
         @failed = false
       end
 
+      def connection
+        @connection ||= Train.create("local")
+      end
+
+      def remote_ohai?
+        !@connection.uri.start_with? "local"
```
phiggins, line 108:
```suggestion
        !connection.uri.start_with? "local"
```

lib/ohai/mixin/command.rb
```diff
@@ -19,33 +19,28 @@
...
+        logger.trace("Plugin #{name}: ran '#{cmd}' and returned #{execution.exit_status}")
+
+        OpenStruct.new({
+          stdout: execution.stdout,
+          stderr: execution.stderr,
+          exit_status: execution.exit_status,
+
+          # Compatibility for old shell_out usage
+          exitstatus: execution.exit_status,
+        })
```
phiggins, line 39: Since this has a static set of keys, it'd be better to use a `Struct` here since `OpenStruct` will respond to any method name.

lib/ohai/plugins/ssh_host_key.rb
```diff
@@ -49,35 +49,33 @@ def extract_keytype?(content)
...
-            keys[:ssh]["host_#{key_type}_public"] = content[1] unless key_type.nil?
-            keys[:ssh]["host_#{key_type}_type"] = key_subtype unless key_subtype.nil?
-          end
+      file_read(sshd_config).each_line do |line|
+        if /^hostkey\s/i.match?(line)
+          pub_file = "#{line.split[1]}.pub"
+          content = file_read(pub_file).split
+          key_type, key_subtype = extract_keytype?(content)
+          keys[:ssh]["host_#{key_type}_public"] = content[1] unless key_type.nil?
+          keys[:ssh]["host_#{key_type}_type"] = key_subtype unless key_subtype.nil?
```
phiggins, line 58:
```suggestion
      file_read(sshd_config).each_line.grep(/^hostkey\s/i) do |line|
        pub_file = "#{line.split[1]}.pub"
        content = file_read(pub_file).split
        key_type, key_subtype = extract_keytype?(content)
        keys[:ssh]["host_#{key_type}_public"] = content[1] unless key_type.nil?
        keys[:ssh]["host_#{key_type}_type"] = key_subtype unless key_subtype.nil?
```

lib/ohai/system.rb
```diff
@@ -69,6 +68,7 @@ def reset_system
 
       @loader = Ohai::Loader.new(self)
       @runner = Ohai::Runner.new(self, true)
+      @runner.backend = backend unless @backend
```
phiggins, line 71: Why not just `@runner.backend = backend`?

lib/ohai/mixin/command.rb
tecracer-theinen: Makes sense. I admit, I don't work too much with Struct/OpenStruct.

lib/ohai/system.rb
tecracer-theinen: I think this was related to Chef Target mode. When you start Ohai standalone, it's ok because it creates a remote backend in that case. If you use Chef TM, there was some Ohai reset call which would overwrite the pre-initialized backend passed by Chef with a new, local one. I'd have to test if this holds true or if my memory is off.

tas50:
Closing in favor of #1503

PR #1481, opened 2020-07-14
Enable frozen string literals throughout

tas50:
This doesn't actually get us any memory allocation improvements in this project, but it's going to be the default eventually so it gets us ready for that. There was one "" that needed to become a String.new instead to avoid failures.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,10 +1,11 @@
+# frozen_string_literal: true
 source "https://rubygems.org"
 
 gemspec
 
 # NOTE: do not submit PRs to add pry as a dep, add to your Gemfile.local
 group :development do
-  gem "chefstyle", git: "https://github.com/chef/chefstyle.git", branch: "master"
+  gem "chefstyle", "= 1.1.3"
   gem "rake", ">= 10.1.0"
   gem "rspec-core", "~> 3.0"
   gem "rspec-expectations", "~> 3.0"
--- a/lib/ohai/loader.rb
+++ b/lib/ohai/loader.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 #
 # Author:: Claire McQuin (<claire@chef.io>)
 # Copyright:: Copyright (c) 2013-2019, Chef Software Inc.
@@ -98,7 +99,7 @@ module Ohai
     # @return [Object] class object for the ohai plugin defined in the file
     def load_plugin_class(plugin_path)
       # Read the contents of the plugin to understand if it's a V6 or V7 plugin.
-      contents = ""
+      contents = String.new
       begin
         logger.trace("Loading plugin at #{plugin_path}")
         contents << IO.read(plugin_path)
--- a/lib/ohai/plugins/aix/virtualization.rb
+++ b/lib/ohai/plugins/aix/virtualization.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 #
 # Author:: Julian C. Dunn (<jdunn@chef.io>)
 # Author:: Isa Farnik (<isa@chef.io>)
--- a/lib/ohai/plugins/aix/virtualization.rb
+++ b/lib/ohai/plugins/aix/virtualization.rb
@@ -2,7 +2,7 @@
 #
 # Author:: Julian C. Dunn (<jdunn@chef.io>)
 # Author:: Isa Farnik (<isa@chef.io>)
-# Copyright:: Copyright (c) 2013-2016 Chef Software, Inc.
+# Copyright:: Copyright (c) Chef Software, Inc.
 # License:: Apache License, Version 2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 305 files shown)


Gemfile
```diff
@@ -1,10 +1,11 @@
+# frozen_string_literal: true
 source "https://rubygems.org"
 
 gemspec
 
 # NOTE: do not submit PRs to add pry as a dep, add to your Gemfile.local
 group :development do
-  gem "chefstyle", git: "https://github.com/chef/chefstyle.git", branch: "master"
+  gem "chefstyle", "= 1.1.3"
```
tas50, line 8: Using chefstyle from git just resulted in constant errors locally and the inability to use dependabot to bump style on our own terms. This way it'll open PRs to bump the version and we can evaluate any fixes that need to happen instead of breaking community PRs potentially due to unrelated rubocop changes.

lib/ohai/loader.rb
```diff
@@ -98,7 +99,7 @@ def load_plugin(plugin_path)
     # @return [Object] class object for the ohai plugin defined in the file
     def load_plugin_class(plugin_path)
       # Read the contents of the plugin to understand if it's a V6 or V7 plugin.
-      contents = ""
+      contents = String.new
       begin
         logger.trace("Loading plugin at #{plugin_path}")
         contents << IO.read(plugin_path)
```
phiggins, line 105:
```suggestion
      contents = begin
        logger.trace("Loading plugin at #{plugin_path}")
        IO.read(plugin_path)
```

I mean, if we're really about saving memory with this PR...

tas50: Wouldn't we want the trace outside the begin there?

lib/ohai/plugins/aix/virtualization.rb
```diff
@@ -91,7 +91,7 @@
                       }
               wpars[wpar_name][title][key] = value
             when "security settings"
-              privileges ||= ""
+              privileges = String.new
```
jaymzh, line 94: to be clear, these aren't the same, you used to preserve any setting of priveleges, now you're deleting it.

lib/ohai/plugins/aix/virtualization.rb
```diff
@@ -140,7 +140,8 @@
         ]
 
         top_level.each do |attribute|
-          evalstr = "wpars['#{wpar_name}']"
+          evalstr = String.new
+          evalstr << "wpars['#{wpar_name}']"
```
jaymzh, line 144: why is this better. than `evalstr = "wpars['#{wpar_name}']".freeze` ?

lib/ohai/plugins/ruby.rb
```diff
@@ -47,13 +46,13 @@ def run_ruby(command)
...
-    values.each_key do |v|
-      env_string << "#{v}=\#{#{values[v]}},"
+    env_string = []
+    values.each_pair do |k,v|
+      env_string << "#{k}=\#{#{v}}"
     end
 
     # Query the system ruby
-    result = run_ruby "puts %Q(#{env_string})"
+    result = run_ruby "puts %Q(#{env_string.join(',')})"
```
jaymzh, line 55: <3

lib/ohai/loader.rb
phiggins: Yeah that probably makes more sense 👍

lib/ohai/plugins/ruby.rb
```diff
@@ -46,13 +46,13 @@ def run_ruby(command)
     }
 
     # Create a query string from above hash
-    env_string = ""
-    values.each_key do |v|
-      env_string << "#{v}=\#{#{values[v]}},"
+    env_string = []
+    values.each_pair do |k,v|
+      env_string << "#{k}=\#{#{v}}"
     end
```
phiggins, line 52:
```suggestion
    env_string = values.map do |k,v|
      "#{k}=\#{#{v}}"
    end
```

spec/unit/plugins/passwd_spec.rb
```diff
@@ -55,7 +56,7 @@
   if "".respond_to?(:force_encoding)
```
phiggins, line 56: `.force_encoding` was added in ruby 2.0 (?) so this check is probably safe to remove.

lib/ohai/plugins/passwd.rb
```diff
@@ -9,7 +10,7 @@
   # @return [String]
   #
   def fix_encoding(str)
-    str.force_encoding(Encoding.default_external) if str.respond_to?(:force_encoding)
+    str.dup.force_encoding(Encoding.default_external) if str.respond_to?(:force_encoding)
```
phiggins, line 13:
```suggestion
    str.dup.force_encoding(Encoding.default_external)
```

`.force_encoding` was added a long time ago, probably safe to remove.

tas50:
I think I'm going to hit reset on this and only set frozen where we're not mutating for now

PR #1486, opened 2020-08-03
Faster ruby language plugin

lamont-granquist:
This drops two additional shell outs to run ruby.


tas50:
This is actually our slowest plugin too

tas50:
Before/after:

```
[2020-08-03T16:50:48-07:00] TRACE: Plugin Ruby took 1.1075100000016391 seconds to run.
[2020-08-03T16:51:50-07:00] TRACE: Plugin Ruby took 0.2576439999975264 seconds to run.
```

PR #1503, opened 2020-08-27
Ohai Target Mode aka Remote Ohai

lamont-granquist:
This implements remote ohai, inspired originally by http://github.com/burtlo/ohbye

1.  When targeting remote unix-like or windows systems it preserves the semantics of ohai run locally (apart from a few known bugs)
2.  It implements a "train" plugin with data in `node['train']` about the remote connection
3.  For non-unix/windows systems it disables most default plugins other than os, ohai_time + chef_version, and introduces a `:target` type which is the default that is used for those kinds of targets
4.  It supports URLs in the `--target` option on the command line and also uses chef-client's code for parsing credentials files
5.  With the chef-client PR this integrates seamlessly with chef-client
6.  Existing plugins have been converted to use the train helpers to enable remote collection (mostly)

Adds train data to ohai when running remotely:

```
  "train": {
    "family_hierarchy": [
      "debian",
      "linux",
      "unix",
      "os"
    ],
    "family": "debian",
    "platform": {
      "arch": "x86_64",
      "release": "18.04"
    },
    "backend": "unknown",
    "scheme": "ssh",
    "uri": "ssh://root@crash.scriptkiddie.org:747"
  },
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,10 @@ source "https://rubygems.org"
 
 gemspec
 
+# pull these gems from master of chef/chef so that we're testing against what we will release
+gem "chef-config", git: "https://github.com/chef/chef", glob: "chef-config/chef-config.gemspec"
+gem "chef-utils", git: "https://github.com/chef/chef", glob: "chef-utils/chef-utils.gemspec"
+
 # NOTE: do not submit PRs to add pry as a dep, add to your Gemfile.local
 group :development do
   gem "chefstyle", git: "https://github.com/chef/chefstyle.git", branch: "master"
--- a/lib/ohai/application.rb
+++ b/lib/ohai/application.rb
@@ -56,6 +56,15 @@ class Ohai::Application
     description: "Set the log file location, defaults to STDOUT - recommended for daemonizing",
     proc: nil
 
+  option :target,
+    short: "-t TARGET",
+    long: "--target TARGET",
+    description: "Target Ohai against a remote system or device",
+    proc: lambda { |target|
+      Ohai::Log.warn "-- EXPERIMENTAL -- Target mode activated -- EXPERIMENTAL --"
+      target
+    }
+
   option :help,
     short: "-h",
     long: "--help",
@@ -93,7 +102,37 @@ class Ohai::Application
 
     load_workstation_config
 
+    merge_configs
+
+    if config[:target]
+      Ohai::Config.target_mode.host = config[:target]
+      if URI.parse(Ohai::Config.target_mode.host).scheme
+        train_config = Train.unpack_target_from_uri(Ohai::Config.target_mode.host)
+        Ohai::Config.target_mode = train_config
+      end
+      Ohai::Config.target_mode.enabled = true
+      Ohai::Config.node_name = Ohai::Config.target_mode.host unless Ohai::Config.node_name
+    end
+
     Ohai::Log.init(Ohai.config[:log_location])
+  end
+
+  # @api private
+  def config_file_defaults
+    Ohai::Config.save(true)
+  end
+
+  # @api private
+  def config_file_settings
+    Ohai::Config.save(false)
+  end
+
+  # See lib/chef/knife.rb in the chef/chef github repo
+  #
+  # @api private
+  def merge_configs
+    config.replace(config_file_defaults.merge(default_config).merge(config_file_settings).merge(config))
+    Ohai::Config.merge!(config) # make them both the same
   end
 
   # Passes config and attributes arguments to Ohai::System then prints the results.
--- a/lib/ohai/mixin/chef_utils_wiring.rb
+++ b/lib/ohai/mixin/chef_utils_wiring.rb
@@ -36,7 +36,7 @@ module Ohai
         transport_connection
       end
 
-      # beause of target mode we cache the PATH to avoid massive amounts of `echo $PATH` remote queries
+      # because of target mode we cache the PATH to avoid massive amounts of `echo $PATH` remote queries
       #
       def __env_path
         PathCache.instance.path_cache ||= super
--- a/lib/ohai/mixin/chef_utils_wiring.rb
+++ b/lib/ohai/mixin/chef_utils_wiring.rb
@@ -16,6 +16,7 @@
 #
 
 require_relative "../config"
+require "singleton" unless defined?(Singleton)
 
 module Ohai
   module Mixin
@@ -32,7 +33,18 @@ module Ohai
       end
 
       def __transport_connection
-        #        Chef.run_context&.transport_connection
+        transport_connection
+      end
+
+      # beause of target mode we cache the PATH to avoid massive amounts of `echo $PATH` remote queries
+      #
+      def __env_path
+        PathCache.instance.path_cache ||= super
+      end
+
+      class PathCache
+        include Singleton
+        attr_accessor :path_cache
       end
     end
   end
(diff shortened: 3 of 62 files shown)


Gemfile
```diff
@@ -2,6 +2,8 @@ source "https://rubygems.org"
 
 gemspec
 
+gem "train-core"
```
tas50, line 5: shouldn't this be a gemspec dep?

lib/ohai/application.rb
```diff
@@ -56,6 +56,12 @@ class Ohai::Application
     description: "Set the log file location, defaults to STDOUT - recommended for daemonizing",
     proc: nil
 
+  option :target,
+    short: "-t TARGET",
+    long: "--target TARGET",
+    description: "Target a remote location to invoke Ohai",
```
tas50, line 62:
```suggestion
    description: "A remote target to run Ohai against",
```

lib/ohai/mixin/os.rb
```diff
@@ -27,6 +27,65 @@ module OS
...
+          collect_os_local
+        end
+      end
+
+      # This should exactly preserve the semantics of collect_os_local below, which is authoritative
+      # for the API and must adhere to pre-existing ohai semantics and not follow inspec's notion of
+      # os/family/hierarchy.
+      #
+      # Right or wrong the ohai `os` variable has matched the ruby `host_os` definition for the past
+      # 10+ years, preceding inspec/train's definitions and this is the docuemnted correct API of
```
tas50, line 42:
```suggestion
      # 10+ years, preceding inspec/train's definitions and this is the documented correct API of
```

lib/ohai/mixin/os.rb
```diff
@@ -27,6 +27,65 @@ module OS
...
+          # The purpose of the next two lines is that anything which runs Unix is presumed to be able to run ruby, and
+          # if it was not caught above, we MUST translate whatever train uses as the 'os' into the proper ruby host_os
+          # string.  If it is not unix and not caught above we assume it is something like a REST API which cannot run
+          # ruby.  If these assumptions are incorrect then it is a bug, which should be submitted to fix it, and the
+          # values should not be relied upon until that bug is fixed.  The train os is NEVER considered authoritataive
+          # for any target which can run ruby.
+          #
+
+        when transport_connection.os.unix?
+          raise "Ohai target mode not supported on this host"
```
tas50, line 81: We probably want to raise with a better message here that gives users a next step to get in contact with us to tell us what they're trying to use.

lib/ohai/mixin/os.rb
```diff
@@ -27,6 +27,65 @@ module OS
...
+
+          #
+          # The purpose of the next two lines is that anything which runs Unix is presumed to be able to run ruby, and
+          # if it was not caught above, we MUST translate whatever train uses as the 'os' into the proper ruby host_os
+          # string.  If it is not unix and not caught above we assume it is something like a REST API which cannot run
+          # ruby.  If these assumptions are incorrect then it is a bug, which should be submitted to fix it, and the
+          # values should not be relied upon until that bug is fixed.  The train os is NEVER considered authoritataive
+          # for any target which can run ruby.
+          #
+
```
tas50, line 79:
```suggestion
```

lib/ohai/mixin/train_helpers.rb
```diff
@@ -0,0 +1,45 @@
...
+  module Mixin
+    module TrainHelpers
+      include ChefUtils::DSL::TrainHelpers
+      include ChefUtilsWiring
+
+      def file_read(path)
+        file_open(path).readlines.join("\n")
+      end
+
+      def file_directory?(path)
```
tas50, line 30: wouldn't this be more clear if it was just directory?

Gemfile
lamont-granquist: yeah that's all a hack right now

lib/ohai/mixin/train_helpers.rb
lamont-granquist: matches the ruby api -- File.open -> file_open, IO.read/File.read -> file_read, File.directory? -> file_directory?

lib/ohai/mixin/train_helpers.rb
```diff
@@ -0,0 +1,45 @@
...
+require_relative "chef_utils_wiring" unless defined?(Ohai::Mixin::ChefUtilsWiring)
+
+module Ohai
+  module Mixin
+    module TrainHelpers
+      include ChefUtils::DSL::TrainHelpers
+      include ChefUtilsWiring
+
+      def file_read(path)
+        file_open(path).readlines.join("\n")
```
phiggins, line 27:
```suggestion
        file_open(path).read
```

Should be equivalent, unless we need strictly one newline between lines.

lamont-granquist: yeah that one was bugging me a lot, to get it going i just copypasta'd what was in the other PR.

so the one complexity of this is that it could be either an IO from the filesystem or a StringIO coming roundabout from train, but ti seems like IO#read and StringIO#read are the same, and we definitely did NOT want that behavior between newlines, so i think that's also a bugfix.

lib/ohai/application.rb
lamont-granquist: that code is all "TBD" placeholder to just bootstrap it, i'm pretty sure it changes and i'm probably copying chef-client's api here and then adding the URI feature to the chef-client code.

lib/ohai/application.rb
```diff
@@ -99,6 +99,14 @@ def configure_ohai
 
     load_workstation_config
 
+    Ohai::Config.merge!(config)
+
+    if config[:target] || Ohai::Config.target
```
jaymzh, line 107: shouldn't the merge on line 102 mean that you only have to check the latter one?

lib/ohai/application.rb
```diff
@@ -93,6 +102,14 @@ def configure_ohai
 
     load_workstation_config
 
+    Ohai::Config.merge!(config)
```
lamont-granquist, line 105: reminder that i should do this better

lib/ohai/application.rb
lamont-granquist: heh that's all WIP copypasta you shouldn't be reading that too closely, this is not ready for review

lib/ohai/mixin/os.rb
```diff
@@ -27,6 +27,71 @@ module OS
...
+          "solaris2"
+        when transport_connection.os.windows?
+          "windows"
+
+          #
+          # The purpose of the next two lines is that anything which runs Unix is presumed to be able to run ruby, and
+          # if it was not caught above, we MUST translate whatever train uses as the 'os' into the proper ruby host_os
+          # string.  If it is not unix and not caught above we assume it is something like a REST API which cannot run
+          # ruby.  If these assumptions are incorrect then it is a bug, which should be submitted to fix it, and the
+          # values should not be relied upon until that bug is fixed.  The train os is NEVER considered authoritataive
```
phiggins, line 76:
```suggestion
          # values should not be relied upon until that bug is fixed.  The train os is NEVER considered authoritative
```

lib/ohai/mixin/chef_utils_wiring.rb
```diff
@@ -32,7 +33,18 @@ def __log
       end
 
       def __transport_connection
-        #        Chef.run_context&.transport_connection
+        transport_connection
+      end
+
+      # beause of target mode we cache the PATH to avoid massive amounts of `echo $PATH` remote queries
```
phiggins, line 39:
```suggestion
      # because of target mode we cache the PATH to avoid massive amounts of `echo $PATH` remote queries
```

phiggins:
I'm impressed at the number of different ways in this repo to open a file and read each line.

PR #1507, opened 2020-09-08
Resolve new chefstyle warnings & use safe navigators

tas50:
We can avoid some duplicate checks by using &. in places

### DIFF
--- a/lib/ohai/plugins/linux/network.rb
+++ b/lib/ohai/plugins/linux/network.rb
@@ -141,11 +141,9 @@ Ohai.plugin(:Network) do
   # using a temporary var to hold routes and their interface name
   def parse_routes(family, iface)
     iface.collect do |i, iv|
-      if iv[:routes]
-        iv[:routes].collect do |r|
-          r.merge(dev: i) if r[:family] == family[:name]
-        end.compact
-      end
+      iv[:routes]&.collect do |r|
+        r.merge(dev: i) if r[:family] == family[:name]
+      end&.compact
     end.compact.flatten
   end
 
--- a/lib/ohai/plugins/linux/network.rb
+++ b/lib/ohai/plugins/linux/network.rb
@@ -141,10 +141,12 @@ Ohai.plugin(:Network) do
   # using a temporary var to hold routes and their interface name
   def parse_routes(family, iface)
     iface.collect do |i, iv|
-      iv[:routes]&.collect do |r|
+      next unless iv[:routes]
+
+      iv[:routes].collect do |r|
         r.merge(dev: i) if r[:family] == family[:name]
-      end&.compact
-    end.compact.flatten
+      end.compact # @todo: when we drop ruby 2.6 this should be a filter_map
+    end.compact.flatten # @todo: when we drop ruby 2.6 this should be a filter_map
   end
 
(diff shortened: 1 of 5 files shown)


lib/ohai/plugins/linux/network.rb
```diff
@@ -141,11 +141,9 @@ def check_routing_table(family, iface, default_route_table)
   # using a temporary var to hold routes and their interface name
   def parse_routes(family, iface)
     iface.collect do |i, iv|
-      if iv[:routes]
-        iv[:routes].collect do |r|
+      iv[:routes]&.collect do |r|
           r.merge(dev: i) if r[:family] == family[:name]
-        end.compact
-      end
+        end&.compact
```
phiggins, line 146: I think it'd be more readable to add a `next unless iv[:routes]` or similar at the top of the loop.

lib/ohai/plugins/solaris2/network.rb
```diff
@@ -86,7 +86,7 @@ def arpname_to_ifname(iface, arpname)
 
   def full_interface_name(iface, part_name, index)
     iface.each do |name, attrs|
-      next unless attrs && attrs.respond_to?(:[])
+      next unless attrs&.respond_to?(:[])
```
phiggins, line 89: Even `nil` responds to `respond_to?`.

lib/ohai/plugins/solaris2/network.rb
```diff
@@ -155,14 +155,12 @@ def full_interface_name(iface, part_name, index)
...
           end
         end
-        if iface[ifn][:arp]
-          iface[ifn][:arp].each_key do |addr|
+        iface[ifn][:arp]&.each_key do |addr|
             if addr.eql?(iaddr)
               iface[ifn][:addresses][iface[ifn][:arp][iaddr]] = { "family" => "lladdr" }
               break
             end
           end
```
phiggins, line 163: I think this indentation is off now?

phiggins (approved):
I'm not a huge fan of the pretzel operator but I could just be being a curmudgeon 😅

lib/ohai/plugins/solaris2/network.rb
phiggins: Chefstyle is unhappy about this.

tas50: Yeah that's me using Rubocop directly. I pushed it up again after the Chefstyle treatment.-

lib/ohai/plugins/linux/network.rb
lamont-granquist: yeah that's probably a good case where the alternative is most likely way more readable.

lamont-granquist:
(I kinda like the pretzel-dot operator, but some of these use cases are a bit funny)

lib/ohai/plugins/solaris2/network.rb
tas50: Oh yep that should be a cop right there

tas50: https://github.com/rubocop-hq/rubocop/issues/8668

tas50:
Nuked the weird ones. We should be good now

PR #1510, opened 2020-09-10
Prefer __dir__ to __FILE__ used to get the dir

tas50:
This avoids some File.dirname calls

### DIFF
--- a/spec/unit/plugins/linux/systemd_paths_spec.rb
+++ b/spec/unit/plugins/linux/systemd_paths_spec.rb
@@ -16,7 +16,7 @@
 # limitations under the License.
 #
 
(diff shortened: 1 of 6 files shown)


mwrock (approved):

spec/unit/plugins/linux/systemd_paths_spec.rb
```diff
@@ -16,7 +16,7 @@
 # limitations under the License.
 #
 
-require File.expand_path(File.dirname(__FILE__) + "/../../../spec_helper.rb")
+require File.expand_path(__dir__ + "/../../../spec_helper.rb")
```
phiggins, line 19:
```suggestion
require "spec_helper.rb"
```

`spec/` should be on the path.

PR #1514, opened 2020-09-17
chefstyle fixes

lamont-granquist:
result of automation.


lib/ohai/plugins/rackspace.rb
```diff
@@ -161,8 +161,8 @@ def get_private_networks
       unless rackspace[:public_ip].nil?
         rackspace[:public_hostname] = begin
                                         Resolv.getname(rackspace[:public_ip])
-                                      rescue Resolv::ResolvError, Resolv::ResolvTimeout
-                                        rackspace[:public_ip]
+        rescue Resolv::ResolvError, Resolv::ResolvTimeout
```
tas50, line 164: hmm I thought I fixed this in the latest with the monkeypatch

PR #1516, opened 2020-09-19
Windows support for Passwd plugin

jaymzh:
This adds Windows support for the passwd plugin.

It's not fast, but I did work to improve the performance dropping it
from 20 seconds to 8 seconds on a VM with only a few users and groups.
The `get-localgroupmember` is the slow part (you have to run it one
group at a time), so the more groups you have, the slower this will go.

Note that unlike the Linux/Mac variety which will pick up nonlocal users
and groups this one does not (yet). It isn't hard to add, there are the
`get-aduser` and `get-adgroup` commands (if you have joined a domain),
but this is a good start.

And yes, I'll add tests when I have a moment.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,3 +1,12 @@
+# Unreleased
+
+## etc Ohai Data on Windows
+
+Ohai's 'passwd' plugin that provides `node['etc']['passwd']` and `node['etc']['group']` now populates data on Windows. Data for all local users and groups is present. A few things to note:
+
+  * If you are on a domain controller, you will get all domain users as domain controllers see domain users as local
+  * If you are not on a domain controller you will only get actual local users
+  * Group members are not recursed, so you if groups are nested, you will simply see the group that is directly a member of the group.
 
 # Ohai Release Notes 15.6
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,7 +4,7 @@
 
 Ohai's 'Passwd' plugin that provides `node['etc']['passwd']` and `node['etc']['group']` now populates data on Windows. Data for all local users and groups is present. A few things to note:
 
-  * If you are on a domain controller, you will get all domain users as domain controllers see domain users as local
+  * If the node is an Active Directory domain controller, you will get all domain users as domain controllers see domain users as local
   * If you are not on a domain controller you will only get actual local users
   * Group members are not recursed, so you if groups are nested, you will simply see the group that is directly a member of the group.
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -5,7 +5,7 @@
 Ohai's 'Passwd' plugin that provides `node['etc']['passwd']` and `node['etc']['group']` now populates data on Windows. Data for all local users and groups is present. A few things to note:
 
   * If the node is an Active Directory domain controller, you will get all domain users as domain controllers see domain users as local
-  * If you are not on a domain controller you will only get actual local users
+  * If the node is not an Active Directory domain controller you will only get actual local users
   * Group members are not recursed, so you if groups are nested, you will simply see the group that is directly a member of the group.
 
 # Ohai Release Notes 15.6
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,7 +2,7 @@
 
 ## etc Ohai Data on Windows
 
-Ohai's 'passwd' plugin that provides `node['etc']['passwd']` and `node['etc']['group']` now populates data on Windows. Data for all local users and groups is present. A few things to note:
+Ohai's 'Passwd' plugin that provides `node['etc']['passwd']` and `node['etc']['group']` now populates data on Windows. Data for all local users and groups is present. A few things to note:
 
   * If you are on a domain controller, you will get all domain users as domain controllers see domain users as local
   * If you are not on a domain controller you will only get actual local users
--- a/lib/ohai/plugins/passwd.rb
+++ b/lib/ohai/plugins/passwd.rb
@@ -12,6 +12,10 @@ Ohai.plugin(:Passwd) do
     str
   end
 
+  def powershell_out(ps)
+    Mixlib::ShellOut.new("powershell.exe", "-c", ps).run_command
+  end
+
   collect_data do
     require "etc" unless defined?(Etc)
 
@@ -42,6 +46,44 @@ Ohai.plugin(:Passwd) do
   end
 
   collect_data(:windows) do
-    # Etc returns nil on Windows
+    unless etc
+      etc Mash.new
+
+      etc[:passwd] = Mash.new
+      s = powershell_out("get-localuser | convertto-json")
+      users = JSON.parse(s.stdout)
+      users.each do |user|
+        uname = user["Name"].strip.downcase
+        Ohai::Log.debug("processing user #{uname}")
+        etc[:passwd][uname] = Mash.new
+        user.each do |key, val|
+          etc[:passwd][uname][key.downcase] = val
+        end
+      end
+
+      etc[:group] = Mash.new
+      s = powershell_out("get-localgroup | convertto-json")
+      groups = JSON.parse(s.stdout)
+      groups.each do |group|
+        gname = group["Name"].strip.downcase
+        Ohai::Log.debug("processing group #{gname}")
+        etc[:group][gname] = Mash.new
+        group.each do |key, val|
+          etc[:group][gname][key.downcase] = val
+        end
+        # calling this for each group is slow, but it requires
+        # a specific group, soooooo....
+        g = powershell_out(
+          "get-localgroupmember -name '#{gname}' | convertto-json"
+        )
+        out = g.stdout
+        if !out.empty?
+          gmem = JSON.parse(g.stdout)
+          etc[:group][gname]["members"] = gmem
+        else
+          etc[:group][gname]["members"] = []
+        end
+      end
+    end
   end
 end
--- a/lib/ohai/plugins/passwd.rb
+++ b/lib/ohai/plugins/passwd.rb
@@ -12,10 +12,6 @@ Ohai.plugin(:Passwd) do
     str
   end
 
-  def powershell_out(ps)
-    Mixlib::ShellOut.new("powershell.exe", "-c", ps).run_command
-  end
-
   collect_data do
     require "etc" unless defined?(Etc)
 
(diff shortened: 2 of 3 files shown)


tas50:
See how long it takes to do this with WMI. The class is `Win32_UserAccount` and it looks like you can apply the`"LocalAccount=True` filter. Without that you're going to end up with 1000 active directory users on these and you'll blow up chef.

jaymzh:
On my linux box that is using sssd, I also get all the AD users. Which is necessary. So we would want remote users here just like we do in linux/mac. 
WMI is a black box to me... but i"ll try to find some time to poke at it... at the moment I did this at 10pm on a Friday to unblock my Monday-at-work self, so no promises.

jaymzh:
Thanks @HeyItsGilbert ! I'll try to poke at that soon.

HeyItsGilbert:
Actually, I think you need to start with Win32_Group. I'll poke around on my end.

jaymzh:
So a quick poking at this - it seems to work identically to the powershell cmdlets. You can request Win32_UserAccount and Win32_Group both of which will list them all, but neither of which list members. This is the same (speed and data wise) as Get-LocalUser and Get-LocalGroup cmdlets. However, getting the members requires querying one-group-at-a-time AFAICT, whihc makes it no faster than `Get-LocalGroupMember`.

jaymzh:
@tas50 thoughts?

jaymzh:
OK after a fair amount of research... this is about as good as it gets.

* WMI still requires a query-per-group for members and is no faster
* Even if I wanted to provide an option for directory users, I can't, `get-aduser` only works on the DC, unless you jump through crazy hoops for credential delegation.

So this is as good as it gets. Everyone happy?

jaymzh:
Added some tests. But I can't test the test easy since I'm developing on linux, so pushed them and we'll see. :)

jaymzh:
OK so I got it working, but based on the `powershell_out` work, I realized that my timing statements were incorrect. If I"m in powershell, the two commands are similar time, but forking powershell to do this turns out to take quite a bit of time. So I started looking into using wmi.... but the wmilite stuff doesn't seem to work as I understand it.

```
      users = wmi.query("SELECT * FROM Win32_UserAccount Where LocalAccount = True") 
      users.each do |user|
```

crashes with:

```
[2020-10-01T21:14:41-07:00] TRACE: Plugin Passwd threw #<NoMethodError: undefined method `each' for #<WmiLite::Wmi::Instance:0x0000000004c74f08>>
```

jaymzh:
OK @tas50 once you take into account the powershell startup time, powershell is indeed the slow part. I have moved this all to WMI, and it drops from 8 seconds to 0.8 seconds.

*HOWEVER*!!! There be dragons here!!

In order to use WMI I had to:

* Hardcode a ton of properties (this could be fixed by making chef/wmi-lite expose properties)
* Do crazy string parsing that is likely error-prone
* *Eval input* which is definitely *bad* and even *more* bad with @lamont-granquist 's TargetMode ohai coming along

So, it's faster, but it's also a lot worse. I kept it as a separate commit so I can go back. Or not. Your call.

The tests will all fail now since they're still written for the powershell version... I'll didn't update those yet.

Have a look and tell me what you prefer and what suggestions you have to make the WMI version less terrifying. I'll finish it up and hopefully we can get this merged and done with.

lib/ohai/plugins/passwd.rb
```diff
@@ -42,6 +42,95 @@ def fix_encoding(str)
...
+
+      wmi = WmiLite::Wmi.new
+
+      etc[:passwd] = Mash.new
+      users = wmi.query("SELECT * FROM Win32_UserAccount Where LocalAccount = True")
+      users.each do |user|
+        uname = user["Name"].strip.downcase
+        Ohai::Log.debug("processing user #{uname}")
+        etc[:passwd][uname] = Mash.new
+        # apparently, chef/wmi-lite is too good to give us
```
jaymzh, line 58: No offense, @adamedx  ;)

adamedx: :) That code was paged out of my memory -- if `wmi-lite` is giving you one of these [`wmi_instance`](https://github.com/chef/wmi-lite/blob/master/lib/wmi-lite/wmi_instance.rb) objects, maybe taking a look at that code will reveal a way to get at the properties. In theory the equivalent of an explicit `select` is more efficient and deterministic in the face of schema changes.
Also looking at `wmi-lite`, its built on top of `WIN32OLE`, so again there may be an implementation clue that might make it easier to use (or maybe hint at an improvement to `wmi-lite`).

At any rate, yeah, I think switching from shelling out to running in process seems like a huge perf gain, so I think the pain here is worth the tradeoff. Another option that lets you keep the PowerShell is to use a PowerShell session -- see [`New-PSSession`](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/new-pssession?view=powershell-7#example-1--create-a-session-on-the-local-computer). This would shell out once, but let you keep issuing commands to the same PowerShell -- you wouldn't pay repeated process startup cost, just the price of initial start and then marshalling input and output across processes.

There are probably other fun tricks such as finding a way to use .net, or some native stuff to call win32 apis, e.g. this fun from `chef`: https://github.com/chef/chef/blob/master/lib/chef/util/windows/net_user.rb#L34-L63. I guess you'd have to port it though since we're in `ohai` here...

phiggins: I am not familiar enough with `wmi-lite` to know if this is similar enough, but there's another plugin that iterates across the properties here: https://github.com/chef/ohai/blob/master/lib/ohai/plugins/windows/dmi.rb

jaymzh: Thanks @phiggins ! I had tried similar things, but didn't make it work, but that's a great example. I will try again.

jaymzh:
OK @phiggins is my hero for today. Was able to pull the properties from the OLE object.

@lamont-granquist and I chatted and he doesn't see a better option for eval, but maybe @phiggins or @adamedx have one?

Other than that, waiting on @tas50 to say he's OK with this approach and if so I'll massage the tests to match

lib/ohai/plugins/passwd.rb
```diff
@@ -42,6 +42,72 @@ def fix_encoding(str)
...
+          end
+          # WARNING! HERE BE DRAGONS!
+          # info['name'] will be a quoted string becuase we parsed a string
+          # of foo="bar",baz="crazy\"\ string"
+          # eval here is sketchy, but I can't think of a better way to get a
+          # raw string out of this. But if someone gets the right shit into
+          # a group name, you could make Chef do bad things. But then again,
+          # if you can do that, you're already the admin. But then again
+          # with TargetMode, this is probably extra terrifying. Someone
+          # give me a better way.
```
phiggins, line 103: I think `String.undump` could be useful here but it's kind of finicky:

```
irb(main):026:0> "\"hello\"".undump
=> "hello"
irb(main):027:0> "foo".undump
Traceback (most recent call last):
        5: from /Users/pete/.asdf/installs/ruby/2.6.6/bin/irb:23:in `<main>'
        4: from /Users/pete/.asdf/installs/ruby/2.6.6/bin/irb:23:in `load'
        3: from /Users/pete/.asdf/installs/ruby/2.6.6/lib/ruby/gems/2.6.0/gems/irb-1.0.0/exe/irb:11:in `<top (required)>'
        2: from (irb):27
        1: from (irb):27:in `undump'
RuntimeError (invalid dumped string; not wrapped with '"' nor '"...".force_encoding("...")' form)
irb(main):028:0> "\"crazy\" string\"".undump
Traceback (most recent call last):
        6: from /Users/pete/.asdf/installs/ruby/2.6.6/bin/irb:23:in `<main>'
        5: from /Users/pete/.asdf/installs/ruby/2.6.6/bin/irb:23:in `load'
        4: from /Users/pete/.asdf/installs/ruby/2.6.6/lib/ruby/gems/2.6.0/gems/irb-1.0.0/exe/irb:11:in `<top (required)>'
        3: from (irb):28
        2: from (irb):28:in `rescue in irb_binding'
        1: from (irb):28:in `undump'
RuntimeError (invalid dumped string; not wrapped with '"' nor '"...".force_encoding("...")' form)
```

Do you have examples of what's actually returned from this?

phiggins: This almost seems like a CSV file with `=` as the field separator and `,` as the line delimiter...

jaymzh:
```
"partcomponent"=>"\\\\MY-HOSTNAME\\root\\cimv2:Win32_SystemAccount.Domain=\"MY-HOSTNAME\",Name=\"Authenticated Users\""
```
is an example. I replaced the hostname with MY-HOSTNAME (which is always in all-caps here, because Windows, I guess)

jaymzh: actually I'm only undumping the *value* here not the key. the value seems less encoded, just surrounding quotes. So undump probably works.

```
irb(main):008:0> "\"Authenticated Users\"".undump
=> "Authenticated Users"
```

will update and test

phiggins: It seems to work for your example:

```
irb(main):042:0> example = "\\\\MY-HOSTNAME\\root\\cimv2:Win32_SystemAccount.Domain=\"MY-HOSTNAME\",Name=\"Authenticated Users\""
=> "\\\\MY-HOSTNAME\\root\\cimv2:Win32_SystemAccount.Domain=\"MY-HOSTNAME\",Name=\"Authenticated Users\""
irb(main):043:0> example.split(",").map { |s| s.split("=") }.map {|a,b| [a, b.undump] }
=> [["\\\\MY-HOSTNAME\\root\\cimv2:Win32_SystemAccount.Domain", "MY-HOSTNAME"], ["Name", "Authenticated Users"]]
```

phiggins: ... and in case it gets more complicated:

```
irb(main):047:0> CSV.parse(example, row_sep: ',', col_sep: '=')
=> [["\\\\MY-HOSTNAME\\root\\cimv2:Win32_SystemAccount.Domain", "MY-HOSTNAME"], ["Name", "Authenticated Users"]]
```

jaymzh: @phiggins thanks! I feel much better having nuked that `eval`. :)

lib/ohai/plugins/passwd.rb
```diff
@@ -42,6 +42,62 @@ def fix_encoding(str)
...
+        # The query returns nothing unless you specify domain *and* name, it's
+        # a path, not a set of queries.
+        subq = "Win32_Group.Domain='#{group["Domain"]}',Name='#{group["Name"]}'"
+        members = wmi.query(
+          "select * from Win32_GroupUser Where GroupComponent=\"#{subq}\""
+        )
+        etc[:group][gname]["members"] = members.map do |member|
+          mi = {}
+          info = Hash[
+            member["partcomponent"].split(",").map do |x| x.split("=") }.map { |a, b| [a, b.undump] }
```
phiggins, line 88:
```suggestion
            member["partcomponent"].split(",").map { |x| x.split("=") }.map { |a, b| [a, b.undump] }
```

spec/unit/plugins/passwd_spec.rb
```diff
@@ -16,50 +16,235 @@
...
+          .and_return(mock_shell_out(0, info, ""))
+      end
       plugin.run
-      root = plugin[:etc][:passwd]["root"]
-      expect(root["gecos"].encoding).to eq(Encoding.default_external)
+      expect(plugin[:etc][:group]["groupone"]["members"]).to eq(group_one_info)
+      expect(plugin[:etc][:group]["grouptwo"]["members"]).to eq(group_two_info)
     end
   end
 end
```
phiggins, line 250: Given our confusion around the unescaping it'd probably be a good idea to add a test that that's happening properly.

jaymzh: Definitely. As I said, I hadn't updated any of the tests for this, so will do that later today.

lib/ohai/plugins/passwd.rb
jaymzh: oops, not sure how that snuck in there

spec/unit/plugins/passwd_spec.rb
phiggins: Oh oops, I didn't see that part.

lib/ohai/plugins/passwd.rb
lamont-granquist: TIL dump/undump

jaymzh: @lamont-granquist right? Me too!

jaymzh:
OK! @tas50 - tests exist and pass, release notes are there, and all works well!

It is worth noting that there are other tests failing chef::config is failing on master.

RELEASE_NOTES.md
```diff
@@ -1,3 +1,12 @@
+# Unreleased
+
+## etc Ohai Data on Windows
+
+Ohai's 'passwd' plugin that provides `node['etc']['passwd']` and `node['etc']['group']` now populates data on Windows. Data for all local users and groups is present. A few things to note:
```
tas50, line 5:
```suggestion
Ohai's 'Passwd' plugin that provides `node['etc']['passwd']` and `node['etc']['group']` now populates data on Windows. Data for all local users and groups is present. A few things to note:
```

RELEASE_NOTES.md
```diff
@@ -1,3 +1,12 @@
...
+
+  * If you are on a domain controller, you will get all domain users as domain controllers see domain users as local
```
tas50, line 7:
```suggestion
  * If the node is an Active Directory domain controller, you will get all domain users as domain controllers see domain users as local
```

RELEASE_NOTES.md
```diff
@@ -1,3 +1,12 @@
...
+  * If you are not on a domain controller you will only get actual local users
```
tas50, line 8:
```suggestion
  * If the node is not an Active Directory domain controller you will only get actual local users
```

tas50:
Very nice to have this all done with WMI now. Thanks for working through that @jaymzh

jaymzh:
Thanks to you, @tas50, as well as @phiggins, @adamedx, and @HeyItsGilbert - this windowsn00b would have been lost without help.
Hopefully now I can go back to pretending Windows doesn't exist. :)

Issue #1524, opened 2020-10-05
include AWS IAM role-name in ec2 plugin data


kcbraunschweig:
### Describe the Enhancement
Currently the ohai ec2 plugin reads from the AWS metadata service but strips out all IAM data to prevent leaking security credentials:
https://github.com/chef/ohai/blob/master/lib/ohai/plugins/ec2.rb#L123-L127

We should modify this slightly to still include the IAM role-name which is part of the metadata at:
iam/security-credentials/_role-name_
see: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-categories.html
Don't include the data returned in the document under _role-name_ as that includes actually includes credentials, but the name from the URL path should be fine.

### Describe the Need
The role assigned to the host provides its identity and is likely to be used to make decisions about how to configure it.

### Current Alternative
Currently this could be found in recipe code by either making http calls to the metadata service during the run and parsing the response, or including the ruby aws sdk as a library can calling STS GetCallerIdentity which includes role information.

### Can We Help You Implement This?
This is a simple modification to the scope of the filtering based on the data already present. Creating an issue in case someone gets to it before me.

PR #1525, opened 2020-10-08
Include IAM role in ec2 data (issue #1524)

kcbraunschweig:
## Description
Currently the ohai ec2 plugin reads from the AWS metadata service but strips out all IAM data to prevent leaking security credentials. Change modifies this filtering slightly to still include the IAM role-name which is part of the metadata at:
iam/security-credentials/role-name
see: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-categories.html
Don't include the data returned in the document under role-name as that includes actually includes credentials, but the name part isn't a security concern and is useful to people to know the IAM identity of the instance where chef is running.

In the solution i'm being fairly paranoid checking explicitly for the security-credentials key and that it has exactly 1 key below it. As far as I can tell 'security-credentials' will only exist at all if there's an associated role, in fact if there's no role 'iam' won't exist at all and there can only be 1 associated role at any one time ever. So in reality if the IAM key exists at all, then there should be exactly 1 role name for us to find always. But yay paranoia, happy to do it however people like.

## Related Issue
https://github.com/chef/ohai/issues/1524

## Types of changes

## Checklist:
- [X] I have updated the documentation accordingly.

### DIFF
--- a/lib/ohai/plugins/ec2.rb
+++ b/lib/ohai/plugins/ec2.rb
@@ -122,11 +122,17 @@ Ohai.plugin(:EC2) do
       fetch_metadata.each do |k, v|
         # fetch_metadata returns IAM security credentials, including the IAM user's
         # secret access key. We'd rather not have ohai send this information
-        # to the server.
-        # http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/AESDG-chapter-instancedata.html#instancedata-data-categories
-        next if k == "iam" && !hint?("iam")
-
-        ec2[k] = v
+        # to the server. If the instance is associated with an IAM role we grab
+        # only the "info" key and the IAM role name.
+        # https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-categories.html
+        if k == "iam" && !hint?("iam")
+          ec2[:iam] = v.select { |key, value| key == "info" }
+          if v["security-credentials"] && v["security-credentials"].keys.length == 1
+            ec2[:iam]["role_name"] = v["security-credentials"].keys[0]
+          end
+        else
+          ec2[k] = v
+        end
       end
       ec2[:userdata] = fetch_userdata
       ec2[:account_id] = fetch_dynamic_data["accountId"]
--- a/spec/unit/plugins/ec2_spec.rb
+++ b/spec/unit/plugins/ec2_spec.rb
@@ -240,13 +240,16 @@ describe Ohai::System, "plugin ec2" do
         allow(plugin).to receive(:hint?).with("iam").and_return(false)
       end
 
-      it "parses ec2 iam/ directory and NOT collect iam/security-credentials/" do
+      it "parses ec2 iam/ directory and collect info and role_name and NOT collect iam/security-credentials/" do
         expect(@http_client).to receive(:get)
           .with("/2012-01-12/meta-data/")
           .and_return(double("Net::HTTP Response", body: "iam/", code: "200"))
         expect(@http_client).to receive(:get)
           .with("/2012-01-12/meta-data/iam/")
-          .and_return(double("Net::HTTP Response", body: "security-credentials/", code: "200"))
+          .and_return(double("Net::HTTP Response", body: "info\nsecurity-credentials/", code: "200"))
+        expect(@http_client).to receive(:get)
+          .with("/2012-01-12/meta-data/iam/info")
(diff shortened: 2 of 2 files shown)


kcbraunschweig:
those test failures don't seem related to this change but lmk if there's something I need to do there.

lib/ohai/plugins/ec2.rb
```diff
@@ -122,10 +122,15 @@ def looks_like_ec2?
...
+        # to the server. If the instance is associated with an IAM role we still
+        # grab the name of the role only.
+        # https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-categories.html
+        if k == "iam" && !hint?("iam")
+          if v["security-credentials"] && v["security-credentials"].keys.length == 1
+            ec2[:role_name] = v["security-credentials"].keys[0]
+          end
+          next
+        end
         ec2[k] = v
```
phiggins, line 134:
```suggestion
        if k == "iam" && !hint?("iam")
          if v["security-credentials"] && v["security-credentials"].keys.length == 1
            ec2[:role_name] = v["security-credentials"].keys[0]
          end
        else
          ec2[k] = v
        end
```

The control flow could be simplified here.

phiggins:
This seems like a good addition, but it makes me wonder if there's more relevant information here or a more complete way to handle this. In other words I'm suspicious that this satisfies one use case and could lead to a creeping addition of more things.

phiggins:
> those test failures don't seem related to this change but lmk if there's something I need to do there.

You're correct, the CI failures are unrelated.

kcbraunschweig:
> This seems like a good addition, but it makes me wonder if there's more relevant information here or a more complete way to handle this. In other words I'm suspicious that this satisfies one use case and could lead to a creeping addition of more things. 

This is the full structure:
```
  "iam": {
    "info": {
      "Code": "Success",
      "LastUpdated": "2020-10-07T18:44:07Z",
      "InstanceProfileArn": "arn:aws:iam::123456789:instance-profile/foo",
      "InstanceProfileId": "AAAAAAAAAAAAAFCK"
    },
    "security-credentials": {
      "fbinstance": {
        "Code": "Success",
        "LastUpdated": "2020-10-07T18:43:37Z",
        "Type": "AWS-HMAC",
        "AccessKeyId": "VERYSECRET",
        "SecretAccessKey": "j+VERYSECRETSETUFF",
        "Token": "IQo/VERYSECRETSTUFFQ==",
        "Expiration": "2020-10-08T01:10:52Z"
      }
    }
  }
```
The only other bits that seem potentially interesting to me are `InstanceProfileArn` and `InstanceProfileId`. It could change if they added something in the future but I'm wary of trying to future proof because they might add something new that we want to exclude.

jaymzh:
It feels to me like keeping all of `info` is harmless and useful. I agree that blacklisting is bad, they could add "additional_creds" or some such, so we should whitelist... but thoughts on grabbing `info` ?

kcbraunschweig:
lmk how you like this. grabs info. moves role_name under an iam key as well. If anything additional is desired in the future, it can be added to the `select`.

phiggins (approved):
The code seems good to me 👍 

In order to merge this PR all commits will need a DCO signoff though.

lib/ohai/plugins/ec2.rb
```diff
@@ -122,11 +122,17 @@ def looks_like_ec2?
...
-        # to the server.
-        # http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/AESDG-chapter-instancedata.html#instancedata-data-categories
-        next if k == "iam" && !hint?("iam")
-
-        ec2[k] = v
+        # to the server. If the instance is associated with an IAM role we grab
+        # only the "info" key and the IAM role name.
+        # https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-categories.html
+        if k == "iam" && !hint?("iam")
+          ec2[:iam] = v.select { |key, value| key == "info" }
```
jaymzh, line 129: I guess you did this for futureness? Less efficient than `ec2[:iam]['info'] = v['info']`, but I'm fine with this.

jaymzh (approved):
As @phiggins says you need to fix your DCO. Just squash them all together and forcepush, or amend your last commit and forcepush, either way.

lib/ohai/plugins/ec2.rb
kcbraunschweig: yeah mostly for futureness and slightly because i thought it was more elegant than having the 'info' string in 2 places and also having to also do ec2[:iam] = {} first. If info isn't there for some reason select will still return an empty hash.

jaymzh:
Looks good, thanks. I'll wait for someone internal to merge, or ping me tomorrow to merge if they haven't.

jaymzh:
Hey @lamont-granquist @tas50 @phiggins ... forgot I'm an approver, not an owner, so I don't have access to merge. Can one of you click the button?

PR #1543, opened 2020-10-29
Add unit testing for omnios platform detection

tas50:
Make sure this actually works and act as a reference for what the output
actually looks like.


jaymzh (approved):
lgtm.

However, since you're not changing the mocks, couldn't you just do `@plugin.run` inside the `before`, and save yourself 2 executions of the plugin?

tas50:
Yep I blindly copied that one. Totally useless to do that in each block there.

jaymzh:
yay faster tests! :)

PR #1545, opened 2020-10-29
Properly detect OpenIndiana and its version numbers

tas50:
1) Our regexes for the most part here were more complex than they needed to be
2) OpenIndiana's is now called OpenIndiana Hipster so the old regex was not working
3) Even the old regex only detected the major release instead of the full 3 point version

I doubt anyone is using the non-Hipster release at this point, but we detect it properly now and we have specs in place.

NOTE: Feel free to leave regex improvements here. I'm sure this could be better. I left the Solaris/Oracle Solaris one repetitive on purpose to make it easy to read what we were going for there. Realistically there we could probably squash that down to just "Oracle Linux" since that gets mid-Solaris 10 onwards.

### DIFF
--- a/lib/ohai/plugins/solaris2/platform.rb
+++ b/lib/ohai/plugins/solaris2/platform.rb
@@ -38,17 +38,17 @@ Ohai.plugin(:Platform) do
     file_open("/etc/release") do |file|
       while ( line = file.gets )
         case line
-        when /^.*(SmartOS).*$/
+        when /.*SmartOS.*/
           platform "smartos"
-        when /^\s*(OmniOS).*r(\d+).*$/
+        when /^\s*OmniOS.*r(\d+).*$/
           platform "omnios"
-          platform_version $2
-        when /^\s*(OpenIndiana).*oi_(\d+).*$/
+          platform_version $1
+        when /^\s*OpenIndiana.*(Development oi_|Hipster )(\d\S*)/ # https://rubular.com/r/iMtOBwbnyqDz7u
           platform "openindiana"
           platform_version $2
-        when /^\s*(Oracle Solaris)/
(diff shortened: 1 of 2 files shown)


lib/ohai/plugins/solaris2/platform.rb
```diff
@@ -38,17 +38,15 @@
...
-          platform_version $2
-        when /^\s*(OpenIndiana).*oi_(\d+).*$/
+          platform_version $1
+        when /^\s*OpenIndiana.*(Development oi_|Hipster )(\d\S*)/ # https://rubular.com/r/iMtOBwbnyqDz7u
           platform "openindiana"
           platform_version $2
-        when /^\s*(Oracle Solaris)/
-          platform "solaris2"
-        when /^\s*(Solaris)\s.*$/
+        when /^\s*(Oracle Solaris|Solaris)/
```
jaymzh, line 49: Why not just `/Solaris/` at this point? Oracle Solaris. Sun Solaris. Solaris... anything that's not the above and has `Solaris` in it, probably should match Solaris.

That said, I see no specific issue with what you have, just seems like more generic catch all would be better.

But if you really prefer to be specific that's cool. Or you could also shorten it to `/^\s*(Oracle )?Solaris/`

jaymzh (approved):
This is a nice cleanup

lib/ohai/plugins/solaris2/platform.rb
tas50: It made me nervous to open it up to that, but I guess as the last match we were doing that anyways

tas50: Thinking about it more I still think we want to match on whitespace and then Oracle Solaris or Solaris so this is probably the right thing to do here.

PR #1548, opened 2020-10-30
Simplify regexes by removing extra character classes

tas50:
We're working harder here than we need to. Let's simplify these and make
them easier to figure out for the humans involved.

### DIFF
--- a/lib/ohai/plugins/aix/uptime.rb
+++ b/lib/ohai/plugins/aix/uptime.rb
@@ -40,9 +40,9 @@ Ohai.plugin(:Uptime) do
     when /^\d+-\d/
       (d, h, m, s) = so.split(/[-:]/)
     when /^\d+:\d+:\d/
-      (h, m, s) = so.split(/[:]/)
+      (h, m, s) = so.split(/:/)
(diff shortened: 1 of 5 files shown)


lib/ohai/plugins/aix/uptime.rb
```diff
@@ -40,9 +40,9 @@
     when /^\d+-\d/
       (d, h, m, s) = so.split(/[-:]/)
     when /^\d+:\d+:\d/
-      (h, m, s) = so.split(/[:]/)
+      (h, m, s) = so.split(/:/)
     else
-      (m, s) = so.split(/[:]/)
+      (m, s) = so.split(/:/)
```
phiggins, line 45: FWIW, these could be the string `":"` instead.

phiggins (approved):

lib/ohai/plugins/aix/uptime.rb
tas50: I saw that one and I'm going to see if there's a cop for it already and if not I'm going to get an issue created for that.

PR #1549, opened 2020-10-30
Properly detect NVME/XVD devices in ZFS zpools

tas50:
Add another match in the zpool status regex and then update the existing unit test to have some NVME/XVD devices in it.


jaymzh (approved):
this seems good to me but only because you don't look at partitions. It's worth noting that because nvme devices end in a number, partitions are `pX` (`nvme0n1p2` for exmaple), so if there are places that handle partitions in this code, we'd have to handle that.

tas50:
True. We should get that noted in a bigger issue of: this plugin needs a better way to determine devices in zpools

tas50:
Also in general you give the pool the whole disk w/o the partition, but we should still support partitions.

PR #1550, opened 2020-10-30
Split on strings intead of regex for 3x speedup

tas50:
This is just over 3x faster than using the regex.


phiggins (approved):

jaymzh (commented):
already merged, but also: <3

PR #1553, opened 2020-10-31
Minor memory optimizations

tas50:
Shave a bit more memory usage off Ohai

### DIFF
--- a/lib/ohai/plugins/elixir.rb
+++ b/lib/ohai/plugins/elixir.rb
@@ -15,7 +15,7 @@
 
 Ohai.plugin(:Elixir) do
   provides "languages/elixir"
-  depends "languages"
+  depends "languages".freeze
 
   collect_data do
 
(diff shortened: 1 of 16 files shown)


lib/ohai/plugins/elixir.rb
```diff
@@ -15,7 +15,7 @@
 
 Ohai.plugin(:Elixir) do
   provides "languages/elixir"
```
jaymzh, line 17: can probably freeze this one too, no?

lib/ohai/plugins/go.rb
```diff
@@ -15,7 +15,7 @@
 
 Ohai.plugin(:Go) do
   provides "languages/go"
```
jaymzh, line 17: freeze?

jaymzh (approved):
You freeze one `provides`, but not the rest... feels like they can all be frozen.

lib/ohai/plugins/elixir.rb
tas50: That wasn't showing up in the retained objects. The langs string was in there 10 times

PR #1560, opened 2020-11-06
Fix parsing of hostnamectl to support values with colons

tas50:
This is probably faster too since we don't have to lstrip each result.

### DIFF
--- a/lib/ohai/plugins/linux/hostnamectl.rb
+++ b/lib/ohai/plugins/linux/hostnamectl.rb
@@ -26,7 +26,7 @@ Ohai.plugin(:Hostnamectl) do
     hostnamectl_path = which("hostnamectl")
     if hostnamectl_path
       shell_out(hostnamectl_path).stdout.split("\n").each do |line|
-        key, val = line.split(": ")
+        key, val = line.split(": ", 2)
         hostnamectl[key.chomp.lstrip.tr(" ", "_").downcase] = val.chomp
       end
     end
--- a/lib/ohai/plugins/linux/hostnamectl.rb
+++ b/lib/ohai/plugins/linux/hostnamectl.rb
@@ -27,7 +27,7 @@ Ohai.plugin(:Hostnamectl) do
     if hostnamectl_path
       shell_out(hostnamectl_path).stdout.split("\n").each do |line|
         key, val = line.split(": ", 2)
-        hostnamectl[key.chomp.lstrip.tr(" ", "_").downcase] = val.chomp
+        hostnamectl[key.chomp.lstrip.tr(" ", "_").downcase] = val
       end
     end
   end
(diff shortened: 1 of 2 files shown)


lib/ohai/plugins/linux/hostnamectl.rb
```diff
@@ -26,8 +26,8 @@
     hostnamectl_path = which("hostnamectl")
     if hostnamectl_path
       shell_out(hostnamectl_path).stdout.split("\n").each do |line|
-        key, val = line.split(":")
-        hostnamectl[key.chomp.lstrip.tr(" ", "_").downcase] = val.chomp.lstrip
+        key, val = line.split(": ")
+        hostnamectl[key.chomp.lstrip.tr(" ", "_").downcase] = val.chomp
```
phiggins, line 30:
```suggestion
        hostnamectl[key.chomp.lstrip.tr(" ", "_").downcase] = val
```

phiggins (approved):

lib/ohai/plugins/linux/hostnamectl.rb
```diff
@@ -26,8 +26,8 @@
     hostnamectl_path = which("hostnamectl")
     if hostnamectl_path
       shell_out(hostnamectl_path).stdout.split("\n").each do |line|
-        key, val = line.split(":")
-        hostnamectl[key.chomp.lstrip.tr(" ", "_").downcase] = val.chomp.lstrip
+        key, val = line.split(": ")
```
ziggythehamster, line 29: This would fail if e.g., I have my location set to something like `AWS: us-west-1b`. This prevents that:

```suggestion
        key, val = line.split(": ", 2)
```

lib/ohai/plugins/linux/hostnamectl.rb
```diff
@@ -26,7 +26,7 @@
     hostnamectl_path = which("hostnamectl")
     if hostnamectl_path
       shell_out(hostnamectl_path).stdout.split("\n").each do |line|
-        key, val = line.split(": ")
+        key, val = line.split(": ", 2)
```
jaymzh, line 29: It's weird to chomp the split value... @phiggins suggestion was confusing, but what I *THINK* he meant was do:
```
line.chomp.split(": ", 2)
```
Then drop the chomp from line 30.

tas50:
@ziggythehamster Do you have an example of the output with the 2nd `: `. I'd like to get that in the spec so we don't refactor this away later.

lib/ohai/plugins/linux/hostnamectl.rb
phiggins: Yes, I made the suggestion to drop the call to chomp before the other suggestion was made. Sorry for the confusion.

PR #1564, opened 2020-11-09
network: populate pause frame config

kuba-moo:
Query via ethtool and populate information whether the node 
has pause frames enabled or not.

## Description
To be able to configure pause parameters its useful to have access
to current settings.

## Types of changes

## Checklist:

### DIFF
--- a/lib/ohai/plugins/linux/network.rb
+++ b/lib/ohai/plugins/linux/network.rb
@@ -273,6 +273,30 @@ Ohai.plugin(:Network) do
     iface
   end
 
+  # determine pause parameters for the interface using ethtool
+  def ethernet_pause_parameters(iface)
+    return iface unless ethtool_binary_path
+
+    iface.each_key do |tmp_int|
+      next unless iface[tmp_int][:encapsulation] == "Ethernet"
+
+      so = shell_out("#{ethtool_binary_path} -a #{tmp_int}")
+      logger.trace("Plugin Network: Parsing ethtool output: #{so.stdout}")
+      iface[tmp_int]["pause_params"] = {}
+      so.stdout.lines.each do |line|
+        next if line.start_with?("Pause parameters for")
+        next if line.strip.nil?
+
+        key, val = line.split(/:\s+/)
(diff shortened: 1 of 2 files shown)


phiggins:
Hi, thanks for your contribution!

* You'll need to complete the DCO signoff process in order for these changes to be accepted. If you're using a relatively recent version of git, you can do this with `git rebase origin/master --signoff` and then force pushing your branch.
* It looks like this plugin has tests for the other sets of data collected, so it'd be good to add some tests for this new code to guard against future regressions: https://github.com/chef/ohai/blob/3ed8ec8524c52ccda346b7de35a949c4a1162b0b/spec/unit/plugins/linux/network_spec.rb

lib/ohai/plugins/linux/network.rb
```diff
@@ -273,6 +273,30 @@ def ethernet_coalesce_parameters(iface)
...
 
+  # determine pause parameters for the interface using ethtool
+  def ethernet_pause_parameters(iface)
+    return iface unless ethtool_binary_path
+
+    iface.each_key do |tmp_int|
+      next unless iface[tmp_int][:encapsulation] == "Ethernet"
+
+      so = shell_out("#{ethtool_binary_path} -a #{tmp_int}")
+      Ohai::Log.debug("Plugin Network: Parsing ethtool output: #{so.stdout}")
```
phiggins, line 284:
```suggestion
      logger.trace("Plugin Network: Parsing ethtool output: #{so.stdout}")
```

This was the only mention of `Ohai::Log` in this file, so it should be changed to match the surrounding functions.

phiggins (approved):
One comment, but otherwise looks good to me. 👍

lib/ohai/plugins/linux/network.rb
kuba-moo: I amended my commit with the suggestion. Hope that's fine, I'm not really sure on the GH etiquette :)

tas50:
Thanks @kuba-moo

PR #1570, opened 2020-11-18
Refactor how we parse ifconfig in AIX

tas50:
There's a few different things going on here.

- more descriptive variable names. Variables like splat were not very useful and others like iface actually held more than 1 value
- replace a gsub and split with just a split on a regex from @phiggins
- avoid the whole weird splat thing and ranges and instead just split each interface into the name and all the data. This makes the variables more clear when we use them throughout
- avoid some positional math and just use shift to simply pop off the leading values to build the mash

### DIFF
--- a/lib/ohai/plugins/aix/network.rb
+++ b/lib/ohai/plugins/aix/network.rb
@@ -38,7 +38,7 @@ Ohai.plugin(:Network) do
     # => state up/down (ifconfig/lsattr)
     # => arp (arp -an)
 
-    iface = Mash.new
+    ifaces = Mash.new
 
     network Mash.new unless network
 
@@ -56,20 +56,18 @@ Ohai.plugin(:Network) do
       network[:default_interface] = default_line[5]
     end
 
-    # Splits the ifconfig output to 1 line per interface
-    if_so = shell_out("ifconfig -a").stdout
-    if_so.gsub(/\n(\w+\d+)/, '___\1').split("___").each do |intraface|
-      splat = intraface.split(":")
-      interface = splat[0]
-      line = splat[1..-1][0]
-      iface[interface] = Mash.new
-      iface[interface][:state] = (line.include?("<UP,") ? "up" : "down")
+    # Splits the ifconfig output into arrays of interface strings
+    shell_out("ifconfig -a").stdout.split(/\n(?=\w)/).each do |int_lines|
+      int_name, int_data = int_lines.split(":", 2)
 
-      intraface.each_line do |lin|
+      ifaces[int_name] = Mash.new
+      ifaces[int_name][:state] = (int_data.include?("<UP,") ? "up" : "down")
+
+      int_data.each_line do |lin|
         case lin
         when /flags=\S+<(\S+)>/
-          iface[interface][:flags] = $1.split(",")
-          iface[interface][:metric] = $1 if lin =~ /metric\s(\S+)/
+          ifaces[int_name][:flags] = $1.split(",")
+          ifaces[int_name][:metric] = $1 if lin =~ /metric\s(\S+)/
         else
           # We have key value pairs.
           if lin =~ %r{inet (\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})(/(\d{1,2}))?}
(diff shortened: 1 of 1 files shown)


lib/ohai/plugins/aix/network.rb
```diff
@@ -123,8 +121,8 @@
       so_n.stdout.each_line do |line|
         if line =~ /(\S+)\s+(\S+)\s+(\S+)\s+(\d+)\s+(\d+)\s+(\S+)/
           interface = $6
-          iface[interface][:routes] = [] unless iface[interface][:routes]
-          iface[interface][:routes] << Mash.new( destination: $1, family: family,
+          ifaces[interface][:routes] = [] unless ifaces[interface][:routes]
```
tas50, line 124:
```suggestion
          ifaces[interface][:routes] ||= []
```

lib/ohai/plugins/aix/network.rb
```diff
@@ -56,20 +56,19 @@
...
+    # Splits the ifconfig output into arrays of interface strings
+    shell_out("ifconfig -a").stdout.split(/\n(?=\w)/).each do |int_lines|
+      int_name, int_data = int_lines.split(":", 2)
 
-      intraface.each_line do |lin|
+      ifaces[int_name] = Mash.new
+      ifaces[int_name][:addresses] ||= Mash.new
+      ifaces[int_name][:state] = (int_data.include?("<UP,") ? "up" : "down")
+
+      int_data.each_line do |lin|
```
phiggins, line 67: While we're on the subject of bad variable names, could this be `line`?

lib/ohai/plugins/aix/network.rb
```diff
@@ -84,56 +83,53 @@
...
+            # add all key value data into the interface mash
+            # for example "tcp_sendspace 131072 tcp_recvspace 131072 rfc1323 1"
+            # has keys tcp_sendspace, tcp_recvspace, and rfc1323
             properties = lin.split
-            n = properties.length / 2 - 1
-            (0..n).each do |i|
-              iface[interface][properties[i * 2]] = properties[(i * 2 + 1)]
+            (properties.size / 2).times do
+              ifaces[int_name][properties.shift] = properties.shift
             end
```
phiggins, line 103:
```suggestion
            lin.split.each_slice(2) do |key, value|
              ifaces[int_name][key] = value
            end
```

phiggins: Or as a one-liner (didn't test, might not work)

```suggestion
            ifaces[int_name].merge!(lin.split.each_slice(2).to_a.to_h)
```

lib/ohai/plugins/aix/network.rb
tas50: Fixed

lib/ohai/plugins/aix/network.rb
tas50: I pulled that in with the line modification done above. The 3 liner is easy on the eyes so I took that one. Thanks for further fixing that up.

phiggins (approved):
Still lots of stuff to nitpick but it looks like you got the grossest ones. Much nicer! 🎉

lib/ohai/plugins/aix/network.rb
```diff
@@ -84,56 +83,52 @@
...
+              ifaces[int_name][key] = value
             end
           end
         end
       end
 
       # Query macaddress
-      e_so = shell_out("entstat -d #{interface} | grep \"Hardware Address\"")
-      iface[interface][:addresses] ||= Mash.new
+      e_so = shell_out("entstat -d #{int_name} | grep \"Hardware Address\"")
```
lamont-granquist, line 108: is the grep still necessary?

tas50: Nope I have another PR that removes that, but it requires a good amount of spec updates

lamont-granquist: was it ever necessary?

lamont-granquist: :+1:

tas50: Like many things in these plugins we double check the data. We grep to get 1 line and then for some reason we use a regex to make sure that's in fact the line we got.

PR #1575, opened 2020-11-18
Simplify aix platform / platform_family detection

tas50:
If we're in a collect block for aix then we're on aix.


phiggins (approved):
Makes sense

PR #1576, opened 2020-11-18
Gather WPAR state in the AIX virtualization plugin

tas50:
This is a pretty important bit of information on the wpar.

### DIFF
--- a/lib/ohai/plugins/aix/virtualization.rb
+++ b/lib/ohai/plugins/aix/virtualization.rb
@@ -32,8 +32,7 @@ Ohai.plugin(:Virtualization) do
       virtualization[:lpar_name] = lpar_name
     end
 
-    so = shell_out("uname -W")
-    wpar_no = so.stdout.split($/)[0]
+    wpar_no = shell_out("uname -W").stdout.chomp
     if wpar_no.to_i > 0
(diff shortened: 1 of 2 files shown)


lib/ohai/plugins/aix/virtualization.rb
```diff
@@ -32,17 +32,19 @@
...
+    wpar_no = shell_out("uname -W").stdout.chomp
     if wpar_no.to_i > 0
       virtualization[:wpar_no] = wpar_no
     else
       # the below parses the output of lswpar in the long format
       so = shell_out("lswpar -L").stdout.scan(/={65}.*?(?:EXPORTED\n\n)+/m)
       wpars = Mash.new
       so.each do |wpar|
-        wpar_name = wpar.lines[1].split[0]
+        wpar_heading = wpar.lines[1].split
```
phiggins, line 43: You could potentially simplify this a bit by using `.split(" - ")`.

phiggins (approved):

lib/ohai/plugins/aix/virtualization.rb
tas50: I'm a bit nervous about relying on particular space counts. I had it that way originally and then went back to this one.

PR #1582, opened 2020-11-23
Misc RuboCop cleanup of the codebase

tas50:
A few misc things

### DIFF
--- a/lib/ohai/dsl/plugin/versionvii.rb
+++ b/lib/ohai/dsl/plugin/versionvii.rb
@@ -195,8 +195,6 @@ module Ohai
           # key? to avoid falsely instantiating a configuration hash.
           if Ohai.config[:plugin].key?(snake_case_name)
             Ohai.config[:plugin][snake_case_name]
-          else
-            nil
           end
         end
       end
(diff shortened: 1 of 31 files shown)


lib/ohai/dsl/plugin/versionvii.rb
```diff
@@ -195,8 +195,6 @@ def fetch_plugin_config
           # key? to avoid falsely instantiating a configuration hash.
           if Ohai.config[:plugin].key?(snake_case_name)
             Ohai.config[:plugin][snake_case_name]
-          else
-            nil
```
phiggins, line 199: If the caller is doing`.nil?` or similar on the return value this change could break it. Something like this would preserve the return value and probably fix the warning:

```
return unless Ohai.config[:plugin].key?(snake_case_name)
Ohai.config[:plugin][snake_case_name]
```

phiggins (approved):

lib/ohai/dsl/plugin/versionvii.rb
tas50: After a slack convo it looks like we're good here since it returned nil before.

PR #1599, opened 2021-01-26
Fix fetching metadata on AWS compatible clouds

tas50:
They're not going to have a IMDSv2 token to fetch and they were failing hard.

### DIFF
--- a/lib/ohai/plugins/rackspace.rb
+++ b/lib/ohai/plugins/rackspace.rb
@@ -162,8 +162,8 @@ Ohai.plugin(:Rackspace) do
       unless rackspace[:public_ip].nil?
         rackspace[:public_hostname] = begin
           Resolv.getname(rackspace[:public_ip])
-        rescue Resolv::ResolvError, Resolv::ResolvTimeout
-          rackspace[:public_ip]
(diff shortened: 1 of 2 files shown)


lamont-granquist (approved):

lamont-granquist:
the whitespace change here is why i tend to like to start a multi-line `begin..end` on its own line to just get a single 2 char indent, which everyone seems to agree on and keeps the indentation level much lower...

lib/ohai/plugins/rackspace.rb
```diff
@@ -162,8 +162,8 @@ def get_private_networks
       unless rackspace[:public_ip].nil?
         rackspace[:public_hostname] = begin
           Resolv.getname(rackspace[:public_ip])
-        rescue Resolv::ResolvError, Resolv::ResolvTimeout
-          rackspace[:public_ip]
+                                      rescue Resolv::ResolvError, Resolv::ResolvTimeout
+                                        rackspace[:public_ip]
```
ramereth, line 166: Was this formatting change intentional?

tas50: Yeah rubocop fixed a bug so this is being caught now. They had a regression before that forced the previous formatting

PR #1616, opened 2021-02-04
Update `packages` for Windows to be complete.

jaymzh:
Per the [docs](https://docs.chef.io/resources/windows_package/), there
are three locations packages can be registered, but we only look in 2,
this adds all 3.

Now packages show up that were otherwise missing.


damon-oss (approved):

jaymzh:
@tas50  / @lamont-granquist - tests are passing, I think this is good to go.

PR #1617, opened 2021-02-05
Add support for Podman and fix docker guest detection on Github Actions

ramereth:
This provides proper support for detecting Podman hosts and guests which is now the default container system on Fedora and RHEL 8+.

In addition, this works around a problem when running containers on Github Actions. They apparently configure the docker on their runs in a non-standard way which renames the cgroup mounts to `actions_job`.

It also shows `/system.slice/containerd.service` as a cgroup mount point and should be a safe assumption this is running inside a docker container. This should be checked last in the logic order since it would be triggered in the first check if it were added there.

You can see an example of this not working here:

https://github.com/sous-chefs/resolver/pull/47/checks?check_run_id=1840420956

### DIFF
--- a/lib/ohai/plugins/linux/virtualization.rb
+++ b/lib/ohai/plugins/linux/virtualization.rb
@@ -35,6 +35,10 @@ Ohai.plugin(:Virtualization) do
     which("docker")
   end
 
+  def podman_exists?
+    which("podman")
+  end
+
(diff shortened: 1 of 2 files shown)


lamont-granquist (approved):

tas50:
Very cool. Thanks @ramereth

PR #1620, opened 2021-02-08
Add Alibaba cloud plugin

tas50:
Detects Alibaba on Linux boxes and fetches metadata

### DIFF
--- a/lib/ohai/mixin/alibaba_metadata.rb
+++ b/lib/ohai/mixin/alibaba_metadata.rb
@@ -24,10 +24,9 @@ module Ohai
     # This code parses the Alibaba Instance Metadata API to provide details
     # of the running instance.
     #
-    # As of 2021-02-07 there is only one API release so this removes the
-    # API parsing logic in the ec2 metadata mixin. If that changes we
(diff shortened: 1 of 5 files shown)


lib/ohai/mixin/alibaba_metadata.rb
```diff
@@ -0,0 +1,87 @@
...
+module Ohai
+  module Mixin
+    #
+    # This code parses the Alibaba Instance Metadata API to provide details
+    # of the running instance.
+    #
+    # As of 2021-02-07 there is only one API release so this removes the
+    # API parsing logic in the ec2 metadata mixin. If that changes we
+    # may want to merge these two codebases
+    #
```
jaymzh, line 30: why split them, just to make someone else (probably you) merge them later? It's a versioned API, so it'll very likely get an additional version... why make that painful for us then?

tas50: I need to actually remove that comment. It became clear pretty quickly that the approaches are very different and not something we could merge / make generic.

lamont-granquist (approved):

PR #1622, opened 2021-02-13
Adds Habitat plugin

collinmcneese:
Signed-off-by: Collin McNeese <cmcneese@chef.io>

## Description

Adds collection of installed Habitat version, packages and services:
__node['habitat']['version']__ : String
__node['habitat']['packages']__ : Array of package identities
__node['habitat']['services']__ : Array of service names

## Types of changes

## Checklist:

### DIFF
--- a/lib/ohai/plugins/habitat.rb
+++ b/lib/ohai/plugins/habitat.rb
@@ -19,7 +19,9 @@ Ohai.plugin(:Habitat) do
   provides "habitat"
 
   def fetch_habitat_version
-    shell_out("hab -V").stdout.gsub(/hab\s*/, "").strip || nil
+    shell_out("hab -V").stdout.gsub(/hab\s*/, "").strip
+  rescue Ohai::Exceptions::Exec
+    logger.trace("Plugin Habitat: No detected version of hab binary found in PATH, skipping collection.")
   end
 
   def fetch_habitat_packages
--- a/lib/ohai/plugins/habitat.rb
+++ b/lib/ohai/plugins/habitat.rb
@@ -19,29 +19,29 @@ Ohai.plugin(:Habitat) do
   provides "habitat"
 
   def fetch_habitat_version
-    shell_out("hab -V").stdout.gsub(/hab\s*/, "").strip() || nil
+    shell_out("hab -V").stdout.gsub(/hab\s*/, "").strip || nil
   end
 
   def fetch_habitat_packages
     if Dir.exist?("C:/hab/pkgs")
-      Dir.glob("C:/hab/pkgs/*/*/*/*/").sort.map{ |pkg| pkg.gsub("C\:\/hab\/pkgs\/", "").chomp("/") }
+      Dir.glob("C:/hab/pkgs/*/*/*/*/").sort.map { |pkg| pkg.gsub("C\:\/hab\/pkgs\/", "").chomp("/") }
     elsif Dir.exist?("/hab/pkgs")
-      Dir.glob("/hab/pkgs/*/*/*/*/").sort.map{ |pkg| pkg.gsub("\/hab\/pkgs\/", "").chomp("/") }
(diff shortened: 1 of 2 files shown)


tas50 (changes requested):

lib/ohai/plugins/habitat.rb
```diff
@@ -0,0 +1,49 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+#
+Ohai.plugin(:Habitat) do
+  provides "habitat"
+
+  def fetch_habitat_version
+    shell_out("hab -V").stdout.gsub(/hab\s*/, "").strip
```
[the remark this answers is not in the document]
collinmcneese: Do you mean to parse the output as an array and select the field I want rather than gsub'ing the string data?  I've never seen an example of `shell_out()` with an array passed for a single command with an argument so just wanting to clarify.  The rest of the plugins in this repo pass the binary and arguments as a string as well -- or are you wanting to change moving forward?

lib/ohai/plugins/habitat.rb
```diff
@@ -0,0 +1,49 @@
...
+  rescue Ohai::Exceptions::Exec
+    logger.trace("Plugin Habitat: No detected version of hab binary found in PATH, skipping collection.")
+  end
+
+  def fetch_habitat_packages
+    if Dir.exist?("C:/hab/pkgs")
+      Dir.glob("C:/hab/pkgs/*/*/*/*/").sort.map { |pkg| pkg.gsub("C\:\/hab\/pkgs\/", "").chomp("/") }
```
tas50, line 29: Do you have an example of this dir structure that would require us to go this deep down?

tas50: If you're just trying to get the children in the dirs w/o the full paths you can do Dir.children(FOO). That may allow you to make the hab dirs a per platform helper and condense a large amount of this code down to just a few children method calls.

collinmcneese: Yeah this structure matches what would be for:
/hab/pkgs/pkg_origin/pkg_name/pkg_version/pkg_release

Trying to avoid shelling out as much as possible and habitat packages will always have these 4 directory paths as part of their tree and is a reliable way to show the installed packages.  Want to capture all 4 of these paths as well because you really need to see everything from origin->release to show the version that is installed for a package.
This will feed back items in to an array that look like `origin/package/version/release`

collinmcneese: I didnt just `**/*` on it though because only want to capture these 4 directories, nothing below, that is why it is explicitly `/hab/pkgs/*/*/*/*`

collinmcneese: I could redo this just to do a `hab pkg list --all` and get the output with a rescue though too

lib/ohai/plugins/habitat.rb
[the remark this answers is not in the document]
tas50:
```suggestion
    shell_out(["hab", "-V"]).stdout.gsub(/hab\s*/, "").strip
```
The shellout helper wants arrays so we can avoid OS specific spacing terribleness.

lib/ohai/plugins/habitat.rb
tas50: Yeah ideally we run the command. That would be more reliable with target mode too where we can't just use these ruby file helpers.

collinmcneese: Cool -- will refactor this a bit, quite a bit actually so i'll close out this PR and create a new one with the updates

collinmcneese:
Closing PR to refactor based on suggestions from @tas50 and will re-submit after changes are made

PR #1623, opened 2021-02-15
adds Habitat plugin

collinmcneese:
Signed-off-by: Collin McNeese <cmcneese@chef.io>

## Description

Adds collection of installed Habitat version, packages and services:
__node['habitat']['version']__ : String
__node['habitat']['packages']__ : Array of package identities
__node['habitat']['services']__ : Array of service names

## Types of changes

## Checklist:

### DIFF
--- a/lib/ohai/plugins/habitat.rb
+++ b/lib/ohai/plugins/habitat.rb
@@ -19,7 +19,9 @@ Ohai.plugin(:Habitat) do
   provides "habitat"
 
   def fetch_habitat_version
-    shell_out("hab -V").stdout.gsub(/hab\s*/, "").strip || nil
+    shell_out("hab -V").stdout.gsub(/hab\s*/, "").strip
+  rescue Ohai::Exceptions::Exec
+    logger.trace("Plugin Habitat: No detected version of hab binary found in PATH, skipping collection.")
   end
 
   def fetch_habitat_packages
--- a/lib/ohai/plugins/habitat.rb
+++ b/lib/ohai/plugins/habitat.rb
@@ -18,16 +18,20 @@
 Ohai.plugin(:Habitat) do
   provides "habitat"
 
+  def habitat_exists?
+    which("habitat")
+  end
+
   def fetch_habitat_version
     shell_out(["hab"], ["-V"]).stdout.gsub(/hab\s*/, "").strip
   rescue Ohai::Exceptions::Exec
-    logger.trace("Plugin Habitat: No detected version of hab binary found in PATH, skipping collection.")
+    logger.trace("Plugin Habitat: Unable to determine the installed version of Habitat, skipping collection.")
   end
 
   def fetch_habitat_packages
     shell_out(["hab", "pkg", "list", "--all"]).stdout.split.sort.select { |pkg| pkg.match?(%r{.*/.*/.*/.*}) }
   rescue Ohai::Exceptions::Exec
-    logger.trace("Plugin Habitat: No detected version of hab binary found in PATH, skipping collection.")
+    logger.trace("Plugin Habitat: Unable to determine the installed Habitat packages, skipping collection.")
   end
 
   def load_habitat_service_via_cli(status_stdout)
@@ -54,10 +58,12 @@ Ohai.plugin(:Habitat) do
     services_shell_out = shell_out(%w{hab svc status}).stdout
     load_habitat_service_via_cli(services_shell_out) if services_shell_out
   rescue Ohai::Exceptions::Exec
-    logger.trace("Plugin Habitat: No detected version of hab binary found in PATH, skipping collection.")
+    logger.trace("Plugin Habitat: Unable to determine the installed Habitat services, skipping collection.")
   end
 
   collect_data(:default) do
+    return unless habitat_exists?
+
     habitat Mash.new
     habitat["version"] = fetch_habitat_version
     habitat["packages"] = fetch_habitat_packages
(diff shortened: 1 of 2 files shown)


lib/ohai/plugins/habitat.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  def fetch_habitat_services
+    services_shell_out = shell_out(%w{hab svc status}).stdout
+    load_habitat_service_via_cli(services_shell_out) if services_shell_out
+  rescue Ohai::Exceptions::Exec
+    logger.trace("Plugin Habitat: No detected version of hab binary found in PATH, skipping collection.")
+  end
+
+  collect_data(:default) do
+    habitat Mash.new
```
jaymzh, line 61: The design of this means you'll shell out 3 times to a non-existent binary if habitat isn't installed. That's a bummer.

Instead why not use `which()` helper, and `return` if you don't get anything as a result?

collinmcneese: 👍  cleared that up

jaymzh: Nice thanks. Though I'd recommend using the results of `which` (which gives you the path), in all your shellouts, rather than just shelling out with no path. Ohai codebase is inconsistent about this, so you don't have to, but that's my suggestion.

spec/unit/plugins/habitat_spec.rb
```diff
@@ -34,6 +34,7 @@
     allow(plugin).to receive(:shell_out).with(["hab", "pkg", "list",
                                                "--all"]).and_return(mock_shell_out(0, pkg_result, ""))
     allow(plugin).to receive(:shell_out).with(%w{hab svc status}).and_return(mock_shell_out(0, svc_result, ""))
+    allow(plugin).to receive(:habitat_exists?).and_return(true)
```
jaymzh, line 37: it returns a path to the found binary, or nil.

jaymzh (approved):
Thanks!

lib/ohai/plugins/habitat.rb
```diff
@@ -0,0 +1,72 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+#
+Ohai.plugin(:Habitat) do
+  provides "habitat"
+
```
jaymzh, line 20: Actually, one more request. Lets make this an optional plugin.
```suggestion
  provides "habitat"
  optional true

```

jaymzh (changes requested):

jaymzh (approved):

jaymzh:
Thanks, if no one's objected (or merged) by Monday's meeting, I'll do it then.

lib/ohai/plugins/habitat.rb
tas50: We'll actually want this one to be a default plugin. We want habitat data anywhere we have it

lib/ohai/plugins/habitat.rb
```diff
@@ -0,0 +1,72 @@
+# frozen_string_literal: true
+#
+# License:: Apache License, Version 2.0
```
tas50, line 3: You should toss a copyright up here as well

lib/ohai/plugins/habitat.rb
```diff
@@ -0,0 +1,72 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+#
+Ohai.plugin(:Habitat) do
+  provides "habitat"
+
+  def habitat_binary
+    which("hab")
```
tas50, line 22: This avoids having to find it in the path multiple times during this plugin run

```suggestion
    @hab ||= which("hab")
```

lib/ohai/plugins/habitat.rb
```diff
@@ -0,0 +1,72 @@
...
+    shell_out([habitat_binary, "pkg", "list", "--all"]).stdout.split.sort.select { |pkg| pkg.match?(%r{.*/.*/.*/.*}) }
+  rescue Ohai::Exceptions::Exec
+    logger.trace("Plugin Habitat: Unable to determine the installed Habitat packages, skipping collection.")
+  end
+
+  def load_habitat_service_via_cli(status_stdout)
+    # package                           type        desired  state  elapsed (s)  pid   group
+    # core/httpd/2.4.35/20190307151146  standalone  up       up     158169       1410  httpd.default
+    @services = []
+    lines = status_stdout.split("\n")
```
tas50, line 41: I think you can just do 

status_stdout.each_line

that should avoid the need to create the lines array in memory

lib/ohai/plugins/habitat.rb
```diff
@@ -0,0 +1,72 @@
...
+    lines.each do |line|
+      @service = {}
```
tas50, line 43: This can just be `service` instead of `@service`

jaymzh: how did I miss that?! Gah. Note to self, don't review code when doing a Czech-timezone conference.

tas50:
```
{
  "version": "1.6.181/20201030173740",
  "packages": [

  ],
  "services": [

  ]
}
```

PR #1632, opened 2021-03-09
Update the XCP-ng Linux Detection

tas50:
- All supported releases have an /etc/os-release file so nuke the legacy detection
- The platform ID in the os-release file is XCP-ng in 7 and xenserver in 8 (this is a bug IMO) so update the XCP match
- Downcase platform IDs. It seems like everyone makes these lowercase, but XCP-ng. There may be others and ohai platforms need to be lowercase

### DIFF
--- a/lib/ohai/plugins/linux/platform.rb
+++ b/lib/ohai/plugins/linux/platform.rb
@@ -54,7 +54,7 @@ Ohai.plugin(:Platform) do
   end
 
   #
-  # Cached /etc/os-release info Hash.  Also has logic for Cisco Nexus
+  # Cached /etc/os-release info Hash. Also has logic for Cisco Nexus
   # switches that pulls the chained CISCO_RELEASE_INFO file into the Hash (other
   # distros can also reuse this method safely).
   #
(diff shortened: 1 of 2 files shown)


lib/ohai/plugins/linux/platform.rb
```diff
@@ -137,13 +137,15 @@ def platform_family_from_platform(plat)
     when /debian/, /ubuntu/, /linuxmint/, /raspbian/, /cumulus/, /kali/, /pop/
       # apt-get+dpkg almost certainly goes here
       "debian"
-    when /oracle/, /centos/, /redhat/, /almalinux/, /scientific/, /enterpriseenterprise/, /xcp/, /xenserver/, /cloudlinux/, /ibm_powerkvm/, /parallels/, /nexus_centos/, /clearos/, /bigip/, /alibabalinux/, /sangoma/ # Note that 'enterpriseenterprise' is oracle's LSB "distributor ID"
+    when /oracle/, /centos/, /redhat/, /almalinux/, /scientific/, /enterpriseenterprise/, /xenserver/, /xcp-ng/, /cloudlinux/, /ibm_powerkvm/, /parallels/, /nexus_centos/, /clearos/, /bigip/, /alibabalinux/, /sangoma/ # Note that 'enterpriseenterprise' is oracle's LSB "distributor ID"
```
jaymzh, line 140: should these be in some sort of order? YOu switched the order here... do we want to alphabetize them?

jaymzh (approved):

PR #1635, opened 2021-03-13
Stick the common matched platforms first

tas50:
Nothing super scientific here, but it's about 10% faster to match if the data is hit sooner. Stick the super obscure stuff last.

### DIFF
--- a/lib/ohai/plugins/linux/platform.rb
+++ b/lib/ohai/plugins/linux/platform.rb
@@ -134,10 +134,10 @@ Ohai.plugin(:Platform) do
   #
   def platform_family_from_platform(plat)
     case plat
-    when /debian/, /ubuntu/, /linuxmint/, /raspbian/, /cumulus/, /kali/, /pop/
+    when /ubuntu/, /debian/, /linuxmint/, /raspbian/, /cumulus/, /kali/, /pop/
       # apt-get+dpkg almost certainly goes here
       "debian"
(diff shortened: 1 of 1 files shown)


lib/ohai/plugins/linux/platform.rb
```diff
@@ -134,10 +134,10 @@ def platform_id_remap(id)
   #
   def platform_family_from_platform(plat)
     case plat
-    when /debian/, /ubuntu/, /linuxmint/, /raspbian/, /cumulus/, /kali/, /pop/
+    when /ubuntu/, /debian/, /linuxmint/, /raspbian/, /cumulus/, /kali/, /pop/
       # apt-get+dpkg almost certainly goes here
       "debian"
-    when /oracle/, /centos/, /redhat/, /almalinux/, /scientific/, /enterpriseenterprise/, /xenserver/, /xcp-ng/, /cloudlinux/, /ibm_powerkvm/, /parallels/, /nexus_centos/, /clearos/, /bigip/, /alibabalinux/, /sangoma/ # Note that 'enterpriseenterprise' is oracle's LSB "distributor ID"
+    when /centos/, /redhat/, /oracle/, /almalinux/, /scientific/, /enterpriseenterprise/, /xenserver/, /xcp-ng/, /cloudlinux/, /alibabalinux/, /sangoma/, /clearos/, /parallels/, /ibm_powerkvm/, /nexus_centos/, /bigip/ # Note that 'enterpriseenterprise' is oracle's LSB "distributor ID"
```
jaymzh, line 140: I'd assume xen would top 5, a bit surprised. But you have the numbers...

jaymzh (approved):

lamont-granquist (approved):

lib/ohai/plugins/linux/platform.rb
tas50: We don't have the data, but xen is probably pretty low. I wasn't able to find any accounts that mentioned it.

PR #1643, opened 2021-04-02
Properly identify CentOS Stream as centos_stream for platform

ramereth:
Currently ohai returns `centos` on CentOS Stream systems for platform which makes it difficult for users to know when they are on Stream or not. This properly identifies the system as `centos_stream` if it is indeed on Stream.

Changes include:

- Fix `read_os_release_info` to use `each_line` instead of split so that it includes the whole line, including spaces. This was causing issues when trying to parse the various fields. Also strip out newline characters which are included.
- Add `os_release_file_is_centos_stream?` method to detect `Stream` in `NAME` field
- Add `centos_stream` as rhel `platform_family`

### DIFF
--- a/lib/ohai/plugins/linux/platform.rb
+++ b/lib/ohai/plugins/linux/platform.rb
@@ -46,8 +46,8 @@ Ohai.plugin(:Platform) do
   def read_os_release_info(file)
     return nil unless file_exist?(file)
 
-    file_read(file).split.inject({}) do |map, line|
-      key, value = line.split("=")
+    file_read(file).each_line.inject({}) do |map, line|
+      key, value = line.chomp.split("=")
       map[key] = value.gsub(/\A"|"\Z/, "") if value
       map
     end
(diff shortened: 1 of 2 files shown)


ramereth:
FWIW I also created https://github.com/chef/os_release/pull/11 so the os-release files could be properly documented.

lib/ohai/plugins/linux/platform.rb
```diff
@@ -81,6 +81,14 @@ def os_release_file_is_cisco?
     file_exist?("/etc/os-release") && os_release_info["CISCO_RELEASE_INFO"]
   end
 
+  #
+  # If /etc/os-release indicates we are CentOS Stream
+  #
+  # @returns [Boolean] if we are CentOS Stream according to /etc/os-release
+  def os_release_file_is_centos_stream?
+    file_exist?("/etc/os-release") && os_release_info["NAME"].match?("Stream")
```
davide125, line 89: I'd recommend specifically testing for `CentOS Stream` here in case something else comes up in the future with `Steam` in the name.

ramereth:
FYI: based on the discussion on slack, I'm going to abandon this PR so that we don't "fork" the two platforms. Instead I'll make a PR in chef-utils for a `centos_stream_platform?` helper.

PR #1645, opened 2021-04-07
Create os_release plugin for parsing data from /etc/os-release

ramereth:
On newer Linux systems which use systemd, they should include `/etc/os-release` [1] which includes a lot of platform information that can be useful downstream.

[1] https://www.freedesktop.org/software/systemd/man/os-release.html

NOTE: If accepted, I'll make a separate PR which will update linux/platform to depend on this plugin and optimize the use of reading `/etc/os-release` in multiple places. This can also be useful in chef/chef#11296 in case `lsb_release` isn't available. I didn't want to include this in this PR in case we could back port this to 16.


tas50:
This is something I've wanted us to do here for a while. Nice work.

PR #1648, opened 2021-04-09
Use scutil to get hostname and computername

tas50:
Fetch the hostname and computername in a more macOS way.

### DIFF
--- a/lib/ohai/plugins/hostname.rb
+++ b/lib/ohai/plugins/hostname.rb
@@ -42,7 +42,7 @@ Ohai.plugin(:Hostname) do
   end
 
   # forward and reverse lookup to canonicalize FQDN (hostname -f equivalent)
-  # this is ipv6-safe, works on ruby 1.8.7+
+  # this is ipv6-safe
   def resolve_fqdn
     require "socket" unless defined?(Socket)
     require "ipaddr" unless defined?(IPAddr)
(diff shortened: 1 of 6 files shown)


lib/ohai/plugins/hostname.rb
```diff
@@ -96,8 +96,8 @@ def collect_hostname
   end
 
   collect_data(:darwin) do
-    hostname from_cmd("hostname -s")
-    machinename from_cmd("hostname")
+    hostname from_cmd("scutil --get HostName")
+    machinename from_cmd("scutil --get ComputerName")
```
lamont-granquist, line 100: machinename is the fallback for the node name and i'm very strongly concerned about using ComputerName here, because mac blogs suggest that this is the "user friendly hostname that shows up on the login screen" and might be "Bob's Awesome MacBook" or some kind of garbage like that.

the name "machinename" is just a label to separate it from everything else.  The way it is implemented in the code this is "the name of the thing which stands a better chance of being a proper FQDN to have be the default node_name which doesn't fall through DNS resolution".  This is because `hostname` is `hostname -s` all over the place which is shit.  Then `fqdn` is `hostname -f` all over the place and that's terrible due to DNS.  So `machinename` is DEFINED BY the usage on linux which is `hostname` and that's the much more sane thing to use to feed to node_name, and it was created to deal with all the issues we had deciding on a default node_name in the absence of one being given on knife bootstrap.

So, both ComputerName and LocalHostName are "names that advertise things to humans" which both seem wrong, and I'd suggest that HostName is the only thing which fits both of these.  If we want to get creative we could go the other way and assign one of those to `node['hostname']` instead of machinename, but IDK what chaos that would cause in cookbook-land.

lamont-granquist (changes requested):

tas50:
Closing this out since there's probably some fallout from doing it this way

PR #1652, opened 2021-04-16
Fall back to v4-only getaddrinfo if we get socket errors

jaymzh:
Handle the case where we have either v4- or v6-specific configuration
or network issues that causes gethostaddr() to fail

Note this was found due to a specific case: On OSX - and possibly others
(freebsd), if you have a linklocal address for your hostname in
/etc/hosts then `getaddrinfo` for your hostname will raise an exception.

### DIFF
--- a/lib/ohai/plugins/hostname.rb
+++ b/lib/ohai/plugins/hostname.rb
@@ -48,7 +48,28 @@ Ohai.plugin(:Hostname) do
     require "ipaddr" unless defined?(IPAddr)
 
     hostname = from_cmd("hostname")
-    addrinfo = Socket.getaddrinfo(hostname, nil).first
+    begin
+      addrinfo = Socket.getaddrinfo(hostname, nil).first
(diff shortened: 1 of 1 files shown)


lib/ohai/plugins/hostname.rb
```diff
@@ -48,7 +48,15 @@ def resolve_fqdn
     require "ipaddr" unless defined?(IPAddr)
 
     hostname = from_cmd("hostname")
-    addrinfo = Socket.getaddrinfo(hostname, nil).first
+    begin
+      addrinfo = Socket.getaddrinfo(hostname, nil).first
+    rescue SocketError
+      # On OSX - (and maybe others), if your v6 address in /etc/hosts happens to
```
tas50, line 54:
```suggestion
      # On macOS - (and maybe others), if your v6 address in /etc/hosts happens to
```

lamont-granquist (approved):

PR #1657, opened 2021-04-22
Dead links fixed

craftsbyshuvro:
Dead link fixed, If you need other revisions of the link you can get that from https://web.archive.org/ . You just need to search by the link and choose available revisions according to date.


tas50:
Thanks @craftsbyshuvro

Issue #1598, opened 2021-01-21
If a pugin throws an exception, it should be logged as a warning


axos88:
Shouldn't this line be a warning? This way if a plugin throws an exception it is swallowed with the default log config, which is highly unexpected. 

https://github.com/chef/ohai/blob/f4e5b68f35de8a5bdca06a2b2fff5fd8337f611e/lib/ohai/runner.rb#L64

lamont-granquist:
We don't seem to throw exceptions on a vanilla ubuntu box or a vanilla macos box so i think i'm okay for this in ohai-17/chef-17

axos88-da:
Custom plugins can potentially throw exceptions. (My plugin threw an exception because the API key was disallowed to connect from that particular IP address).
Currently these exceptions are entirely swallowed, and the provided data is left out from the automatic attributes, and it's not straightforward at all why, not even setting the log level to info or debug.

Some of the possible reasons could include 
  * The file is not parsed / 
  * the plugin file not found 
  * the plugin file is not in the correct place 
  * the plugin it is not working correctly - but why isn't it, since there are no exceptions logged?
and so on.

The fact that a plugin throws an exception is definitely not a trace level exception, it's an info level at least, but I'd vote for a warning. It is an exceptional condition, that can be attempted to be fixed by swallowing it, and leaving out those ohai attributes, but that will almost certainly cause the che to fail along the run, since it was probably installed because something needs it.

axos88-da:
Also it's not clear from the docs that custom plugins' exceptions will be silently swallowed either.

andyjdavis:
Here is a (very small) pull request that makes this change. https://github.com/chef/ohai/pull/1668

PR #1668, opened 2021-05-19
log plugin exception at warn level

andyjdavis:
## Description
This is a potential fix for https://github.com/chef/ohai/issues/1598

If a plugin throws an exception it should be logged at a higher level to boost visibility.

## Related Issue
https://github.com/chef/ohai/issues/1598

## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [ x] I have read the **CONTRIBUTING** document.
- [ x] I have run the pre-merge tests locally and they pass.
- [ -] I have updated the documentation accordingly.
- [? ] I have added tests to cover my changes.
- [ x] All new and existing tests passed.
- [ x] All commits have been signed-off for [the Developer Certificate of Origin](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).

### DIFF
--- a/lib/ohai/runner.rb
+++ b/lib/ohai/runner.rb
@@ -61,7 +61,7 @@ module Ohai
         rescue Ohai::Exceptions::Error, SystemExit # SystemExit: abort or exit from plug-in should exit Ohai with failure code
           raise
         rescue Exception => e
-          logger.trace("Plugin #{plugin.name} threw exception #{e.inspect} #{e.backtrace.join("\n")}")
+          logger.warn("Plugin #{plugin.name} threw exception #{e.inspect} #{e.backtrace.join("\n")}")
         end
       end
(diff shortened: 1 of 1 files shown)


andyjdavis:
There are some linting errors not related to my change. I'm unsure if I should just fix them or leave them be.

jaymzh:
I'm weighing in as in full support of this.
@lamont-granquist is trying to verify this won't be too noisy... if he feels it is, we might suggest `debug` instead of `warn`.
But I don't think that's the case and I'm a fan of `warn`.

So :+1: from me, but waiting on Lamont.

lamont-granquist:
Yeah, it looks like I'm finding less issues than I expected and lots of the typical plugins that go surfing for things like erlang when they don't find them are now well-behaved, so my brain may be years out of date at this point.

PR #1694, opened 2021-09-29
Add plugin for btrfs

boryas:
The plugin collects btrfs usage information by block_group type (data,
metadata, system) for every btrfs filesystem and mount.

N.B. it doesn't currently support any raid mode besides "dup"
correctly, since the main motivation is analyzing metadata usage.

Signed-off-by: Boris Burkov <boris@bur.io>

## Types of changes

## Checklist:

### DIFF
--- a/lib/ohai/plugins/linux/btrfs.rb
+++ b/lib/ohai/plugins/linux/btrfs.rb
@@ -1,0 +1,60 @@
+# frozen_string_literal: true
+#
+# Author:: Boris Burkov <boris@bur.io>
+# Copyright:: Copyright (c) 2021 Facebook, Inc.
+# License:: Apache License, Version 2.0
+#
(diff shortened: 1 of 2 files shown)


boryas:
I was testing this further on a live system and realized the way I handled "dup" was incorrect. Rather than try to read the file in the raid dir, I changed it to just collect the raid mode and make it available to the consumer. I then force pushed the fix to my branch.

jaymzh:
Having a whole separate plugin for this that roughly mirrors filesystem doesn't seem like the right approach to me.

Instead, I'd say just add this to the `filesystem` plugin. If the filesystem is btrfs, grab btfrs info and add it to the data.

tas50:
I'm with @jaymzh. This seems like a lot of duplicate data which we're trying to avoid in Ohai.

PR #1718, opened 2021-11-19
Add new rpm plugin

davide125:
Add a basic plugin to expose configuration information from the RPM
package manager. I specifically wrote this to expose the `_db_backend`
macro, but figured it'd be useful to make it cover a bit more while at
it.

This plugin is enabled on all non-Windows platforms, because a port of
RPM exists (and is sometimes used) on pretty much all platforms, even
where it's not usually installed by default.

Signed-off-by: Davide Cavalca <dcavalca@fb.com>

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Chore (non-breaking change that does not add functionality or fix an issue)

## Checklist:

- [x] I have read the **CONTRIBUTING** document.
- [x] I have run the pre-merge tests locally and they pass.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.
- [x] All commits have been signed-off for [the Developer Certificate of Origin](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).

PR #1719, opened 2021-12-02
Fix parsing bug in the rpm plugin

davide125:

## Description
The rpm plugin that was added in https://github.com/chef/ohai/pull/1718 had a bug in the parsing logic that caused it to swallow one-line macros when they were immediately followed by another one. This PR fixed that, and adds a few comments to hopefully make the logic easier to understand.

## Related Issue
https://github.com/chef/ohai/pull/1718

## Types of changes

## Checklist:


lib/ohai/plugins/rpm.rb
```diff
@@ -91,17 +91,25 @@
...
+          _prefix, name, value = line.split(" ", 3)
         else
+          # continuations have no prefix and just append to the previous one
           value += "\n#{line}"
         end
       end
+      if name
+        # ensure the last macro definition is also stored
+        rpm[:macros][name] = value
+      end
```
jaymzh, line 112: The comment where it is and the loop being big enough, I'd say this is a bit unclear. Perhaps change it this way.

```suggestion

      # Once we're done parsing all lines, we might have a parsed definition
      # we haven't stored - if so, store it.
      if name
        rpm[:macros][name] = value
      end
```

jaymzh (approved):
I think this would clarify things a bit. LMK what you think. Will merge either way. Clearly correct bugfix.

lib/ohai/plugins/rpm.rb
davide125: Comment updated, thanks!

PR #1735, opened 2022-02-17
Stop checking encoding names

casperisfine:
Comparing the names is much less efficient than comparing the instance directly.
    
It may also change in the future: https://bugs.ruby-lang.org/issues/18576


lamont-granquist (approved):

lamont-granquist:
Even though this is kinda tiny it would help to have the DCO signoff to get it merged:

https://github.com/chef/chef/blob/main/CONTRIBUTING.md#developer-certification-of-origin-dco

casperisfine:
> it would help to have the DCO signoff

Done.

PR #1736, opened 2022-02-17
Network: Add support for parsing multipath routing

anitazha:
Signed-off-by: Anita Zhang <the.anitazha@gmail.com>

## Description
We have some hosts using multipath routing and the network plugin is failing to set any interface information for the interfaces with such multipath configuration. This PR updates `check_routing_table()` in the network plugin to support multipath routing.

## Types of changes

## Checklist:

### DIFF
--- a/lib/ohai/plugins/linux/network.rb
+++ b/lib/ohai/plugins/linux/network.rb
@@ -80,9 +80,22 @@ Ohai.plugin(:Network) do
       line.strip!
       logger.trace("Plugin Network: Parsing #{line}")
       if /\\/.match?(line)
+        # If we have multipath routing, then the first part will be a normal
+        # looking route:
+        #   default proto ra metric 1024 <other options>
+        # Each successive part after that is a hop without those options.
+        # So the first thing we do is grab that first part, and split it into
+        # the route destination ("default"), and the route options.
         parts = line.split("\\")
-        route_dest = parts.shift.strip
-        route_endings = parts
+        route_dest, dest_opts = parts.first.split(nil, 2)
+        # Then all the route endings, generally just nexthops.
+        route_endings = parts[1..]
+        if dest_opts && !dest_opts.empty?
+          # Route options like proto, metric, etc. only appear once for each
+          # multipath configuration. Prepend this information to the route
+          # endings so the code below will assign the fields properly.
+          route_endings.map! { |e| e.include?("nexthop") ? "#{dest_opts} #{e}" : e }
+        end
       elsif line =~ /^([^\s]+)\s(.*)$/
         route_dest = $1
         route_endings = [$2]
--- a/spec/unit/plugins/linux/network_spec.rb
+++ b/spec/unit/plugins/linux/network_spec.rb
@@ -1219,6 +1219,46 @@ describe Ohai::System, "Linux Network Plugin" do
         end
       end
 
+      describe "when there're multipath routes" do
+        let(:linux_ip_route) do
+          <<~EOM
+            10.5.4.0/24 proto static \\ nexthop via 10.5.4.1 dev eth0 weight 1\\ nexthop via 10.5.4.2 dev eth0 weight 1
+          EOM
+        end
+
+        let(:linux_ip_route_inet6) do
+          <<~EOM
+            default proto ra metric 1024 expires 1797sec \\        nexthop via fe80::c018:50ff:fe41:ca33 dev eth0.11 weight 1 \\        nexthop via fe80::c018:50ff:fe41:ca33 dev eth0 weight 1 pref medium
+            default proto ra metric 1050 expires 1796sec \\        nexthop via fe80::c018:50ff:fe3f:9184 dev eth3 weight 1 \\        nexthop via fe80::c018:50ff:fe3f:9184 dev eth13 weight 1 pref medium
+          EOM
+        end
+
+        before do
+          plugin.run
+        end
+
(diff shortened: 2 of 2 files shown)


lib/ohai/plugins/linux/network.rb
```diff
@@ -81,8 +81,16 @@ def check_routing_table(family, iface, default_route_table)
       logger.trace("Plugin Network: Parsing #{line}")
       if /\\/.match?(line)
         parts = line.split("\\")
-        route_dest = parts.shift.strip
+        dest_parts = parts.shift.strip.split("\s", 2)
+        route_dest = dest_parts[0]
         route_endings = parts
```
jaymzh, line 86: This is pretty confusing way to do this in ruby.

The hidden shift changes parts, and then you assign a pointer of parts to route_endings (everything in ruby is a pointer).

It totally works, but I'd much rather something like this, which I find easier to follow:

```ruby
# If we have multiroute, then the first part will be a normal looking route:
#   default proto ra metric 1024 <other options>
# and then each successive part after that is hops that won't have those options.
# so the first thing we do, is grab that first part, and split it in to the route dest ("default")
# in the above, and then all the options for that route. 
parts = line.split('\\')
route_dest, dest_opts = parts.first.split(nil, 2)
# then all the route endings, generally just hops:
route_endings = parts[1..]
if dest_parts.count > 1
  # your comment here, it's excellent.
  route_endings.map! { |e| e.include?("nexthop") ? "#{dest_opts} #{e}" : e }
end
```

spec/unit/plugins/linux/network_spec.rb
```diff
@@ -1219,6 +1219,38 @@
...
+          expect(plugin["network"]).not_to be_nil
+        end
+
+        it "sets default ipv6 interface and gateway" do
+          expect(plugin["network"]["default_inet6_interface"]).to eq("eth0")
+          expect(plugin["network"]["default_inet6_gateway"]).to eq("fe80::c018:50ff:fe41:ca33")
+        end
+
+        it "sets the routes properly" do
+          expect(plugin["network"]["interfaces"]["eth0"]["routes"]).to eq([Mash.new( destination: "default", family: "inet6", proto: "ra", "via": "fe80::c018:50ff:fe41:ca33", "metric": "1024" )])
```
jaymzh, line 1247: be consistent on your hash syntax
```suggestion
          expect(plugin["network"]["interfaces"]["eth0"]["routes"]).to eq([Mash.new( destination: "default", family: "inet6", proto: "ra", via: "fe80::c018:50ff:fe41:ca33", metric: "1024" )])
```

spec/unit/plugins/linux/network_spec.rb
```diff
@@ -1219,6 +1219,38 @@
...
+          expect(plugin["network"]["interfaces"]["eth0.11"]["routes"]).to eq([Mash.new( destination: "default", family: "inet6", proto: "ra", "via": "fe80::c018:50ff:fe41:ca33", "metric": "1024" )])
+          expect(plugin["network"]["interfaces"]["eth13"]["routes"]).to eq([Mash.new( destination: "default", family: "inet6", proto: "ra", "via": "fe80::c018:50ff:fe3f:9184", "metric": "1050" )])
+          expect(plugin["network"]["interfaces"]["eth3"]["routes"]).to eq([Mash.new( destination: "default", family: "inet6", proto: "ra", "via": "fe80::c018:50ff:fe3f:9184", "metric": "1050" )])
```
jaymzh, line 1250: same

lib/ohai/plugins/linux/network.rb
```diff
@@ -81,8 +81,16 @@ def check_routing_table(family, iface, default_route_table)
       logger.trace("Plugin Network: Parsing #{line}")
       if /\\/.match?(line)
         parts = line.split("\\")
-        route_dest = parts.shift.strip
+        dest_parts = parts.shift.strip.split("\s", 2)
```
jaymzh, line 84: don't specify `\s`, just use the default splitting which is smart whitespace split.

jaymzh (changes requested):

anitazha:
Thanks Phil! Updated with your suggestions

spec/unit/plugins/linux/network_spec.rb
```diff
@@ -1219,6 +1219,38 @@
         end
       end
 
+      describe "when there're multipath routes" do
+        let(:linux_ip_route) { "" }
```
jaymzh, line 1223: just to make sure people don't break this in the future, can we add a v4 test, please?

jaymzh (commented):
Code looks good, sorry I missed the lack of v4-test below, but I suspect that's gonna be the most-commonly-used path in the code, so lets just have a test for it as well?

anitazha:
There was an existing test case for multipath IPv4 but without the extra options on the first line (line 165 of network_spec.rb). The IPv4 test case I added includes options.

jaymzh (approved):

jaymzh:
Looks great. I'll merge when tests pass. I'm out of town this weekend so feel free to poke me next week if I haven't merged.

tas50:
@anitazha is this something we'd want to backport to Ohai 16 and 17?

anitazha:
@tas50 It could be useful to have since any default IPv6 routes with options using multipath just break without it. But I think multipath routing isn't super widely used so most people won't run into it.

PR #1741, opened 2022-03-11
networking/linux: map src only default routes accordingly

mattp-:
in certain configurations a default route can be set via src only / not
dev, but should still be accounted for in terms of calculating the
'favored' default route / what is chosen as ipaddress for a given node.

This change augments the previously added changes to account for nexthop
multipath routing; but only in scenarios src address is set but only in
scenarios src address is set.

## Types of changes

## Checklist:


spec/unit/plugins/linux/network_spec.rb
```diff
@@ -1335,6 +1335,42 @@
...
       end
 
+      describe "when there's a source field in a local route entry but no dev" do
+        let(:linux_ip_route) do
+          <<~EOM
+            default proto bird src 172.16.19.39 metric 10 \\      nexthop via 10.1.81.1 dev eth0 weight 20
+            default via 10.116.201.1 dev eth0 metric 10 src 10.116.201.1
+            10.116.201.0/24 dev eth0  proto kernel  src 10.116.201.76
+            192.168.0.0/24 dev eth0  proto kernel  src 192.168.0.2
+            default via 10.116.201.254 dev eth0 metric 20 src 10.116.201.74
```
jaymzh, line 1345: It's unclear to me why we pick the no-dev route here over the dev route at the same metric.

Sorry, we're heading into parts of routing that I'm not super familiar with. Do we always prefer a no-dev route over a dev route at the same metric? Or are we picking it some other way? Based on the code it seems like we just always prefer no-dev routes?

jaymzh (changes requested):
Can we get some extra comments on how this is supposed to work?

spec/unit/plugins/linux/network_spec.rb
mattp-: woops, good catch. in this specific case the route on line 1342 is being discarded by the condition here https://github.com/chef/ohai/blob/main/lib/ohai/plugins/linux/network.rb#L135
however, your question is valid. I made it so we explicitly rank routes that define a device over ones that don't if they have the same metric. if theres a specific part you'd like more comments/explanation on let me know I can try to add/explain further

spec/unit/plugins/linux/network_spec.rb
```diff
@@ -1335,6 +1335,59 @@
...
+        end
+        let(:linux_ip_route_inet6) { "" }
+
+        it "maps the no-dev route and sets ipaddress" do
+          plugin.run
+          expect(plugin["ipaddress"]).to eq("172.16.19.39")
+        end
+      end
+
+      # note: default via 10.116.201.1 discarded because src not on eth0
```
jaymzh, line 1355: I"m not following. It has `dev eth0` in it, how do we know that 10.116.201.1 isn't the right source?  Or 172.16.19.39? How do we know that 10.116.201.74 is the "right" source?

jaymzh, line 1355: I"m not following. It has `dev eth0` in it, how do we know that 10.116.201.1 isn't the right source?  Or 172.16.19.39? How do we know that 10.116.201.74 is the "right" source?

svenkatr: 10.116.201.1 is the gateway IP, and not considered as source.
metric 5 is better (higher priority) than metric 10 - Hence line 1363 wins against everything else.
[The test data should've been sorted by metric, but the concept is the same]
Hence the src related to that route is chosen - 10.116.201.74

rmcleod8:
@jaymzh is there anything we can help clear up or include regarding this pr to continue moving it forward?

spec/unit/plugins/linux/network_spec.rb
jaymzh: I had missed the ifconfig at the top of the file and the code below your code that does the 'Sanity check'.

But I'm still confused. Your comment here says:

```
# note: default via 10.116.201.1 discarded because src not on eth0
```

OK, I see the code you mean:

```ruby
        # a sanity check, especially for Linux-VServer, OpenVZ and LXC:
        # don't report the route entry if the src address isn't set on the node
        # unless the interface has no addresses of this type at all
        if route_entry[:src]
          addr = iface[route_int][:addresses]
          unless addr.nil? || addr.key?(route_entry[:src]) ||
              addr.values.all? { |a| a["family"] != family[:name] }
            logger.trace("Plugin Network: Skipping route entry whose src does not match the interface IP")
            next
          end
        end
```

OK got it. But then why are any routes considered:

* `10.1.81.1` is discarded because the src `172.16.19.39` isn't on eth0
* `10.116.201.1` is discarded because the src `10.116.201.1` (same thing, nonsensicle route), isn't on eth0
* Leaving the only route left, `10.116.201.254`, but that src, `10.116.201.74` ALSO isn't on the interface, so, why is that one considered? Shouldn't that also be discarded?

The routing code here is pretty dense, each of these test cases needs a pretty clear comment on why each route is considered or not, and why of the ones that are considered, one wins (usually just "metric is lowest").

spec/unit/plugins/linux/network_spec.rb
```diff
@@ -1335,6 +1335,59 @@
         end
       end
 
+      describe "when there's a source field in a local route entry but no dev" do
```
jaymzh, line 1338: So if I understand, in this one:
* 172.16.19.39 is in fact on eth0, so that one is considered
* 10.116.201.254 is not on eth0, so that one is not considered
* Metric 10 is lower than metric 20, so the multipath route wins over the only other default route.

spec/unit/plugins/linux/network_spec.rb
```diff
@@ -1335,6 +1335,59 @@
...
+          <<~EOM
+            default proto bird src 172.16.19.39 metric 10 \\      nexthop via 10.1.81.1 dev eth0 weight 20
+            10.116.201.0/24 dev eth0  proto kernel  src 10.116.201.76
+            192.168.0.0/24 dev eth0  proto kernel  src 192.168.0.2
+            default via 10.116.201.254 dev eth0 metric 20 src 10.116.201.74
+          EOM
+        end
+        let(:linux_ip_route_inet6) { "" }
+
+        it "maps the no-dev route and sets ipaddress" do
```
jaymzh, line 1349: what do you mean by no-dev? That route has a dev... every route here has a dev.

jaymzh (changes requested):
Sorry for the delay here. Feel free to ping me on slack if I miss an update.

And sorry if I'm being dense... but I'm still not quite following here, and I'm not sure if I'm just missing something or there's a big gap in the code or tests.

spec/unit/plugins/linux/network_spec.rb
rmcleod8: @jaymzh if you look at line 1341 above, there is no "dev" shown which is what we're attempting to assert.

jaymzh: what ... are you talking about. It's a multipath route that explicit;y has "dev eth0" in the nexthop.

mattp-: devices associated with multi-hop routes aren't considered 'the device associated with the route' ; that logic exists as is today - unrelated to this PR

spec/unit/plugins/linux/network_spec.rb
mattp-: added a clear route by route explanation of why choices are made for each of the new test cases - please take a look

mattp-:
@jaymzh updated and added explicit comments on why each route is/is not considered on each of the added tests; hoping that makes things more clear wrt your previous comments.

spec/unit/plugins/linux/network_spec.rb
```diff
@@ -1335,6 +1336,74 @@
...
+          plugin.run
+          expect(plugin["ipaddress"]).to eq("172.16.19.39")
+        end
+      end
+
+      # notes:
+      # route 172.16.19.39 is the addr of tun0 ; see :linux_ifconfig at top of file
+      # route 10.116.201.1 discarded because src 192.116.201.1 does not match eth0 ("Skipping route entry whose src does not match the interface IP")
```
jaymzh, line 1363: That doesn't match the route:
```
default via 10.116.201.1 dev eth0 metric 10 src 10.116.201.1
```
Did you mean "src 10.116.201.1 does not match..." ?

jaymzh (changes requested):
Thanks! That helps a lot.

The code all seems correct to me, one of your comments I think has a typo in it.

My last request though is - should we not have IPv6 in these tests? It's a lot of code to only add tests for a deprecated IP family :)

spec/unit/plugins/linux/network_spec.rb
mattp-: oops , that line shouldn't be there sorry. its actually not discarded, just not chosen because it is higher priority. Updated the comments to reflect this

mattp-:
> Thanks! That helps a lot.
> 
> The code all seems correct to me, one of your comments I think has a typo in it.
> 
> My last request though is - should we not have IPv6 in these tests? It's a lot of code to only add tests for a deprecated IP family :)

spoke with @svenkatr , I don' think we have a reasonable way to actually create an ipv6 environment to know what this kind of setup would emit for stdout to create the tests from on our end. I'm not sure I should try making it up as it's already caused a bit of confusion from this PR  :) Given the existing coverage is clean I don't think that isn't necessarily a problem. do you have any thoughts on either merging as is or any ideas how to get expected test cases on the ipv6 front?

svenkatr (approved):
Looks good.

jaymzh (approved):
LGTM.

We can loop back around for v6 test if/when we can get the right data.

Thanks!

PR #1754, opened 2022-06-29
Fix FIPS mode detection

stanhu:
## Description

Previously FIPS detection relied on the `OpenSSL::OPENSSL_FIPS`
constant being defined. However, on RedHat operating systems, this
constant is always defined in
`/usr/include/openssl/opensslconf-x86_64.h`. As a result, on such
operating systems FIPS mode would erroneously be labeled as enabled.
This constant is a necessary but not sufficient condition to determine
whether FIPS is actually enabled.

OpenSSL has a runtime `fips_mode` check
(https://wiki.openssl.org/index.php/FIPS_mode()) that should be used
instead. Ruby will use this if the `OPENSSL_FIPS` compile-time
constant is available:
https://github.com/ruby/ruby/blob/685efac05983dee44ce2d96c24f2fcb96a0aebe2/ext/openssl/ossl.c#L413-L428

## Types of changes

## Checklist:

### DIFF
--- a/lib/ohai/plugins/fips.rb
+++ b/lib/ohai/plugins/fips.rb
@@ -30,6 +30,7 @@ Ohai.plugin(:Fips) do
     fips Mash.new
 
     require "openssl" unless defined?(OpenSSL)
-    fips["kernel"] = { "enabled" => OpenSSL::OPENSSL_FIPS }
+
+    fips["kernel"] = { "enabled" => defined?(OpenSSL.fips_mode) && OpenSSL.fips_mode }
   end
 end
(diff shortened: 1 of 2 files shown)


lib/ohai/plugins/fips.rb
```diff
@@ -30,6 +30,6 @@
     fips Mash.new
 
     require "openssl" unless defined?(OpenSSL)
-    fips["kernel"] = { "enabled" => OpenSSL::OPENSSL_FIPS }
+    fips["kernel"] = { "enabled" => OpenSSL.fips_mode }
```
stanhu, line 33: Oh, I see this was reverted in ab6180be02712512a3a8309e75424dc56520b936. OpenSSL v3.0 may not have `fips_mode`.

stanhu:
Thanks @jaymzh. What's left to merge this?

marcparadise (approved):

Issue #1760, opened 2022-07-27
lscpu parser doesn't parse Raspberry Pi 4 (ARM Cortex-A72) output


stanhu:
## Description

As a result of https://github.com/chef/ohai/pull/1454, `node['cpu']` is nil since the `lscpu` output is not parsed properly. I would have expected the fallback to go to `/proc/cpuinfo`.

## Ohai Version

v17.9.0

## Platform Version

Ubuntu 22.04 Linux arm64 

## Ohai Output

[ohai-output.txt](https://github.com/chef/ohai/files/9195438/ohai-output.txt)

## `lscpu` output

Attached are the outputs that can be used in the tests. Notice the indentation and `Core(s) per cluster` instead of `Core(s) per socket`: https://github.com/util-linux/util-linux/blob/60803291acc7f431fd7750de7b2a2f1d45db1f4d/sys-utils/lscpu.c#L883

```
Architecture:            aarch64
  CPU op-mode(s):        32-bit, 64-bit
  Byte Order:            Little Endian
CPU(s):                  4
  On-line CPU(s) list:   0-3
Vendor ID:               ARM
  Model name:            Cortex-A72
    Model:               3
    Thread(s) per core:  1
    Core(s) per cluster: 4
    Socket(s):           -
    Cluster(s):          1
    Stepping:            r0p3
    CPU max MHz:         1500.0000
    CPU min MHz:         600.0000
    BogoMIPS:            108.00
    Flags:               fp asimd evtstrm crc32 cpuid
Caches (sum of all):
  L1d:                   128 KiB (4 instances)
  L1i:                   192 KiB (4 instances)
  L2:                    1 MiB (1 instance)
Vulnerabilities:
  Itlb multihit:         Not affected
  L1tf:                  Not affected
  Mds:                   Not affected
  Meltdown:              Not affected
  Mmio stale data:       Not affected
  Spec store bypass:     Vulnerable
  Spectre v1:            Mitigation; __user pointer sanitization
  Spectre v2:            Vulnerable
  Srbds:                 Not affected
  Tsx async abort:       Not affected
```

[cpuinfo-aarch64-rpi4.output.txt](https://github.com/chef/ohai/files/9195444/cpuinfo-aarch64-rpi4.output.txt)
[lscpu-aarch64-rpi4-cores.output.txt](https://github.com/chef/ohai/files/9195445/lscpu-aarch64-rpi4-cores.output.txt)
[...]

Issue #1755, opened 2022-07-04
CPU total is 0 when threads-per-core is 0


stanhu:
## Description

Since https://github.com/chef/ohai/pull/1454, using `node['cpu']['total']` appears to return 0 for some virtualized platforms when `threads per core` is 0.

This was reported in https://gitlab.com/gitlab-org/omnibus-gitlab/-/issues/6895.

## Ohai Version

v17.9.0

## Platform Version

Linux x86 and arm64

## Ohai Output

Example `lscpu` output:

```
Architecture:         x86_64
CPU op-mode(s):       32-bit, 64-bit
Byte Order:           Little Endian
Address sizes:        40 bits physical, 48 bits virtual
CPU(s):               16
On-line CPU(s) list:  0,2,5,9
Off-line CPU(s) list: 1,3,4,6-8,10-15
Thread(s) per core:   0
Core(s) per socket:   4
Socket(s):            2
NUMA node(s):         2
Vendor ID:            GenuineIntel
CPU family:           6
Model:                26
Model name:           Intel(R) Xeon(R) CPU           X5570  @ 2.93GHz
Stepping:             5
CPU MHz:              1596.085
BogoMIPS:             5852.18
Virtualization:       VT-x
L1d cache:            32K
L1i cache:            32K
L2 cache:             256K
L3 cache:             8192K
NUMA node0 CPU(s):    0,2,4,6,8,10,12,14
NUMA node1 CPU(s):    1,3,5,7,9,11,13,15
Flags:                fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ht tm pbe syscall nx rdtscp lm constant_tsc arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc cpuid aperfmperf pni dtes64 monitor ds_cpl vmx est tm2 ssse3 cx16 xtpr pdcm dca sse4_1 sse4_2 popcnt lahf_lm pti tpr_shadow vnmi flexpriority ept vpid dtherm ida
```

cc @ramereth @jaymzh

PR #1761, opened 2022-07-27
Fall back to /proc/cpuinfo if lscpu doesn't have CPU totals

stanhu:
## Description

https://github.com/chef/ohai/pull/1454 made `lscpu` the primary way Ohai gathers CPU data, but the parser doesn't handle the output from a Raspberry Pi 4 Cortex-A72. For example, an ARM cluster has the output `Core(s) per cluster` instead of `Core(s) per socket`.

We should fall back to `/proc/cpuinfo` if we don't get reasonable data back. It's not enough just to have `lscpu` output.

## Related Issue

* https://github.com/chef/ohai/issues/1760
* https://github.com/chef/ohai/issues/1755

## Types of changes

## Checklist:

### DIFF
--- a/lib/ohai/plugins/cpu.rb
+++ b/lib/ohai/plugins/cpu.rb
@@ -344,15 +344,23 @@ Ohai.plugin(:CPU) do
     cpuinfo
   end
 
+  # Check if the `lscpu` data looks reasonable
+  def valid_lscpu?(lscpu)
+    return false if lscpu.empty?
+    return false if %i{total real cores}.any? { |key| lscpu[key].to_i == 0 }
+
+    true
+  end
+
   collect_data(:linux) do
     cpuinfo = parse_cpuinfo
     lscpu = parse_lscpu(cpuinfo)
 
-    # If we don't have any data from lscpu then get it from /proc/cpuinfo
-    if lscpu.empty?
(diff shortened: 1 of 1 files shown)


lib/ohai/plugins/cpu.rb
```diff
@@ -344,12 +344,20 @@ def parse_cpuinfo
...
+  end
+
   collect_data(:linux) do
     cpuinfo = parse_cpuinfo
     lscpu = parse_lscpu(cpuinfo)
 
-    # If we don't have any data from lscpu then get it from /proc/cpuinfo
-    if lscpu.empty?
+    # If we don't have any sensible data from lscpu then get it from /proc/cpuinfo
+    if !valid_lscpu?(lscpu)
```
marcparadise, line 360: Could you reverse this, so that: 

```
if valid_lscpu?(lscpu) 
  cpu lscpu
else 
...
````
Keeping the positive condition first is consistent with style throughout chef-infra.

marcparadise (dismissed):

jaymzh:
Ping @stanhu

stanhu:
@jaymzh Done, thanks.

jaymzh (approved):

Issue #1765, opened 2022-08-10
fetching metadata on VMs with pre-defined user-data of Alibaba Cloud causing a never-end recursion, maybe other Cloud Service Providers are affected too


ETiV:
## Description

> The content of `user-data` API should be stored properly, not to be continuously requested line by line.

On Alibaba Cloud, the user-data API (URL: `http://100.100.100.200/2016-01-01/user-data` ) responds the cloud-init config file in a multi-lined plain text way, and the first line is `#cloud-config`( or `#!/bin/bash` ) , which caused the URL will be requested later becomes `http://100.100.100.200/2016-01-01/user-data/#cloud-config`, the metadata API will respond the same content as before, so the recursion will never end.

## Ohai Version

I'm using GitLab: GitLab (latest, 15.2.2) -> chef (17.10.0) -> ohai (17.9.0)

## Platform Version

VMs on Alibaba Cloud, but I think those Clouds having the same user-data responding logic will have the same issue.

## Ohai Output

I'm sorry I'll put the output of GitLab here.

```
WARN: Plugin Alibaba threw exception #<SystemStackError: stack level too deep> /opt/gitlab/embedded/lib/ruby/2.7.0/net/protocol.rb:277:in `block in write0'

/opt/gitlab/embedded/lib/ruby/2.7.0/net/protocol.rb:275:in `each'
/opt/gitlab/embedded/lib/ruby/2.7.0/net/protocol.rb:275:in `each_with_index'
/opt/gitlab/embedded/lib/ruby/2.7.0/net/protocol.rb:275:in `write0'
/opt/gitlab/embedded/lib/ruby/2.7.0/net/protocol.rb:248:in `block in write'
/opt/gitlab/embedded/lib/ruby/2.7.0/net/protocol.rb:265:in `writing'
/opt/gitlab/embedded/lib/ruby/2.7.0/net/protocol.rb:247:in `write'
/opt/gitlab/embedded/lib/ruby/2.7.0/net/http/generic_request.rb:335:in `write_header'
/opt/gitlab/embedded/lib/ruby/2.7.0/net/http/generic_request.rb:129:in `exec'
/opt/gitlab/embedded/lib/ruby/2.7.0/net/http.rb:1521:in `block in transport_request'
/opt/gitlab/embedded/lib/ruby/2.7.0/net/http.rb:1519:in `catch'
/opt/gitlab/embedded/lib/ruby/2.7.0/net/http.rb:1519:in `transport_request'
[...]

PR #1766, opened 2022-08-11
Fetch the user-data on Alibaba Cloud correctly

ETiV:
The response of `user-data` API should not be fetched line by line.

## Description
On Alibaba Cloud, the response of `user-data` metadata API is not processed properly.

## Related Issue

#1765 

## Types of changes

## Checklist:

This is my first time to coding in Ruby, I've did my best 😂.
If there's any error or the coding style is not meet the repo, please just fix it, thank you.


lib/ohai/mixin/alibaba_metadata.rb
```diff
@@ -38,18 +38,21 @@ def http_get(uri)
         conn.get("/2016-01-01/#{uri}", { "User-Agent" => "chef-ohai/#{Ohai::VERSION}" })
       end
 
-      def fetch_metadata(id = "")
+      def fetch_metadata(id = "", is_dir = true)
```
marcparadise, line 41: What's the intent here for `is_dir`?  What could be a dir[ectory] in this context?

marcparadise (changes requested):
I was able to find docs here: https://www.alibabacloud.com/help/en/elastic-compute-service/latest/view-instance-metadata

But they don't include example output. I don't have this environment available - would you be able to include example output being parsed for user-data and other newline-delimited data in the comments, or a reference to where the behavior is defined?

lib/ohai/mixin/alibaba_metadata.rb
ETiV: because the original code has a problem: it treated a multi-lined content as a result of directory list (you may see the actual multi-lined content can also be a configuration at the bottom), but the correct way is by checking the trailing slash of each item.

the example:
```
-------- only the items with a trailing slash is a directory, those have no trailing slash is a file
$ curl http://100.100.100.200/latest/meta-data/
dns-conf/
eipv4
hostname
instance-id
instance/
...
vpc-id
zone-id

-------- get meta-data/dns-conf/ , as list the directory, it has only one file: nameservers
$ curl http://100.100.100.200/latest/meta-data/dns-conf/
nameservers

-------- meta-data/dns-conf/nameservers is a multi-lined file, not a directory. its content is the config value
$ curl http://100.100.100.200/latest/meta-data/dns-conf/nameservers
100.100.2.136
100.100.2.138

-------- if we get into each line of these contents, like the original code, the metadata API returns a 404, and this metadata will be lost
$ curl -i http://100.100.100.200/latest/meta-data/dns-conf/nameservers/100.100.2.136
HTTP/1.1 404 Not Found
Date: Wed, 17 Aug 2022 09:39:47 GMT
Content-Length: 345
Content-Type: text/plain

```

ETiV:
And, in my opinion, if the purpose of this module is to **ONLY retrieve the [meta-data/...](the meta-data and its sub-directory's contents)**, the root URL should be `http://100.100.100.200/2016-01-01/meta-data/`, not `http://100.100.100.200/2016-01-01/`.

If the starting point is `http://100.100.100.200/2016-01-01/`([as the code here](https://github.com/chef/ohai/blob/main/lib/ohai/mixin/alibaba_metadata.rb#L38)), it will also get [user-data], [dynamic/...], [global-config], [maintenance/...], these things are not [meta-data/...] related I think.

ETiV:
Currently the meta-data API's structure is like this, I've pasted some of the file's content after `=>`

```
http://100.100.100.200/2016-01-01/meta-data/
├── dns-conf/
│   └── nameservers        => `100.100.2.136\n100.100.2.138`
├── eipv4                  => ``
├── hibernation/
│   └── configured         => `false`
├── hostname               => `iZuf64XXXXXXXXXXXX`
├── image-id               => `debian_10_9_x64_20G_alibase_20210521.vhd`
├── instance/
│   ├── instance-name      => `vm-instance-name`
│   ├── instance-type      => `ecs.c5.2xlarge`
│   ├── last-host-landing-time
│   ├── max-netbw-egress
│   ├── max-netbw-ingress
│   ├── virtualization-solution
│   └── virtualization-solution-version
├── instance-id
├── mac
├── network/
│   └── interfaces/
│       └── macs/
│           └── 00:11:22:xx:yy:zz/
│               ├── gateway
│               ├── netmask
│               ├── network-interface-id
│               ├── primary-ip-address
│               ├── private-ipv4s
│               ├── vpc-cidr-block
│               ├── vpc-id
│               ├── vswitch-cidr-block
│               └── vswitch-id
├── network-type
├── ntp-conf/
│   └── ntp-servers
├── owner-account-id
├── private-ipv4
├── public-keys/
│   └── 0/
│       └── openssh-key
├── ram/
│   └── security-credentials/
│       └── AnRAMRoleName
├── region-id
├── serial-number
├── source-address
├── sub-private-ipv4-list
├── vpc-cidr-block
├── vpc-id
├── vswitch-cidr-block
├── vswitch-id
└── zone-id
```

marcparadise:
Thanks for the additional info @ETiV ; these changes look good. 

I agree that it would make more sense to include `meta-data` at the URL; that would also let us get rid of the special case for `user-data`; could you make those updates?

ETiV:
hi @marcparadise , I just pushed a commit to apply what we've talked earlier.

jaymzh (approved):
Assuming tests pass, I'm :+1:

jaymzh (changes requested):
Code is fine, but the tests need updating...

jaymzh:
@ETiV - ping?

spec/unit/mixin/alibaba_metadata_spec.rb
```diff
@@ -0,0 +1,205 @@
...
+
+    ### There's no more /user-data
+    #  context "when fetching user-data" do
+    #   let(:response) { double("Net::HTTP Response", body: CONF_STR, code: "200") }
+    #   it "should not to be continuously requested line by line, and equals to its content" do
+    #     expect { mixin.fetch_metadata("/user-data") }.not_to raise_error
+    #     expect(mixin.fetch_metadata("/user-data")).to eq(CONF_STR)
+    #   end
+    #  end
+
```
jaymzh, line 118: just delete all this.

jaymzh (changes requested):
You have 10 unit tests failing for this plugin specifically.

marcparadise:
@ETiV these changes look good to me, but we're still seeing test failures in the new specs.

jaymzh (changes requested):
All the tests around the alibaba test are failing... please fix them and we'll merge.

jaymzh (changes requested):
Your tests still need updating, please.

Issue #1779, opened 2023-01-21
cloud plugin fail on Oracle Cloud Infrastructure (OCI)


rnt:
### Describe the Enhancement
**cloud plugin** need to detect if virtual machine is located in Oracle Cloud Infrastructure.

Right now we get, in Oracle Linux 8.5 with `chef-18.1.0-1.el8.x86_64` :

```[opc@core-oracle-85 ~]$ ohai cloud
/opt/chef/embedded/lib/ruby/gems/3.1.0/gems/ohai-18.0.26/lib/ohai/system.rb:185:in `attributes_print': I cannot find an attribute named cloud! (ArgumentError)
	from /opt/chef/embedded/lib/ruby/gems/3.1.0/gems/ohai-18.0.26/lib/ohai/application.rb:151:in `block in run_application'
	from /opt/chef/embedded/lib/ruby/gems/3.1.0/gems/ohai-18.0.26/lib/ohai/application.rb:150:in `each'
	from /opt/chef/embedded/lib/ruby/gems/3.1.0/gems/ohai-18.0.26/lib/ohai/application.rb:150:in `run_application'
	from /opt/chef/embedded/lib/ruby/gems/3.1.0/gems/ohai-18.0.26/lib/ohai/application.rb:92:in `block in run'
	from /opt/chef/embedded/lib/ruby/3.1.0/benchmark.rb:311:in `realtime'
	from /opt/chef/embedded/lib/ruby/gems/3.1.0/gems/ohai-18.0.26/lib/ohai/application.rb:90:in `run'
	from /opt/chef/embedded/lib/ruby/gems/3.1.0/gems/ohai-18.0.26/bin/ohai:25:in `<top (required)>'
	from /usr/bin/ohai:184:in `load'
	from /usr/bin/ohai:184:in `<main>'
```

### Describe the Need
We need to allow the **cloud plugin** to detect OCI and return at least the cloud provider.
All users using chef in OCI, must require this feature.

### Current Alternative
Right now we create a method in our cookbook, to check if we are located in OCI or not.

```ruby
  def in_oci?
    has_oci_chassis_asset_tag = false
    if file_exist?("/sys/devices/virtual/dmi/id/chassis_asset_tag")
      file_open("/sys/devices/virtual/dmi/id/chassis_asset_tag").each do |line|
        if /OracleCloud.com/.match?(line)
          has_oci_chassis_asset_tag = true
          break
        end
      end
    end
    has_oci_chassis_asset_tag
  end
```

[...]

PR #1780, opened 2023-01-21
Add cloud provider oci

rnt:
`ohai cloud` fail if the vm is in Oracle Cloud Infrastructure (OCI). This PR add support to this cloud provider to cloud plugin.

## Description
Ohai don't detect OCI has a cloud provider, so for cookbook thats require some changes related to the cloud provider, this is a issue.

Following the [OCI Metadata Documentation](https://docs.oracle.com/en-us/iaas/Content/Compute/Tasks/gettingmetadata.htm) I build this change.

## Related Issue
#1779

## Types of changes

## Checklist:

### DIFF
--- a/lib/ohai/mixin/oci_metadata.rb
+++ b/lib/ohai/mixin/oci_metadata.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+
 #
 # Author:: Renato Covarrubias (<rnt@rnt.cl>)
 # License:: Apache License, Version 2.0
@@ -15,15 +16,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-require "net/http" unless defined?(Net::HTTP)
+require 'net/http' unless defined?(Net::HTTP)
 
 module Ohai
   module Mixin
     module OCIMetadata
-
-      # Trailing dot to host is added to avoid DNS search path
-      OCI_METADATA_ADDR ||= "169.254.169.254"
-      OCI_METADATA_URL ||= "/opc/v2"
+      OCI_METADATA_ADDR = '169.254.169.254'
+      OCI_METADATA_URL = '/opc/v2'
 
       # fetch the meta content with a timeout and the required header
       def http_get(uri)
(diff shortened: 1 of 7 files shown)


lib/ohai/plugins/oci.rb
```diff
@@ -0,0 +1,94 @@
...
+        logger.trace('Plugin oci: Found OracleCloud.com chassis_asset_tag used by oci.')
+        has_oci_chassis_asset_tag = true
+        break
+      end
+    end
+    has_oci_chassis_asset_tag
+  end
+
+  def parse_metadata
+    return nil unless can_socket_connect?(Ohai::Mixin::OCIMetadata::OCI_METADATA_ADDR, 80)
```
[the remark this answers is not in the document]
rnt: I read the thread, and in this case the check is only to not make the call to the API if it is not accessible.
Then if the first call (_instance_) is not a json, nil is also returned.
Only the second call (_vnics_) and the third (_volumeAttachments_) do not include the element in case the response was not valid.

But it has already been detected to be OCI, and the `oci` element will be created empty.

lib/ohai/mixin/oci_metadata.rb
```diff
@@ -0,0 +1,69 @@
...
+        return nil unless response.code == "200"
+
+        if json?(response.body)
+          data = String(response.body)
+          parser = FFI_Yajl::Parser.new
+          parser.parse(data)
+        else
+          response.body
+        end
+      end
```
jaymzh, line 52: If I was able to parse it, then I parse it again? That's really inefficient. How about we only parse it once? :)

jaymzh (changes requested):
We definitely shouldn't fire off the YAJL parser twice if we don't need to.

lib/ohai/mixin/oci_metadata.rb
rnt: Change apply in f7fb897a and change again in e78dbfb7.

In this case, I only followed the same logic for the other cloud providers. and I just looked at the one that was wrong. 

Shall we leave those changes for another PR?

- [lib/ohai/mixin/alibaba_metadata.rb#L41-L72](../blob/main/lib/ohai/mixin/alibaba_metadata.rb#L41-L72)
- [lib/ohai/mixin/gce_metadata.rb#L38-L69](../blob/main/lib/ohai/mixin/gce_metadata.rb#L38-L69)

jaymzh (approved):
Nice, thanks!

lib/ohai/mixin/oci_metadata.rb
rnt: The other PR: #1786  ;)

PR #1788, opened 2023-03-03
Add support for macOS Rapid Security Response updates

nstrauss:
## Description
Apple is adding a new update type to macOS called Rapid Security Responses. These updates change the build version, but not the OS version reported by `sw_vers`. Instead, a new `ProductVersionExtra` key is returned to denote a specific supplemental version only when a RSR is installed. If no RSR is installed the key is not present. 

```
$ sw_vers
ProductName: macOS
ProductVersion: 13.3
ProductVersionExtra: (a)
BuildVersion: 22E7752300f
```

https://support.apple.com/guide/deployment/rapid-security-responses-dep93ff7ea78/web

This PR adds support for reporting on the RSR version when a patch is installed. The information is important to Mac admins since it gives a full picture of current OS build, whether a security vulnerability has been patched, and, depending on Apple's future plans, if there is functionality change which could impact Chef. I'll admit the spec test isn't totally useful, especially since most machines running tests likely won't have an RSR, but it does match real world scenarios. 

## Types of changes

## Checklist:

### DIFF
--- a/spec/unit/plugins/darwin/platform_spec.rb
+++ b/spec/unit/plugins/darwin/platform_spec.rb
@@ -19,36 +19,54 @@
 require "spec_helper"
 
 describe Ohai::System, "Darwin plugin platform" do
+  SW_VERS = "/usr/bin/sw_vers".freeze
+
   before do
     @plugin = get_plugin("darwin/platform")
     allow(@plugin).to receive(:collect_os).and_return(:darwin)
-    @stdout = "ProductName:	Mac OS X\nProductVersion:	10.15.6\nBuildVersion:	19G46c"
-    allow(@plugin).to receive(:shell_out).with("/usr/bin/sw_vers").and_return(mock_shell_out(0, @stdout, ""))
+    @stdout = <<~SW_VERS
+      ProductName:         macOS
+      ProductVersion:      13.3
+      BuildVersion:        22E252
+    SW_VERS
+    allow(@plugin).to receive(:shell_out).with("#{SW_VERS}").and_return(mock_shell_out(0, @stdout, ""))
   end
 
   it "runs sw_vers" do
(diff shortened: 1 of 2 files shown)


tpowell-progress (approved):
Confirmed tests pass.

marcparadise (dismissed):
Please update the commit message to match the PR wording so we have local/repo access to this history and background.

spec/unit/plugins/darwin/platform_spec.rb
```diff
@@ -41,6 +41,11 @@
     expect(@plugin[:platform_version]).to eq("10.15.6")
   end
 
+  it "sets platform_version_extra to ProductVersionExtra" do
+    @plugin.run
+    expect(@plugin[:platform_version_extra]).to eq("(a)").or be_nil
```
jaymzh, line 46: this isn't a meaningful test - it'll be null given the test data... but it'd also be null if you were parsing wrong and the code was broken.

You can have this test be "returns null if no PVE is available", but then add another test that changes the test data (currently set on line 25) and ensure it's parsed properly.

Thanks!

jaymzh (changes requested):

spec/unit/plugins/darwin/platform_spec.rb
nstrauss: @jaymzh Thanks for the guidance here. I'm not overly familiar with writing tests, but did what I could here. I've updated `sw_vers` to match more recent output on macOS as well. Please let me know if there's anything else I should do to get the PR approved.

jaymzh (approved):
We theoretically should move `@stdout` to be a `let` instead of a classvar, but we won't hold you to that standard since you found it this way.

Tests look good, code looks good. I'll merge if the tests pass.

jaymzh:
Thanks!

PR #1804, opened 2023-08-09
Add 3.2 to the list of Ruby CI versions

igor-drozdov:
## Description

Adds 3.2 to the list of Ruby CI versions to run tests against Ruby 3.2

### DIFF
--- a/.github/workflows/exec.yml
+++ b/.github/workflows/exec.yml
@@ -18,7 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ['macos-latest', 'ubuntu-latest', 'windows-latest']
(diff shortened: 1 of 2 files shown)


tpowell-progress:
@igor-drozdov do you only want to add runs of the unit test or exec as well? exec current only runs against 2.7 and 3.0

igor-drozdov:
@tpowell-progress adding it to the tests seems sufficient, but adding it to both CI is more reliable. I've added the Ruby versions in both `unit.yml` and `exec.yml`, but will follow your recommendations on this one 👍

jaymzh (approved):

Issue #1733, opened 2022-01-24
node['fqdn'] missing the DNS Suffix on windows nodes


PrajaktaPurohit:
Internal issue [CHEF-614](https://chefio.atlassian.net/browse/CHEF-614)

When using test-kitchen, I found that on Windows hosts (that are not domain joined) the node['fqdn'] includes the DNS Suffix when using Chef 16.17.18, but not on chef 17.9.26

To Reproduce

I reproduced this by running a kitchen suite against the same platforms (win 2012, 2016, and 2019), with 2 different chef versions.

Expected behavior

node['fqdn'] = hostname.dnssuffix

Actual behavior

node['fqdn'] = hostname

[CHEF-614]: https://chefio.atlassian.net/browse/CHEF-614?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1810, opened 2023-10-03
Add getnameinfo as a fall through case for fqdn resolution

tpowell-progress:
## Description
node['fqdn'] missing the DNS Suffix on windows nodes

If `canonname` includes multiple parts (`'.'`), then assume that that is a FQDN and return.

Otherwise, use `getnameinfo` as a fallback. If that return value is not an IP, return it.

If all else fails, just return what we originally received as an argument.

## Related Issue
Internal issue: CHEF-605

Broken in PR [#1705](https://github.com/chef/ohai/pull/1705)

Fixes [#1733](https://github.com/chef/ohai/issues/1733)

## Types of changes

## Checklist:

### DIFF
--- a/lib/ohai/mixin/network_helper.rb
+++ b/lib/ohai/mixin/network_helper.rb
@@ -65,7 +65,7 @@ module Ohai
         # However, we have found that Windows hosts that are not joined to a domain
         # can return a non-qualified hostname).
         # Use a '.' in the canonname as indicator of FQDN
-        return canonname if /\./.match?(canonname)
+        return canonname if canonname.include?(".")
 
         # If we got a non-qualified name, then we do a standard reverse resolve
         # which, assuming DNS is working, will work around that windows bug
--- a/lib/ohai/mixin/network_helper.rb
+++ b/lib/ohai/mixin/network_helper.rb
@@ -19,6 +19,7 @@
 #
 
 require "socket" unless defined?(Socket)
+require "resolv" unless defined?(Resolv)
 
 module Ohai
   module Mixin
@@ -37,6 +38,11 @@ module Ohai
         dec
       end
 
+      def ip?(hostname)
+        !!(canonname =~ Resolv::IPv4::Regex) || !!(canonname =~ Resolv::IPv6::Regex)
+
+      end
+
       # This does a forward and reverse lookup on the hostname to return what should be
       # the FQDN for the host determined by name lookup (generally DNS).  If the forward
       # lookup fails this will throw.  If the reverse lookup fails this will return the
@@ -52,12 +58,21 @@ module Ohai
           .first
 
         canonname = ai&.canonname
-        # use canonname
-        return canonname if canonname != hostname || !ChefUtils.windows?
+        # use canonname if it's an FQDN
+        # This API is preferred as it never gives us an IP address for broken DNS
+        # (see https://github.com/chef/ohai/pull/1705)
+        # However, we have found that Windows hosts that are not joined to a domain
+        # can return a non-qualified hostname)
+        return canonname unless ip?(canonname)
 
-        # canonname does not fully qualify the hostname if on Windows node that
-        # is not joined to a domain, but getnameinfo does.
-        ai&.getnameinfo&.[](0) || hostname
+        # If we got a non-qualified name, then we do a standard reverse resolve
+        # which, assuming DNS is working, will work around that windows bug
+        # (and maybe others)
+        canonname = ai&.getnameinfo&.first
+        return canonname unless ip?(canonname)
+
+        # if all else fails, return the name we were given as a safety
+        hostname
       end
 
       def canonicalize_hostname_with_retries(hostname)
--- a/lib/ohai/mixin/network_helper.rb
+++ b/lib/ohai/mixin/network_helper.rb
@@ -41,7 +41,7 @@ module Ohai
       # Addrinfo#ip*? methods return true on AI_CANONNAME Addrinfo records that match
       # the ipv* scheme and #ip? always returns true unless a non tcp Addrinfo
       def ip?(hostname)
-        !!(canonname =~ Resolv::IPv4::Regex) || !!(canonname =~ Resolv::IPv6::Regex)
+        Resolv::IPv4::Regex.match?(hostname) || Resolv::IPv6::Regex.match?(hostname)
       end
 
       # This does a forward and reverse lookup on the hostname to return what should be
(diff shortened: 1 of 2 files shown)


lib/ohai/mixin/network_helper.rb
```diff
@@ -47,7 +47,17 @@ def hex_to_dec_netmask(netmask)
...
+          .getaddrinfo(hostname, nil, nil, nil, nil, Socket::AI_CANONNAME)
+          .first
+
+        canonname = ai&.canonname
+        # use canonname
+        return canonname if canonname != hostname || !ChefUtils.windows?
+
+        # canonname does not fully qualify the hostname if on Windows node that
+        # is not joined to a domain, but getnameinfo does.
+        ai&.getnameinfo&.[](0) || hostname
```
jaymzh, line 60:
```suggestion
        ai = Addrinfo
          .getaddrinfo(hostname, nil, nil, nil, nil, Socket::AI_CANONNAME)
          .first

        canonname = ai&.canonname
        # use canonname if it's an FQDN
        # This API is preferred as it never gives us an IP address for broken DNS
        # (see https://github.com/chef/ohai/pull/1705)
        # However, we have found that Windows hosts that are not joined to a domain,
        # they can return a non-qualified hostname)
        if canonname&.include?('.')
          return canonname
        end
        
        # If we got a non-qualified name, then we do a standard reverse resolve
        # which, assuming DNS is working, will work around that windows bug
        # (and maybe others)
        canonname = ai&.getnameinfo&.first
        if Addrinfo.ip?(canonname)
          return canonname
        end
        
        # if all else fails, return the name we were given as a safety
        return hostname
```

jaymzh (changes requested):
how's this?

lib/ohai/mixin/network_helper.rb
```diff
@@ -37,6 +38,11 @@ def hex_to_dec_netmask(netmask)
         dec
       end
 
+      def ip?(hostname)
```
jaymzh, line 41: Addrinfo has it's own `ip?`, use that instead

lib/ohai/mixin/network_helper.rb
```diff
@@ -47,7 +53,26 @@ def hex_to_dec_netmask(netmask)
...
+          .getaddrinfo(hostname, nil, nil, nil, nil, Socket::AI_CANONNAME)
+          .first
+
+        canonname = ai&.canonname
+        # use canonname if it's an FQDN
+        # This API is preferred as it never gives us an IP address for broken DNS
+        # (see https://github.com/chef/ohai/pull/1705)
+        # However, we have found that Windows hosts that are not joined to a domain
+        # can return a non-qualified hostname)
+        return canonname unless ip?(canonname)
```
jaymzh, line 66: this isn't what you want, this will have the same bug. If this returns "myhostname", you'll return it, and that's not what you want. 

We KNOW this won't be an IP address since `canonname` only returns names, so we just want to check it has a dot in it.

lib/ohai/mixin/network_helper.rb
```diff
@@ -47,7 +53,26 @@ def hex_to_dec_netmask(netmask)
...
+
+        # If we got a non-qualified name, then we do a standard reverse resolve
+        # which, assuming DNS is working, will work around that windows bug
+        # (and maybe others)
+        canonname = ai&.getnameinfo&.first
+        return canonname unless ip?(canonname)
```
jaymzh, line 72: this works, but use `Addrinfo.ip?`

jaymzh (changes requested):

lib/ohai/mixin/network_helper.rb
tpowell-progress: Per discussion... `Addrinfo#ip*?` methods aren't quite what we want.

lib/ohai/mixin/network_helper.rb
tpowell-progress: Discussed. `Addrinfo#ip?` doesn't quite get us there.

lib/ohai/mixin/network_helper.rb
```diff
@@ -47,7 +54,27 @@ def hex_to_dec_netmask(netmask)
...
+          .first
+
+        canonname = ai&.canonname
+        # use canonname if it's an FQDN
+        # This API is preferred as it never gives us an IP address for broken DNS
+        # (see https://github.com/chef/ohai/pull/1705)
+        # However, we have found that Windows hosts that are not joined to a domain
+        # can return a non-qualified hostname).
+        # Use a '.' in the canonname as indicator of FQDN
+        return canonname if /\./.match?(canonname)
```
jaymzh, line 68: Don't regex when you can include, much cheaper.

```suggestion
        return canonname canonname.include?('.')
```

jaymzh (changes requested):
one more minor fix

lib/ohai/mixin/network_helper.rb
tpowell-progress: `return canonname if canonname.include?(".")`

Stromweld:
That's M$ narrow minded attempt to say that you have to be joined to an active directory domain and should be using their AD managed DNS. In reality if there is a DNS server defining a domain and a host entry for it, then there is a FQDN for a host It's just being managed by other means.

jaymzh:
> On non-domain joined devices, there is no FQDN
> 
> **Important**
> 
> By default the computer name of a computer that is not joined to a domain is a host name, and not a fully qualified domain name (FQDN). Topology Builder uses FQDNs, not host names. So, you must configure a DNS suffix on the name of the computer to be deployed as an Edge Server that is not joined to a domain. Use only standard characters (including A-Z, a-z, 0-9, and hyphens) when assigning FQDNs to your servers running Skype for Business Server. Do not use Unicode characters or underscores. Nonstandard characters in an FQDN are often not supported by external DNS and public CAs (that is, when the FQDN must be assigned to the SN in the certificate).
> 
> Source - [Microsoft](https://learn.microsoft.com/en-us/skypeforbusiness/plan-your-deployment/network-requirements/basics)

That's just garbage. That's only true in the "Windows Domain" sense, not in the "real world of networks" sense. If you have a DNS name pointing at you from a domain, you have an FQDN.

jaymzh (approved):

tpowell-progress:
I'm going forward with this. The fall through handles the (mostly Windows-specific) use case, restoring the previous functionality there. If all else fails we'll return the original hostname.

PR #1824, opened 2024-06-07
Add platform_backend attribute for Target Mode

thheinen:
## Description
Enable Chef to detect if a node is managed in Target Mode vs agent mode


tpowell-progress (approved):
Approving for merge with a minor version bump. Need to evaluate tests.

PR #1838, opened 2024-09-11
FreeBSD: Skip inet6 addresses matching ::1

### LINKED ISSUE #1835: ip6address not detected correctly on FreeBSD
Running Chef on FreeBSD and the `node['ip6address']` is always `::1`

feld:
Fixes #1835

### DIFF
--- a/lib/ohai/plugins/freebsd/network.rb
+++ b/lib/ohai/plugins/freebsd/network.rb
@@ -68,6 +68,7 @@ Ohai.plugin(:Network) do
         end
       end
       if line =~ /\s+inet6 ([a-f0-9\:]+)%?(\w*)\s+prefixlen\s+(\d+)\s*\w*\s*([\da-fx]*)/
+        next if $1 == "::1"
         iface[cint][:addresses] ||= Mash.new
         if $4.empty?
           iface[cint][:addresses][$1] = { "family" => "inet6", "prefixlen" => $3 }


tpowell-progress:
@feld please add DCO

lib/ohai/plugins/freebsd/network.rb
```diff
@@ -68,6 +68,7 @@
         end
       end
       if line =~ /\s+inet6 ([a-f0-9\:]+)%?(\w*)\s+prefixlen\s+(\d+)\s*\w*\s*([\da-fx]*)/
+        next if $1 == "::1"
```
tpowell-progress, line 71: I'm not sure if this is the right approach. Is `::1` the only `inet6` line in your `ifconfig`?

feld: No, I have fe80 addresses on all the interfaces and those are properly excluded.

Sometimes ::1 is on lo1 and not lo0, but it should always see my inet6 addresses starting with 2600 first and it doesn't pick them up. It always takes the ::1 as the default address for some strange reason.

On a very vanilla out-of-the-box FreeBSD install with IPv6 enabled Chef will pick up ::1 as the IPv6 address. This is the only way I found to fix it.

jaymzh:
After writing https://github.com/chef/ohai/pull/1841 I THINK I know what the problem is here.

FreeBSD reports no scope for global addresses. Further, it uses scopeIDs, not scopes, so the sorting [that happens here](https://github.com/chef/ohai/blob/main/lib/ohai/plugins/network.rb#L54) doesn't work at all.

So the right fix here is to modify [this code](https://github.com/chef/ohai/blob/main/lib/ohai/plugins/freebsd/network.rb#L70) to:

1. Convert scopeIds into scope names
2. If no scopeID is there, just set it to 'global' which seems to be what FreeBSD is assuming.

jaymzh:
@feld - ping?

dafyddcrosby:
> @dafyddcrosby wants to look at this - will wait to close.

If If I haven't tested this by December 3rd's community PR review (or if @feld responds), can close then

dafyddcrosby:
> > @dafyddcrosby wants to look at this - will wait to close.
> 
> If If I haven't tested this by December 3rd's community PR review (or if @feld responds), can close then

@jaymzh Didn't get time over the weekend, and don't know when I will in the next few weeks, so let's close for now.

PR #1863, opened 2025-02-06
Updating FFI support

johnmccrae:
## Description
Ohai hasn't been updated in a bit. It's current dependency on a stale version of ffi is causing errors in Chef-19. We also needed to update the supported Ruby versions

## Types of changes

## Checklist:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,7 @@ gem "chef-utils", git: "https://github.com/chef/chef", branch: "main", glob: "ch
 group :development do
   install_if -> { Gem::Version.new("3.3.0").to_s >= (RUBY_VERSION).to_s } do
     gem "racc", "~> 1.8.0"
+    gem "base64"
   end
   gem "cookstyle", ">= 7.32.8"
   gem "ipaddr_extensions"
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,9 @@ gem "chef-utils", git: "https://github.com/chef/chef", branch: "main", glob: "ch
 group :development do
   install_if -> { Gem::Version.new("3.3.0").to_s >= (RUBY_VERSION).to_s } do
     gem "racc", "~> 1.8.0"
+  end
+  install_if -> { Gem::Version.new("3.4.0").to_s >= (RUBY_VERSION).to_s } do
+    gem "racc", "~> 1.8.0"
     gem "base64"
   end
   gem "cookstyle", ">= 7.32.8"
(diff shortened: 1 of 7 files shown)


tpowell-progress:
@johnmccrae looks like you still have the cookstyle load error due to `racc/parser` not being available.

tpowell-progress (changes requested):
`install_if` `racc` if >= 3.3 ruby version.

Gemfile
```diff
@@ -10,6 +10,13 @@ gem "chef-config", git: "https://github.com/chef/chef", branch: "main", glob: "c
 gem "chef-utils", git: "https://github.com/chef/chef", branch: "main", glob: "chef-utils/chef-utils.gemspec"
 # NOTE: do not submit PRs to add pry as a dep, add to your Gemfile.local
 group :development do
+  install_if -> { Gem::Version.new("3.4").to_s >= (RUBY_VERSION).to_s } do
+    gem "racc", "~> 1.8.0"
+    gem "base64"
+  end
+  install_if -> { Gem::Version.new("3.3").to_s >= (RUBY_VERSION).to_s } do
+    gem "racc", "~> 1.8.0"
```
tpowell-progress, line 18: @johnmccrae this conflicts with the above `racc` specification for 3.4. The `install_if` ultimately acts as though the `gem` command were evaluated at the top or group level for the purpose of recording the `Gemfile.lock`.

Also the version should be the following (the `.to_s` is unnecessary and breaks the comparison logic in certain scenarios such as (because string comparison instead of version comparison)
```ruby
irb(main):011> Gem::Version.new("3.3").to_s >= "3.3.0".to_s
=> false
```

This should be the order of things (validated that `RUBY_VERSION` first works)
```ruby
  install_if -> { RUBY_VERSION >= Gem::Version.new("3.4") } do
    gem "base64"
  end
  install_if -> { RUBY_VERSION >= Gem::Version.new("3.3") } do
    gem "racc", "~> 1.8.0"
  end
  ```

Comment (tpowell-progress) [reply Gemfile:18]: Also, there should be `Gemfile.lock` output that has `racc` and `base64`

**Review by tpowell-progress — CHANGES_REQUESTED** [review]

File: ohai.gemspec [hunk 2]
```diff
@@ -12,17 +12,17 @@ Gem::Specification.new do |s|
   s.email = "adam@chef.io"
   s.homepage = "https://github.com/chef/ohai/"
 
-  s.required_ruby_version = ">= 2.7"
+  s.required_ruby_version = ">= 3.0"
 
   s.add_dependency "chef-config", ">= 14.12", "< 20"
   s.add_dependency "chef-utils", ">= 16.0", "< 20"
-  s.add_dependency "ffi", "~> 1.9"
+  s.add_dependency "ffi", ">= 1.15.5", "< 1.17.0"
```
jaymzh, line 19: we should probably do https://github.com/chef/ohai/pull/1866 first, I think.

Gemfile
jaymzh: FWIW, I found `install_if` to be super unreliable and switched to simple `if`s which worked much better.

johnmccrae:
this PR was rendered obsolete.

PR #1885, opened 2025-08-05
Fixing network detection when using multipath route

achamo:
Some setup could have a multipath route with a src assigned to the loopback
In this situation, it fails to find a mac because there is none found on the loopback
Better early return 00:00:00:00:00:00 when checking for a mac on lo

## Description
In a setup with BGP and multipath routes with a src address defined on the loopback interface, `ohai network` returns an empty `interfaces` set
It fails in `get_mac_for_interface` raising an error because there is no `lladdr` address.

## Types of changes

## Checklist:


jaymzh (changes requested):
This seems sane, but you have lint failures, please fix those up and we'll get this merged.

Issue #1890, opened 2025-10-03
oci_metadata does not collect on Windows


snohio:
### Describe the Enhancement
Currently the oci_metadata.rb file is not written to collect metadata from Oracle Cloud on Windows platforms.

### Describe the Need
All Windows platforms that run on OCI need to collect metadata for use in cookbooks.

### Current Alternative
None - other than creating a custom ohai plugin.

### Can We Help You Implement This?
I believe I can handle it.

P.S. There is also an enhancement for adding more detail to the event handler when it is not a 200 to indicate which endpoint might be causing issues.

snohio:
Fixed by #1891

PR #1891, opened 2025-10-03
Added support for Windows Platform for oci metadata capturing.

snohio:
Author: michael.butler@progress.com

Added support for Windows platforms to collect OCI Metadata

## Description
The update adds in support for the Windows platforms to collect OCI Metadata. Previously, the logic only validated a file that is placed on the linux filesystem. It now looks at the Winodws Registry with wmi_lite if that file is not in place.

This also adds a small fix for the reporting if anything other than a 200 comes back from the environment. We moved from a warn to a debug and also included the endpoint that failed for better error tracking. This is to help prevent incidents in the future.

## Related Issue
https://github.com/chef/ohai/issues/1890

## Types of changes

- [X] Chore (non-breaking change that does not add functionality or fix an issue)

## Checklist:

### DIFF
--- a/spec/unit/mixin/oci_metadata_spec.rb
+++ b/spec/unit/mixin/oci_metadata_spec.rb
@@ -49,7 +49,7 @@ describe Ohai::Mixin::OCIMetadata do
       http_mock = double("http", { code: "404" })
       allow(mixin).to receive(:http_get).and_return(http_mock)
 
-      expect(mixin.logger).to receive(:warn)
+      expect(mixin.logger).to receive(:debug)
       vals = mixin.fetch_metadata
       expect(vals).to eq(nil)
     end


tpowell-progress (changes requested):
@snohio `cookstyle --chefstyle -a` for the whitespace complaints below

spec/unit/mixin/oci_metadata_spec.rb
```diff
@@ -49,7 +49,7 @@
       http_mock = double("http", { code: "404" })
       allow(mixin).to receive(:http_get).and_return(http_mock)
 
-      expect(mixin.logger).to receive(:warn)
+      expect(mixin.logger).to receive(:debug)
```
Stromweld: Mike and I fixed this spec test. We want to have the log message for debug and not do a warn as it is confusing to see that warning that looks like a failure when volumes endpoint is expected not to exist unless you configure the cloud instance to have a special high performance volume. This is creating confusion for some users and moving to debug makes sense for troubleshooting and to remain silent in normal operation.

jaymzh (approved):

PR #1963, opened 2026-07-30
CHEF-37444: add ~/.chef/ruby/VERSION/gems to GEM_PATH for dynamic plugin loading

sanghinitin:
## Summary

Ohai plugins installed via `chef gem install` land in
`~/.chef/ruby/<RUBY_API_VERSION>/gems`, but this path was never added to
the runtime `GEM_PATH`. As a result, any gem installed that way was
invisible to the packaged ohai binary even after re-running the binstub.

This PR fixes the issue by expanding `GEM_PATH` in three places: the Ruby
binstub patch, and both the Linux/macOS and Windows Habitat wrappers.

## Root Cause Evidence

```
$ gem info some-ohai-plugin
some-ohai-plugin (x.y.z)
    Installed at: /Users/nsanghi/.chef/ruby/3.4.0/gems
# → not in GEM_PATH → LoadError at runtime
```
https://progresssoftware.atlassian.net/browse/CHEF-34004
### binstub_patch.rb

- Removed the old single-path `GEM_PATH` assignment.
- Now prepends three paths before any existing `GEM_PATH`:
  1. `$pkg_prefix/vendor` — ohai vendored deps (flat layout)
  2. `Gem.user_dir` → `~/.gem/ruby/VERSION` — standard `gem install` target
  3. `~/.chef/ruby/RUBY_API_VERSION/gems` — `chef gem install` target
- Uses `.uniq` so paths are never duplicated when the block runs twice.

### habitat/plan.sh & habitat/aarch64-darwin/plan.sh

- Runtime wrapper now exports four `GEM_PATH` entries:
  1. `$pkg_prefix/vendor` — flat vendor tree
  2. `$pkg_prefix/vendor/ruby/VERSION` — bundler-nested tree (train-core etc.)
  3. `${HOME}/.gem/ruby/VERSION` — standard gem install
  4. `${HOME}/.chef/ruby/VERSION/gems` — chef gem install
- `ruby_gem_version` is now derived from the packaged Ruby's on-disk
  `lib/ruby/gems/` directory via `basename` instead of executing
  `ruby -e`. The latter can silently fail during `do_install` (runtime lib
  paths are not yet set up), producing an empty version string that
  corrupts all `GEM_PATH` entries. Falls back to `3.4.0` with a
  `build_line` warning if the glob still yields nothing.

### habitat/plan.ps1

- After `appbundler.bat` generates Windows binstubs, injects
  `binstub_patch.rb` immediately after the `require "rubygems"` line in
  each binstub — the same insertion point used on Linux/macOS.

## Verification

After `hab pkg build .` and `hab pkg install`:

```
# 1. Confirm wrapper GEM_PATH has all four segments and a non-empty version:
grep GEM_PATH /hab/pkgs/chef/ohai/*/*/bin/ohai

# 2. Install a test gem into the chef dir:
RUBY_API=3.4.0
gem install paint --install-dir "$HOME/.chef/ruby/$RUBY_API" --no-document

# 3. Run ohai with a plugin that requires the gem — must print version, not LoadError:
cat > /tmp/test.rb <<'PLUGIN'
Ohai.plugin(:GemPathCheck) do
  provides "gem_path_check"
  collect_data do
    gem_path_check Mash.new
    gem_path_check[:gem_path] = Gem.path
    gem_path_check[:paint_found] = begin
      require "paint"; Gem.loaded_specs["paint"]&.version.to_s
    rescue LoadError => e; "FAILED: #{e.message}"; end
  end
end
PLUGIN
/hab/pkgs/chef/ohai/*/*/bin/ohai -d /tmp gem_path_check
# Expected: paint_found: "2.x.x"  ← gem loaded from ~/.chef/ruby path
```

This work was completed with AI assistance following Progress AI policies

### DIFF
--- a/binstub_patch.rb
+++ b/binstub_patch.rb
@@ -1,4 +1,18 @@
 unless ENV["APPBUNDLER_ALLOW_RVM"]
   ENV["APPBUNDLER_ALLOW_RVM"] = "true"
-  ENV["GEM_PATH"] = [File.expand_path(File.join(__dir__, "..", "vendor")), ENV["GEM_PATH"]].compact.join(File::PATH_SEPARATOR)
 end
+
+# Prepend the package vendor dir, the current user's gem dir, and the Chef gem dir to
+# GEM_PATH so that:
+#   1. Ohai's vendored dependencies are always found.
+#   2. Plugins installed via `gem install <plugin>` are discoverable on both
+#      Linux (~/.gem/ruby/VERSION) and Windows (%USERPROFILE%\.gem\ruby\VERSION).
+#   3. Gems installed via `chef gem install` (~/.chef/ruby/VERSION/gems) are immediately available.
+#
+# This block runs before appbundler's env_sanitizer (which calls Gem.clear_paths and
+# re-reads GEM_PATH from ENV), so these additions are always picked up correctly.
+ohai_vendor = File.expand_path(File.join(__dir__, "..", "vendor"))
+# chef-cli gem install places gems under ~/.chef/ruby/RUBY_API_VERSION/gems
+chef_gem_dir = File.join(Dir.home, ".chef", "ruby", RbConfig::CONFIG["ruby_version"], "gems")
+existing_paths = ENV["GEM_PATH"]&.split(File::PATH_SEPARATOR) || []
+ENV["GEM_PATH"] = ([ohai_vendor, Gem.user_dir, chef_gem_dir] + existing_paths).uniq.join(File::PATH_SEPARATOR)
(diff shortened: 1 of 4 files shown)


tpowell-progress (dismissed):
Are... we doing this on every tool that has this problem? Sounds like the loader and/or pathing of habitat or chef install needs to be fixed instead.

tpowell-progress (dismissed):
This should be fixed once at the top level rather than duplicated per-repo. Each of ohai, knife, chef-cli, and chef-test-kitchen-enterprise is independently patching its own binstub/Habitat wrapper to add `~/.chef/ruby/VERSION/gems` to `GEM_PATH`. That logic belongs in a single shared place (the Habitat plan tooling/base plan, or the chef-cli/chef loader that these tools share) so all consumers inherit the fix instead of hand-rolling the same GEM_PATH patch N times with N chances to drift out of sync.

tpowell-progress:
Discussed with @sanghinitin and while this isn't ideal, the only solution that comes close is an intermediate package which would still require repo-specific changes

sanghinitin:
https://github.com/chef/appbundler/issues/84

tpowell-progress (approved):

